PR #64, opened 2026-08-12
statusline: warn when the plane root is being worked in

diazoxide:
The plane root — the directory holding `charter.toml` — is the control plane, and since
[ADR 0007](docs/adr/0007-one-plane-shape.md) it is not drawn as a repo at all. [ADR
0008](docs/adr/0008-the-plane-root-is-not-a-work-tree.md) is about the gap that leaves:
**not presenting a tree is not the same as preventing work in it.** Two sessions that both
sit in the plane root share one working tree and one HEAD and thrash each other's branches,
while charter reports two different workspaces and lists no tree that would hint at why.

So the root gets **one line, only when it is dirty or off its default branch**:

```
┌──────────────────────────────────────────────────────────────────────────────────┐
│ ⬢ default · ws 1                                                                 │
├──────────────────────────────────────────────────────────────────────────────────┤
│ ▪ repos 1                                                    │ ▪ personas 1      │
│   └─ iam-service            main                             │ ▫ steward ·       │
│ ⚠ plane root acme · dirty · on hotfix/thing, not main · work belongs in a … clone │
├──────────────────────────────────────────────────────────────────────────────────┤
│ ctx 22% · ⚡97%                                                  ⬢ charter 0.20.0 │
└──────────────────────────────────────────────────────────────────────────────────┘
```

### Where it goes, and why

`_alerts` — the full-width lines between the table and the session strip. The warning is a
property of neither the active workspace (the top line) nor this session (the bottom
strip), but of the plane itself: the same scope as the nested-plane and pinned-version
warnings it now sits with, and the same discipline — it renders only when real, so a
healthy plane spends no row on it.

Full width also happens to be the safe width. The row has no right-hand neighbour, so
unlike a row in the repo column it cannot shear a column by being one cell wider than
`tui.width` believes, and the only glyph it adds is the `⚠` those alerts already use.

It is checked **last** inside `_alerts`, whose single guard returns what it has already
collected — so a failure here costs its own line rather than the warnings above it that
carry commands of their own. There is a test for that ordering.

### What "default branch" means

`origin/HEAD` is the repository's own answer and wins outright; then the conventional
names, and only when a ref by that name actually exists here (loose *or* packed). When
nothing says, **nothing is claimed** — a plane with no remote and a hand-named branch has
no default to be off, and a warning manufactured from a guess would fire forever on a tree
nobody is misusing. Dirtiness still speaks there.

Read from the filesystem, like `util.branch_of`: this runs on every render. Dirtiness
reuses `_repo_states`, so it shares the existing TTL cache — one `git status` per few
seconds, not per render. The root deliberately does **not** join `render`'s scan list,
because that list also feeds `glstate`'s background forge fetches, and a repo with no row
must not spawn network work for itself.

### Silence

- clean root on its default branch → no row, no zero, no badge;
- plane root that is not a git repo (`charter init` in a fresh directory — the README's
  60-second path) → silent;
- plane created *inside* another repo → silent, because asking git about a directory with
  no `.git` of its own answers for the surrounding tree, and reporting a tree the user
  never named as "the plane root" is worse than saying nothing;
- the root is still not counted among your repos, and still gets no tree row.

### Tests

23 new, `tests/test_statusline_plane_root_warning.py`, real git in a temp dir throughout —
every fact under test (HEAD, `origin/HEAD`, the ref store, dirtiness) is a fact about a
repository on disk, and a mocked git would prove nothing about any of them. Each was seen
to fail before the implementation existed, and the load-bearing ones were mutation-checked
afterwards (hardcoding `main` as the default, dropping the "has its own `.git`" guard, and
moving the call to the front of `_alerts` each break a specific test).

`python3 -m unittest discover -s tests -t .` → **1424 tests, OK** (1401 on `main`).

https://claude.ai/code/session_0145Anhzojfo151sYXYwwest

### DIFF
--- a/charter/statusline.py
+++ b/charter/statusline.py
@@ -1097,9 +1097,33 @@ def _default_branch(gitdir: Path) -> str | None:
     return None
 
 
+def _head_detached(gitdir: Path) -> bool:
+    """True when HEAD names a commit rather than a branch.
+
+    Its own read rather than an inference from :func:`_branch`, which collapses the two:
+    a detached HEAD reads back from there as a short sha, and a short sha is not
+    distinguishable from a branch someone named after one. The distinction has to be
+    made where git makes it — HEAD is a symref (``ref: refs/heads/…``) or it is not.
+
+    Reads the gitdir it is GIVEN, never the common dir — the opposite of
+    :func:`_default_branch`. HEAD is per-worktree (that is the whole point of a linked
+    worktree), while refs are shared, so following ``commondir`` here would report the
+    main tree's branch for a worktree standing on its own.
+
+    ``False`` when HEAD cannot be read at all. "Unknown" is not "detached", and
+    manufacturing the most alarming of the three states out of a failed read is exactly
+    the false alarm this element cannot afford.
+    """
+    try:
+        txt = (gitdir / "HEAD").read_text().strip()
+    except OSError:
+        return False
+    return bool(txt) and not txt.startswith("ref:")
+
+
 def _plane_root_alert() -> str | None:
-    """One line when the **plane root** is being worked in — dirty, or off its default
-    branch. ``None`` otherwise, which is the ordinary case.
+    """One line when the **plane root** is being worked in — dirty, detached, or off its
+    default branch. ``None`` otherwise, which is the ordinary case.
 
     The root is the directory holding ``charter.toml``: the control plane itself, and
     since ADR 0007 not a repo row at all — a plane's `repo_trees` is its clones and
@@ -1115,7 +1139,7 @@ def _plane_root_alert() -> str | None:
     nothing — the row has no right-hand neighbour, so unlike a row in the repo column it
     cannot shear a column by being one cell wider than `tui.width` believes.
 
-    Both findings share ONE line. A row is spent on every single turn, and "dirty AND off
+    All findings share ONE line. A row is spent on every single turn, and "dirty AND off
     main" is one situation with two symptoms.
 
     "Dirty" here means ``tracked_dirty`` — untracked files excluded, which is exactly
@@ -1159,11 +1183,20 @@ def _plane_root_alert() -> str | None:
             # The word, not the repo table's `*`. A marker reads as a marker beside a
             # column of them; this line has no siblings to be read against.
             bits.append(f"{_YELLOW}dirty{_R}")
+        if _head_detached(gitdir):
+            # Reported on its own terms and WITHOUT a default to compare against, unlike
+            # the branch case below. Detachment is a fact about HEAD alone, it is the
+            # most alarming of these states rather than the least — a commit made here is
+            # unreachable the moment anything else is checked out — and staying silent
+            # about it on a plane whose default cannot be discovered would be the wrong
+            # way round. `doctor`'s `check_plane_root` reports it the same way and on the
+            # same terms; the words match so the two cannot be read as different findings.
+            bits.append(f"{_YELLOW}detached HEAD{_R}")
         # `?` is `_branch`'s "HEAD unreadable" — nothing to compare, so nothing to claim.
-        # A detached HEAD is not `?`: it reads as a short sha, which differs from any
-        # default, and a detached HEAD in the plane root is the same accident wearing a
-        # different hat.
-        if default and branch not in ("?", default):
+        # A detached HEAD never reaches here: it is already said above, and saying it
+        # again as "on 1a2b3c4, not main" would dress the worst state up as an ordinary
+        # branch you happened to be on.
+        elif default and branch not in ("?", default):
             bits.append(f"{_DIM}on{_R} {branch}{_DIM}, not{_R} {default}")
         if not bits:
             return None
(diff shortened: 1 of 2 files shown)


diazoxide:
Rebased onto `d963887` and fixed the disagreement — thanks for catching it before it merged.

**Approach for (1): a second notion of dirt derived from the `git status` `_run_state` already runs**, not a second status call and not a change to what `dirty` means.

`_run_state` now returns `tracked_dirty` alongside `dirty`, from the same command:

- `dirty` — unchanged, untracked included. Still what a repo row's `*` means, what `_pick_rows` ranks on and what the `…(+N more, all clean)` line claims. No other row's behaviour moves.
- `tracked_dirty` — the same minus `??` entries, which is exactly what `--untracked-files=no` prints. Only `_plane_root_alert` reads it.

Why this rather than its own status call: only `_repo_states`' path is behind the TTL cache, so a separate `git status` for the root would be a subprocess on **every render**, for a question that changes about as often as a commit. Deriving both from the one answer keeps it at one `git status` per repo per few seconds and puts the two notions of dirt side by side, in the one place that reads git's output, with the reason written down. `.get("tracked_dirty")` defaults to `False`, not to `dirty` — a cache entry written seconds ago by an older charter has no such key, and a few silent seconds beats a warning derived from the notion `doctor` disagrees with.

**(2)** `test_an_untracked_file_alone_does_not_raise_the_warning`, with the `share = "local"` reasoning in the docstring, plus `test_a_tracked_edit_beside_untracked_files_still_speaks` so the narrower notion cannot become a mute button. Both were mutation-checked: flipping the read back to `dirty` fails exactly these two and the agreement test below.

**(3)** The docstring paragraph is rewritten — it now states what was chosen, that untracked is excluded, and that `doctor` asks git the same `-uno` question, with the furniture argument as the reason and the agreement as the constraint.

**New: `DoctorAndTheStatusLineCannotDisagree`.** Four cases asserting the two modules *agree* on the same tree — dirt with an untracked file, dirt with a tracked edit, the default branch from `origin/HEAD`, and no default when nothing says — rather than each matching its own expected string, so whichever side changes its mind next has to change the other with it. No code is shared between the modules, per your steer; the lock is in the test.

I confirmed their default-branch resolutions were already identical in behaviour (`origin/HEAD` → a `main`/`master` that actually exists → no claim), reached by reading refs here and asking git there.

**One narrow difference I did *not* change, flagged for your call:** a plane root with a detached HEAD *and* no discoverable default. `doctor` warns (`detached HEAD` is a finding independent of the default); the status line stays silent, because its branch clause requires a default to compare against. With a default present both warn, so this only bites a plane with no `origin/HEAD`, no `main` and no `master`. It is a branch-side difference rather than the dirt one you asked about, and fixing it means the status line reading `HEAD` itself to tell detachment from a branch name — happy to do it in this PR or leave it.

`python3 -m unittest discover -s tests -t .` → **1447, OK** (1418 on `main` at `d963887`; 29 in the new module, up from 23).

diazoxide:
Closed it — you were right on both counts, and the first one is the one that decided it: a class named `DoctorAndTheStatusLineCannotDisagree` with a documented case where they do makes the class claim more than it checks, and the name is what the next reader trusts.

`ec3d123`:

**`_head_detached(gitdir)`** — reads `.git/HEAD` and asks whether it is a `ref:` line, which is where git itself makes the distinction. Filesystem-only, per the module contract. Its own read rather than an inference from `_branch`, which collapses the two states: a detached HEAD reads back from there as a short sha, and a short sha is not distinguishable from a branch someone named after one. It reads the gitdir it is *given* and deliberately does **not** follow `commondir` — the opposite of `_default_branch`, because HEAD is per-worktree even though refs are shared; following it would report the main tree's branch for a worktree standing on its own. Unreadable HEAD returns `False`: "unknown" is not "detached", and manufacturing the most alarming state out of a failed read is the false alarm this element cannot afford.

**Reported on its own terms**, not as a branch comparison. With a default discoverable the old code printed `on 1a2b3c4, not main`, which dresses "whatever you commit here is unreachable the moment anything else is checked out" up as an ordinary branch you happened to be on. It now reads `detached HEAD`, the same words `doctor` uses, so the two cannot be read as different findings:

```
⚠ plane root acme · detached HEAD · work belongs in a workspace clone
doctor: warn | detached HEAD
```

(that pair is from a real plane on `trunk`-renamed-away, no `origin/HEAD`, no `main`, no `master`, plus an untracked memory — which correctly counts as neither side's dirt.)

**Three tests**, all mutation-checked by forcing `_head_detached` to `False` (the pre-review behaviour), which fails exactly these three:

- `test_a_detached_head_is_reported_as_detachment_not_as_a_branch` — asserts `detached` present and `not main` absent, so the wording cannot regress to a branch comparison.
- `test_a_detached_head_speaks_even_when_no_default_is_discoverable`.
- `DoctorAndTheStatusLineCannotDisagree.test_they_agree_that_a_detached_head_speaks_with_no_default_to_compare` — the one that would have caught this. It guards its own fixture too (asserts `doctor._plane_default_branch` really is `None`, and that `doctor` really does warn), so it cannot pass by both sides being quiet.

`python3 -m unittest discover -s tests -t .` → **1449, OK** (31 in the module, up from 29).

PR #229, opened 2026-08-17
0.41.0 — doctor says whether a guard ever actually ran, and init stops printing one enormous line

diazoxide:
Version bump plus two fixes that could not wait for a separate PR, because the second one is what makes the release's own assets correct.

## What ships

**#226 — declared is not dispatched.** `check_guard_wired` proves a hook line exists in a settings file. It cannot see whether anything ever *called* it, and it would have reported a tick throughout the afternoon a plane root was switched between branches four times, unguarded. `pretooluse` now records that it ran and under which harness; doctor reports an **age**, never a verdict:

```
✓  guard seen    last ran 3m ago under claude-code
!  guard seen    no guard has ever run in this plane
```

**#227 — two hook tests wrote into the developer's real control plane.** `PersonaIso` cannot reach a subprocess, and a plain `TestCase` does not use it. Harmless while those handlers wrote nothing; not harmless once `find_root` began hopping outward through `workspaces/` in 0.37.0, because a checkout sitting inside somebody's plane then resolves to **that plane**. It had been leaking trace rows already — `guard-seen` only made it visible.

## Two fixes found while regenerating the assets

**`charter init` printed one enormous line.** It named every path it had written, unwrapped. It grew every time charter learned to write something new — **184 characters at 0.38, 254 once harness wiring landed** — so it was unreadable in an 80-column terminal *and* made the README's demo image 254 columns wide. The headline is a count now; the paths are listed beneath it. A test pins the width so it cannot creep back.

**The capture pipeline was documenting the wrong charter.** It ran whatever `charter` was on `PATH`, so regenerating against an installed 0.40.0 faithfully reproduced an `init` line the tree had *already fixed*. It now runs the tree's own code — which is what "generated from real command output" was supposed to mean all along.

```
demo capture width:   254 → 134 columns
```

## The freshness guard did its job

`captured.json` was stamped 0.39.0. Bumping to 0.41.0 put it two minors behind and the suite failed, which is exactly the design: one release of slack, then it speaks. That failure is what sent me to regenerate, and regenerating is what surfaced both fixes above.

Suite: **2321 passing**.

https://claude.ai/code/session_01MUyeVUu8iLZHLsUDUMvdUX


diazoxide:
Superseded. #228 released 0.41.0 concurrently and carried the guard-seen and isolation work, so the only parts of this PR still needed are the init summary shape and the capture pipeline — reopened as a focused PR on top of the release.

PR #348, opened 2026-08-21
Resolve and bound every read of plane data (#336)

### LINKED ISSUE #336: Nothing checks for a symlink when charter reads plane data, so a committed link redirects the read and can block it forever
Charter reads plane data by listing a directory and opening what it finds. Across `charter/persona.py`, `charter/memstore.py` and `charter/recall.py` there is **no call to `is_symlink`, `lstat` or `realpath`** for containment, and no bound on how much is read. Git stores symlinks, so both properties are attacker-reachable from a commit.

**Consequence one — the read goes somewhere else, into a place a guard already protects.** `pretooluse-read` denies file-reading tools pointed at a vault (`charter/hooks.py`, `_VAULT_PATH_RE`), and `SECURITY.md` names that guard explicitly. It works — pointed at a vault file it returned:

```
{"permissionDecision":"deny",
 "permissionDecisionReason":"charter guard: reads a vault/secret file directly
  (would print plaintext). Use `charter … secret exec`/`cp` instead of reading `.charter/`"}
```

But when a persona's `persona.md` is a link rather than a file, `persona.load()` → `read_text()` (`charter/persona.py:143-152`) follows it, and the target's contents become that persona's charter. Verified on 0.47.2: with `persona.md` resolving to a plain-file vault under `.charter/vaults/`, `resolve()` returned the vault's JSON as the charter body, and `charter persona sync-agents` wrote it into `.claude/agents/<name>.md` — an agent's system prompt. The canary I planted, `CANARY-LOCAL-VAULT-SECRET-4d8e`, appeared in the generated file. The guard denied the model reading the file; charter read it and handed it over. Note the path involved needs no knowledge of the operator's home directory — the vault directory sits at a fixed place relative to the plane.

The same applies in `refs/`, where `recall.py:127-142` walks with `rglob` and `is_dir()`, both of which follow links. `charter recall` printed content from outside the plane under an in-plane path label.

**Consequence two — the read never finishes.** Nothing bounds these reads: no size cap, no timeout. With `persona.md` resolving to a path that blocks on read, on 0.47.2:

```
charter doctor:            HUNG >10s
charter statusline:        HUNG >10s
charter hook sessionstart: HUNG >10s
charter persona show:      HUNG (indefinitely)
```

`hooks/hooks.json` caps the SessionStart hooks at 5s and 20s, so a session loses its briefing rather than wedging. The status line and `persona show` have no charter-side bound at all. A link to a device that yields endlessly gives the memory-exhaustion variant instead.

**Why this is not #324.** #324 is about `util.run` lacking a `timeout` and inheriting stdin — a *subprocess* that never returns. This is a blocking read inside charter's own process, with no child involved, so a subprocess timeout cannot see it. The two want different fixes and closing one does not close the other.

**Reachability, honestly.** Commit access to the plane. The read half also needs whatever consumes the charter — `sync-agents`, `persona show`, or `recall` — and the vault-file variant needs a plain-file or reference vault to exist locally, which is the common case. Nothing here is remote. #90 already established that the vault guard is Bash-only and that other tools walk past it; this is the same lesson arriving from inside charter rather than from a sibling tool.

**Suggested direction, none chosen here.**

1. One resolver that refuses a path whose `realpath` leaves the plane, used by every reader of plane data — the same helper #328 proposes.
2. Refuse a `persona.md`, memory file or ref that is a symlink at all, and say so in `lint`. Blunt, and it would break anyone deliberately linking shared refs.
3. Independently, a size cap and a read deadline on plane-data reads, so consequence two stays closed even if a link is later allowed on purpose.

Two parts, one root cause — filed together deliberately, because fixing the containment without the bound leaves the hang, and vice versa.

- [ ] containment: reads of plane data cannot leave the plane
- [ ] bound: reads of plane data cannot block or grow without limit

---
Found during an authority audit of 0.47.2.

diazoxide:
Closes #336 — both boxes, and the reachable variants of each.

## What was open

Charter read plane data by listing a directory and opening what it found. Nothing on any of those paths called `is_symlink`, `lstat` or `realpath`, and nothing bounded what came back. Git commits symlinks, so both properties belonged to whoever last committed to the plane.

Verified A/B on a throwaway plane, installed 0.47.2 against this branch:

| | 0.47.2 | this branch |
|---|---|---|
| `persona.md` → `../../.charter/vaults/devops.json`, then `sync-agents` | canary in `.claude/agents/reader.md` | no agent file, `✗ no persona 'reader'` |
| FIFO at the active `persona.md` → `charter statusline` | **hung >10s** | renders |
| … → `charter doctor` | **hung >10s** | reports |
| … → `charter persona show reader` | **hung >10s** | `✗ no persona 'reader'` |

The vault case is the one that matters: charter's own `pretooluse-read` guard denies an agent reading that file — it denied *me* mid-investigation, when a `cat` of the fixture tripped it — and the link walked around it into a sub-agent's system prompt.

## (a) Containment — closes in full

`contain.py` grows the resolving layer beside its lexical one: `dir_refusal` (a listing) and `file_refusal` (a read). Callers: `memstore.files` (so `entries`, `search`, `resolve`, `index_drift`, `count_open`, `curate` and `todos` all inherit one gate), `memstore.resolve`'s direct hit, `persona.load` via a new `definition_refusal`, and `recall._ref_dirs`.

**One correction to the issue's suggested direction.** "Refuse a path whose `realpath` leaves the plane" would have admitted the flagship demonstration unchanged: `.charter/` sits *under* ROOT (`config._migrate_state_dir`), so `../../.charter/vaults/devops.json` never leaves the plane. The boundary is the plane's **data** roots — `personas/`, `workspaces/`, `PERSONA_STATE_DIR` — which excludes the secrets home and includes the one part of it that is data. Two tests hold that line from opposite sides: a `refs/` linked into `.charter/reports/` is refused, and ephemeral memory (which lives *inside* `.charter/`) is still read. Drawing the boundary at ROOT fails the first; drawing it by banning `.charter/` fails the second.

**Links are followed, escapes are refused** — #342's reason for staying lexical was that it would break a plane legitimately symlinking a persona directory. Resolving keeps that working: a link landing inside the data roots is followed, and a plane that relocates `personas/` or `workspaces/` behind a link keeps working because the roots are resolved too. Four of the nine cases in `test_plane_reads_are_contained.py` are that benign half.

## (b) Liveness — closes in full

The brief's hypothesis held, and it is the whole reason this half is cheap: **`os.lstat` answers regular-or-not, how big, and link-or-not in one syscall — the syscall containment already pays for — and it answers without opening anything.** A read deadline would have had nothing to time out, because nothing is opened. A FIFO is not `S_ISREG`; neither is `/dev/zero`.

The cap is **1 MiB**, set where nothing an editor produces can reach it (largest persona charter on this plane: 6.8 KB; largest `docs/` page: 34 KB) rather than close to real content, where the first long runbook somebody curates would trip it. The just-under-cap case is a test, because a cap that never fires is decoration and one that fires on real content is worse.

Three more committed files read on the same paths get the same gate — `personas/.default` (answers "who am I" every turn), a persona's `mcp.json` (which already promised never to raise on a bad file), and a workspace's `workspace.md` (read once per workspace at SessionStart). And the session briefing counted another workspace's todos with `len(open_todos(w))` — every todo of every workspace, read in full, for a number — while the status line counted the same todos with `todos.count_open`. One question, two implementations, the expensive one at session start.

## Cost, measured on this plane

`file_refusal` 1.2µs (one `lstat`); `dir_refusal` 5.5µs on a persona directory via a fast path that is exact — a path whose parent *is* a data root and which is not itself a link cannot have moved relative to that root — and 20µs on a deeper base, paid once per listing. End to end: a 5-persona `structural_errors` sweep 1.22ms → 1.38ms, `recall(query)` 4.97ms → 5.06ms.

## Tests

20 new cases in two files, one per half. Every case asserts its preconditions: the escaping links assert `is_symlink` and that the OS follows the link to the canary *before* charter is asked to refuse; every hang case proves a thread that opens the fixture is still blocked when charter has already returned. The RED run of the liveness file spent 21s in watchdog timeouts; it now runs in 1.9s.

Five existing test files built a memory store in a bare `mkdtemp`, which no production caller produces — every one derives its base from `memory_dir`/`todos_dir`/`refs_dir`. They now build it in a plane, and their headers say why so nobody reverts it and reads the empty listing as a search bug.

Full suite: **3065 passed** (3043 on `main` + 22), 82s — green locally and in CI on 3.11, 3.12, 3.13 and 3.14.

## Found while here, not fixed here

**A committed link redirects charter's *writes* too, and `MEMORY.md` is a fixed name.** With `personas/x/memory/MEMORY.md` a link to a file outside the memory directory, `charter persona remember` appends its index line *through* the link. Demonstrated on 0.47.2 **and on this branch** — the write paths (`ensure_index`, `index_append`, `_drop_index_line`, `write`) do not go through `files()`, so this PR does not close it. Pointed at `.charter/vaults/<name>.json` it corrupts a credential store from a commit, with no guessing required, and `file_refusal` is the same helper it would need (ENOENT would have to read as "nothing to object to" rather than a refusal, since a write's target usually does not exist yet). Reported rather than folded in: this workspace's rule is findings before fixes, with a stop in between.

https://claude.ai/code/session_01RLeoNSEaQceHHvn4AhP8xo

### DIFF
--- a/charter/contain.py
+++ b/charter/contain.py
@@ -212,7 +212,7 @@ def within_data(path) -> bool:
         roots = [os.path.abspath(r) for r in data_roots()]
         if os.path.dirname(target) in roots and not stat.S_ISLNK(os.lstat(target).st_mode):
             return True
-    except OSError:
+    except (OSError, ValueError):
         pass                                   # vanished or unreadable — ask the slow path
     try:
         target = os.path.realpath(path)
@@ -221,14 +221,18 @@ def within_data(path) -> bool:
             # `+ os.sep` so a sibling named like a root ("personas-old") is not a child.
             if target == root or target.startswith(root + os.sep):
                 return True
-    except OSError:
+    except (OSError, ValueError):
         return False
     return False
 
 
 def _not_plane_data(path) -> str:
     roots = ", ".join(sorted(Path(r).name for r in data_roots()))
-    return NOT_PLANE_DATA.format(name=path, target=os.path.realpath(path), roots=roots)
+    try:
+        target = os.path.realpath(path)
+    except (OSError, ValueError):
+        target = path            # unresolvable — say what was asked for, never raise here
+    return NOT_PLANE_DATA.format(name=path, target=target, roots=roots)
 
 
 def dir_refusal(directory) -> str | None:
@@ -274,6 +278,11 @@ def file_refusal(path) -> str | None:
             st = os.stat(path)
     except OSError as e:
         return UNREADABLE.format(name=path, error=e.strerror or e)
+    except ValueError as e:
+        # `os.lstat` raises ValueError, NOT OSError, on a path holding a NUL — the one
+        # input shaped to get past a check (`segment_ok` refuses it for the same reason).
+        # "Nothing here raises" is this module's promise; catching only OSError broke it.
+        return UNREADABLE.format(name=path, error=e)
     if not stat.S_ISREG(st.st_mode):
         kind = next((k for test, k in _KINDS if test(st.st_mode)), "not a file")
         return NOT_A_FILE.format(name=path, kind=kind)
--- a/charter/contain.py
+++ b/charter/contain.py
@@ -32,14 +32,26 @@ charter's creation-time alphabet on someone else's forge would refuse to clone b
 :func:`segment_ok` is therefore the permissive rule: it forbids traversal and separators
 and nothing else.
 
-**Containment is lexical, and does not follow symlinks.** :func:`docsrc.read` resolves its
-page precisely to catch a symlink pointing out of the directory, and copying that here
-would be a mistake in two directions. It would do half of #336 — whose containment half is
-about symlinks in *every* file charter reads, not only the ones named by a name — while
-claiming none of it; and it would refuse a plane that legitimately symlinks a persona
-directory today, which is a working plane broken in exchange for a hole that stays open
-anyway. Traversal and absoluteness are what these five issues are about. Symlinks are
-filed, and stay filed.
+**Two layers, and the second one resolves.** :func:`segment_ok` and :func:`child` stay
+lexical: they answer a question about a *string*, and asking the filesystem would make a
+traversal succeed exactly when the attacker's target happens to exist. That left every
+read charter performs — not only the ones named by a name — open to a committed symlink,
+which is #336 and is now :func:`file_refusal` / :func:`dir_refusal` below. The two are
+deliberately separate functions: a name is refused before anything is opened, and a *path*
+is refused before it is read, and neither can stand in for the other.
+
+**What "inside the plane" had to mean.** The obvious rule — refuse a path whose `realpath`
+leaves ROOT — admits #336's own demonstration unchanged, because ``.charter/`` sits
+*under* ROOT (:func:`config._migrate_state_dir`), so ``personas/x/persona.md`` →
+``../../.charter/vaults/devops.json`` never leaves the plane while doing precisely what the
+`pretooluse-read` guard exists to stop. The boundary is therefore the directories a plane
+keeps its **data** in — :func:`data_roots` — which excludes the secrets home and includes
+the one part of it that is data (ephemeral memory, which lives inside ``.charter/``).
+
+**Links are followed; escapes are refused.** Refusing every symlink would close the hole
+and break a plane that legitimately links a persona directory. Resolving keeps that plane
+working, and #342's reason for staying lexical was never that symlinks are acceptable —
+only that doing half of this while claiming all of it would be worse than filing it.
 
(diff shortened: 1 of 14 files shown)


diazoxide:
Verification note, added after the initial push: `contain`'s promise that nothing in it raises did not survive its own new code path. `os.lstat` answers a path holding a NUL with `ValueError`, not `OSError`, so the one input shaped to get past a check turned a refusal into a crash — four lines below the `segment_ok` comment that refuses NUL for exactly that reason. Fixed in `ef8d255`, with a case that asserts both `file_refusal` and `dir_refusal` refuse rather than raise.

Final suite on the branch head: 3064 passed, 85s.

Issue #411, opened 2026-08-23
frame: panels do not follow a workspace switch made inside the harness


diazoxide:
Reported by the operator: switching workspace from inside `charter claude` (asking the harness to switch) updates the **status line** but the **frame panels keep showing the old workspace**.

## Where it breaks — candidates, not yet confirmed

`slots._top` calls `workspace.resolve()` at paint time, so the renderer is live and is not the bug. `resolve()`'s precedence is:

```
--workspace → $CHARTER_WORKSPACE → the tree you are standing in → per-session pointer → per-terminal pointer → declared default → default
```

For a panel: `$CHARTER_WORKSPACE` is normally unset, and its cwd is the plane root, so `from_path` yields nothing. That leaves the two pointers.

**The terminal pointer cannot match.** `_terminal_id()` is `session.terminal()`, which reads `TMUX_PANE`. The harness and each panel are *different panes*, so `workspace.use()` writing `_terminal_file(<harness pane>)` is invisible to a panel reading `_terminal_file(<panel pane>)`.

**So everything rests on the session pointer**, and that only works by a collision #386 documents: the frame launcher exports the frame id under `$CHARTER_SESSION_ID`, the variable `session.current()` already owns. Writer and reader agree only while both resolve to the frame id.

Candidate root causes, in the order worth checking:

1. **An explicit session id defeats the collision.** `workspace.use(name, session_id=...)` and `_session_id(explicit)` both accept one. If any path passes the harness's real session id, the pointer is written under that key while panels read under the frame id — a silent miss with no error anywhere.
2. **The version never bumps**, so the panel simply never repaints. `notify.plane_changed` is called from seven `hooks.py` sites and bumps `state.version` for the frame id; worth confirming a Bash-tool `charter ws use` actually reaches `posttooluse-bash` inside a frame.
3. **The panel repaints against a stale gather.** `notify` refreshes the cache before bumping specifically to close that window, but `_left`/`_right` read `gather.read(fid)` and only `_top` resolves the workspace live — so the sidebars could lag even when the top line is correct.

## Why this matters more than it looks

The frame's whole claim is that it shows the plane's live state. A panel that silently pins to the workspace it launched with is worse than no panel: it is confidently wrong, and the status line sitting next to it disagrees.

## Note for whoever fixes it

[...]

Issue #386, opened 2026-08-23
frame v2 (2/3): the frame owns the surface — suppress the status line inside a frame, and ADR 0019


diazoxide:
With parity (#385), the status line and the frame show the same things twice. The frame should own the surface.

**Suppression means render empty, not stop running** — and this is the non-obvious part. Claude Code invokes `charter statusline` every ~10s and passes token usage in the payload; `_record_turn` writes the history the context gauge reads. Hooks never see those fields (`hooks.py` has zero references to usage). So disabling the command would destroy the data source for a renderer we just ported. Returning empty keeps the recording and removes the duplicate.

**The decision belongs in `cmd_statusline`, never in `render()`.** A check on ambient `$CHARTER_SESSION_ID` inside `render` would fire during the test suite — which now runs inside a frame — making eight-plus `test_statusline_*` modules pass or fail depending on where the developer ran them. Keeping it at the command edge means no existing statusline test changes.

**Liveness, cheaply:** the frame id ends with the launcher's pid, so `os.kill(pid, 0)` answers "is that frame alive" with no subprocess on a 10-second render path. A stale directory must not blank the status line forever.

**A human running `charter statusline` by hand should still get output** — render when stdout is a TTY, blank when piped, which is how Claude Code calls it.

**Settle the `CHARTER_SESSION_ID` collision here.** The launcher exports the frame id under the variable `session.current()` already owns. It is currently load-bearing and undocumented: panels follow `charter ws use` **only because** of the collision. Suppression keys on the same variable, so this issue must decide and document it, with a test pinning that a panel follows `charter ws use`.

**ADR 0019 — the frame owns the surface.** ADR 0018 settled that charter may run the harness but never draws it; this is its sequel: which charter surface draws when two exist. The part that will otherwise be "cleaned up" by someone who does not know why is a command kept running for its recording side effect.

Honest limit to record: codex and opencode still get no context gauge, because nothing feeds them one.

Depends on #385. Follow-up from #380.

Issue #383, opened 2026-08-23
frame: reap can delete a sibling frame's recorded exit code


diazoxide:
`state.reap(live)` runs at every launch and removes any frame directory not naming a live tmux session — including a sibling whose session has just ended but whose launcher has not yet read its `exit` file. Demonstrated: `record_exit(fidA, 42)` → `reap({other})` → `exit_code(fidA)` is `None`, which `cmd_launch` turns into a returned 0.

There is a precise, non-heuristic fix: **the frame id ends with the launcher's pid**, so reap can refuse any directory whose trailing pid is still a live process. No age heuristic, no new state.

Follow-up from #380.

Issue #402, opened 2026-08-23
The test suite runs `frame.state.reap` against the developer's real .charter/, deleting live frame state


diazoxide:
Running `python3 -m unittest discover -s tests` on `main` calls `frame.state.reap` against the
developer's **real** `.charter/frame/` three times, and creates directories in it. On a machine
with a live `charter` frame, that deletes the frame state of running sessions — including an
unread `exit` file, the thing #383 exists to protect.

**Measured, not inferred.** `state.reap` and `state.frame_dir` were replaced with recorders
*before* the tests ran, so this observation destroyed nothing:

    REAL plane state dir at import: /Users/aharon/IdeaProjects/charter/.charter
    total recorded touches: 9
    touches resolving INSIDE the operator's real plane: 9
      REAP  /Users/aharon/IdeaProjects/charter/.charter/frame
      MKDIR /Users/aharon/IdeaProjects/charter/.charter/frame/demo-13779
      MKDIR /Users/aharon/IdeaProjects/charter/.charter/frame/demo-13779
      ... (x3)

**Cause.** Two classes reach the real `commands_frame.cmd_launch` through the module's
`_launch` helper and are plain `unittest.TestCase`:

- `tests/test_frame_launcher.py:88` `class MissingHarnessBinary(unittest.TestCase)`
- `tests/test_frame_launcher.py:1718` `class BypassRouting(unittest.TestCase)`

Every *tmux* call on that path is faked, which is what makes the module read as harmless.
`state.reap` is not faked — and `_FakeTmux` answers the pre-`new-session` `list-sessions` with
an empty set, so those tests run `reap(set())` against the operator's plane.

Two more write into it without reaping, in the one file allowed to start real tmux:

- `tests/test_frame_tmux_integration.py:960` `MenuFormatIntegration(_NeedsAttachedClient, unittest.TestCase)`
- `tests/test_frame_tmux_integration.py:1137` `MenuClientIntegration(_NeedsAttachedClient, unittest.TestCase)`

They leave `menu-fmt-integ*-<pid>/` and `menu-client-integ-<pid>/` behind on every run —
present in this plane right now. Their sibling `MenuIntegration` **does** inherit `PersonaIso`
and leaves nothing, which is the contrast that identifies it.

**Why it needs an issue rather than a patch in passing.** The two halves have different fixes.
`MissingHarnessBinary`/`BypassRouting` need `PersonaIso` (or an explicit refusal of the real
[...]

PR #419, opened 2026-08-23
frame v2 (3/3): density presets and animation scoped to in-flight work (#387)

### LINKED ISSUE #387: frame v2 (3/3): density presets and animation scoped to in-flight work
Once there is content worth showing (#385), the frame needs a way to show less of it.

**Density is a preset over `slots`, not a second config system.** `[frame] slots` is the real primitive; `density` expands to a slot list plus a per-slot verbosity, and an explicit `slots` overrides it. Levels: `minimal` (one-line top and bottom), `normal`, `full` (all four edges).

**The hotkey overrides for the running frame only.** The config sets the default; a keypress must not rewrite `charter.toml`, which is hand-maintained — charter's own rule is that machine-written config belongs somewhere a machine may rewrite whole. Live re-layout via `split-window`/`kill-pane`.

**Animation runs only while work is in flight.** A spinner while a dispatch, clone or `gl-refresh` runs; complete stillness at idle. `inflight.live_records()` is file-based and cheap, so the panel self-ticks while records exist and returns to version-watching when they clear — self-limiting by construction. The retreat, if it measures badly, is a static `⋯ 2 running`.

Nothing here may cost anything at idle. That is the property the frame was built for.

**Fold in the resize-hook warning's placement** while moving warnings: it is the one standing warning still printed into the pre-attach window, which tmux's alternate screen hides milliseconds later (measured: it lands 86 bytes before `\e[?1049h`). `RESIZE_HOOK_FLOOR` appears in neither `frame_ready` nor `doctor.check_frame`, and it sits *above* `FLOOR`, so a tmux 3.2 operator passes the floor cleanly and still silently loses resize recovery.

Depends on #385. Follow-up from #380.

diazoxide:
Closes #387.

Three things, plus the standing warning #387 asked to be folded in.

## 1. Density is a preset over `slots`

`[frame] density` expands to a slot list **plus** a per-slot verbosity, and an explicit
`slots` overrides it — `slots` stays the primitive. Nothing downstream of
`instance.frame_of` learns that presets exist: `frame_ready`, `doctor.check_frame`,
`_drawable_slots` and `layout.panel_argvs` all keep reading the one resolved list.

| level | edges | verbosity |
|---|---|---|
| `minimal` | `top`, `bottom` | terse |
| `normal` | `top`, `bottom` | normal |
| `full` | all four edges | normal |

`terse` is real, not decorative: `top` drops the charter version (the install is not where
you are standing), `bottom` keeps only its highest-priority non-empty field through the
same `_fit_fields` priority order it already used for width, and `left`/`right` cap at four
rows and say how many they hid.

**Expansion happens only when the key is DECLARED**, never from the shipped default. That
is what keeps the shipped `slots` list load-bearing rather than dead — see the
reconciliation section below.

## 2. The hotkey overrides the running frame only

The three levels join "Detach" in the menu `[frame] hotkey` already opens, with a `•` on
the one in effect. Selecting one runs `charter frame-density <level>`, which re-lays the
frame out live: disarm and `kill-pane` what is no longer wanted, `split-window` what is,
rebuild the `window-resized` hook from the new pane set, re-assert every size, give the
harness pane focus back, bump the version so surviving panels repaint at the new
verbosity.

**It writes nothing to `charter.toml`.** The override lands in the frame's own state
directory, which `reap` deletes whole when the frame ends — machine-written state in a
place a machine may rewrite whole, per charter's own rule. Relaunch and the configured
default is back. Pinned by a test that makes every `charter.toml` writer in `instance`
raise.

No second `bind -n`: a tmux key table is server-wide with no per-session form, so a
second key would be taken from every frame on `SOCKET` and could not be bound at all
inside an operator's own tmux. One bind, already paid for.

Two things that were not obvious and are now enforced:

- **Disarm before killing.** `kill-pane` on a panel fires that pane's own `pane-died`
  hook; `cmd_respawn` waits out its backoff, finds the session perfectly alive (only the
  layout changed) and puts the panel straight back, one respawn life poorer. The hook is
  unset first, so there is nothing left to fire whichever way this tmux treats it.
- **New panes carry the frame environment on the operator's server** (`-e`), for #411's
  root cause one command over. `_pane_env_for` names the private/operator split once so a
  second pane-creating caller cannot get it the other way round, and rebuilds the env
  through `_frame_env` so `TMUX`/`TMUX_PANE`/`COLUMNS`/`LINES` from the pane the command
  was fired in never reach a new panel.

## 3. Animation scoped to in-flight work — and the measured idle cost

`⠙ 2 running` on the bottom row while dispatches are in flight; complete stillness
otherwise. A presumed-dead record is reported and deliberately **not** animated (`⋯ 1
stalled`) — such a record is kept for a day, and a spinner beside it would claim progress
that stopped half an hour ago.

**The number, measured rather than asserted** (macOS/APFS, `mutate`-free build):

| per panel tick | ns |
|---|---|
| `state.version` poll (what a panel already paid) | 26 035 |
| the in-flight gate this PR adds | 4 252 |
| **added by #387** | **~5 100** |
| an ungated `inflight.live_records()` per tick, for comparison | 17 307 (3.8×) |

At `panel.TICK` (5/s) that is ~0.003% of one core, and **zero repaints**. The mechanism:
`inflight.stamp()` is a single `stat` of the tracker's directory, whose mtime moves only
when a record is created or removed; the records are re-read only when that number moves,
or when the earliest presumed-dead deadline passes (the one way the answer changes with no
file touched — and a deadline only exists while something is running, so the idle path
never computes one). Pinned as a **syscall count**, not a timing assertion:
`test_an_idle_tick_costs_exactly_one_stat`.

`statusline._session_news` gained one keyword-only `inflight=` so the panel can ask for
everything except the tracker's own `⚡ N` — otherwise the row read `⚡ 2 · ⠙ 2 running`,
one fact twice. The status line is unchanged.

## 4. `RESIZE_HOOK_FLOOR`, moved where it will be read

It appeared in neither `frame_ready` nor `doctor.check_frame`, and it sits *above*
`FLOOR` — so a tmux 3.2 operator passed the floor cleanly, saw a green tick on both
surfaces, and silently lost resize recovery. Both now name it
(`tmuxctl.below_resize_hook_message`, one sentence shared, so the two cannot drift), and
`_install_resize_hook`'s own `util.warn` — the last standing ceiling still printed into
the pre-attach window, 86 bytes before `\e[?1049h` — is gone. The two *discoveries* there
(a tmux that turns out not to know the hook name, a `set-hook` that fails for another
reason) stay: `frame_ready` starts nothing and cannot answer them.

`check_frame` collects the three ceilings into a list rather than nesting `if`s, which
also removes the duplicated slot ceiling the old shape needed in two branches.

## Reconciling the default with #386 — please check this

#387 defines `full` as all four edges; #386 makes all four edges the shipped `slots`
default. **These are two ways of asking for the same frame, so they must expand to the
same one**, or writing nothing and writing `density = "<the shipped default>"` give
different frames, silently.

I did **not** make the shipped `density` drive the slot list. Precedence is:

    explicit `slots`  >  explicit `density`  >  the shipped `slots` default

If the shipped density expanded unconditionally, `FRAME_FIELDS["slots"]`'s default value
would become dead code and **#386's change to that line would silently do nothing**. This
way #386 keeps owning the shipped frame and `density` is a preset, exactly as #387 says.

The agreement is then enforced **mechanically**, not by memory —
`test_frame_density.ShippedDefaultsAgree`:

- the shipped `density` must be a level whose slot list *is* the shipped `slots` list;
- **in the same order**, because order is geometry (see below);
- and whose verbosity is `normal` — `minimal` and `normal` expand to the same two edges,
  so the slot-list check alone cannot tell them apart, and a shipped `minimal` would
  quietly ship a terse frame to everyone.

**On this branch (`slots` still `["top", "bottom"]`) the shipped `density` is `normal`.**
When #386 lands, that test goes red until one word changes to `full`. That is the intended
behaviour of the test: a forced one-word edit at merge time instead of a divergence nobody
sees. I verified the whole thing rather than predicting it — cherry-picked this commit onto
`frame-owns-surface-386`, resolved the two conflicts (both adjacent-line, in
`instance.py`'s `FRAME_FIELDS` and in `docs/frame.md`), flipped the default to `full`, and
ran the suite: **4129 tests, OK**. Mutating the default back to `normal` or to `minimal` on
that merged tree is red both ways.

**Order is geometry, and every level's list says so.** `full` is
`["top", "bottom", "left", "right"]`, matching #386's measured order: at 200×50 that gives
a 200-column `bottom`, while `["top", "left", "right", "bottom"]` gives **154** columns,
inset between the sidebars — and `bottom` is the row carrying the alert and the command
that fixes it, which `_bottom` drops whole fields to fit.
`test_full_puts_the_strips_before_the_side_panels` pins it as index comparisons (strips
before sides) rather than as a frozen literal, so the next person cannot alphabetise it.

## Two defects found on the way

- **A corrupt live density override silently discarded the configured one.**
  `state.density(fid) or config.FRAME["density"]` looks right and is not: the file holds
  text, so a truncated write is truthy, beats the config, and *then* degrades to the
  default verbosity — the operator's own `[frame] density` thrown away by a byte in a file
  they have never heard of. The override is validated before it is allowed to win.
- **A recycled launcher pid handed a new frame a dead one's density and pane map** —
  #383's bill, two files longer. `state.clear_shape` is called on both launch paths beside
  `clear_exit`/`gather.discard`.

## Verification

- **Full suite: `Ran 4115 tests … OK`** on this branch; `Ran 4129 tests … OK` on the trial
  merge with #386.
- **Mutation-tested: 44 mutants, 0 survivors**, restored suite green, `__pycache__` cleared
  between every run. Three survivors in the first pass were real test gaps and are the
  reason this PR has `TheLoopAsksBeforeItAnimates`,
  `Probe.test_a_tmux_without_the_resize_hook_is_a_ceiling_the_probe_names`, and the
  shipped-verbosity assertion; a fourth pointed at a `_NEVER_READ` sentinel that no test
  could falsify, which is deleted rather than documented.
- `tests/test_frame_density.py` uses `PersonaIso` throughout, and a genuinely dead pid for
  every fixture id (a `something-1` reads as `launchd` and cannot fail).
- No new runtime dependency; stdlib `unittest` only.

## One thing in the issue I could not deliver as written

#387 says "a spinner while a dispatch, **clone or `gl-refresh`** runs". Only dispatches are
tracked: `inflight.start` is called from exactly one place (the `Task`/`Agent` PreToolUse
handler), and clones and `gl-refresh` register nothing. I did not wire them in, because
`inflight`'s records are also what feeds the dispatch overlap nudge via `still_running()` —
a record named `clone` would make it tell an operator that "`x` writes code and `clone` are
already running". That needs a second channel or a kind field on the record, which is a
design decision, not an implementation detail. The mechanism paragraph of the issue names
`inflight.live_records()` as the source, and that is what shipped.

### DIFF
--- a/charter/cli.py
+++ b/charter/cli.py
@@ -676,7 +676,7 @@ def _add_frame_parsers(sub) -> None:
     # naming `frame-probe` silently launching a harness instead of reading tmux's own
     # version (see `commands_frame.cmd_probe`'s own docstring).
     _core_commands = set(sub.choices) | {"frame", "panel", "frame-menu", "frame-action",
-                                         "frame-probe", "frame-respawn"}
+                                         "frame-probe", "frame-respawn", "frame-density"}
 
     # Which harness (by `.name`, never `.cli_name` — that's the dict key below) has
     # already claimed each word, so a SECOND harness wanting it is told who got there
@@ -758,6 +758,20 @@ def _add_frame_parsers(sub) -> None:
     rs.add_argument("slot")
     rs.add_argument("--pane", dest="pane", required=True)
     rs.set_defaults(func=commands_frame.cmd_respawn)
+
+    # Internal, and a top-level sibling for the same `_split_frame_argv` reason as the
+    # three above. Fired by a hotkey-menu selection (`commands_frame._menu_entries`), whose
+    # stored argv is exactly `util.self_relaunch_argv("frame-density", <level>)`. It
+    # changes the RUNNING frame only — never charter.toml, which is hand-maintained (see
+    # `cmd_density`'s own docstring) — and, like every other `frame-*` command here,
+    # resolves which frame from `$CHARTER_SESSION_ID` at the moment it fires rather than
+    # from anything baked into a shared menu action. Deliberately NOT `choices=` on the
+    # level: `instance.density_level` is the one gate on that closed set, and a second
+    # copy of it in argparse would mean a level added to the table and not to the parser
+    # exits 2 from inside a `run-shell` where nothing prints the reason.
+    dn = sub.add_parser("frame-density")
+    dn.add_argument("level")
+    dn.set_defaults(func=commands_frame.cmd_density)
 
     # A TOP-LEVEL sibling of `frame` for a DIFFERENT reason than `frame-menu`/
     # `frame-action` above: those two exist because `_split_frame_argv` eats everything
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -290,18 +290,26 @@ def frame_ready() -> tuple[int, str, str]:
     launcher goes on to draw regardless — a probe that lies about `cmd_launch`'s own
     behaviour is worse than one that runs nothing at all.
 
-    **The two STANDING conditions are reported here and nowhere else.** Both used to be
-    `util.warn` calls inside `cmd_launch` itself, and both were measured to be
-    unreadable there: `util.warn` for an unimplemented slot lands 86 bytes before tmux's
-    own `\\x1b[?1049h`, so the operator's terminal switches to the alternate screen
-    milliseconds later and the line is restored to view only when the frame EXITS. A
-    warning printed where it cannot be read is worse than silence, because it creates a
-    record that the operator was told. Neither is per-launch news anyway — a tmux below
-    the floor, and a configured slot with no renderer, are true on every launch on this
-    machine and this plane until something changes. They are capability ceilings, so
-    they belong to the two surfaces built to report ceilings on demand: this one, and
-    `doctor.check_frame`. (A per-launch notice mechanism, for conditions specific to one
-    launch, is deliberately NOT built here.)
+    **The three STANDING conditions are reported here and nowhere else.** All three used
+    to be `util.warn` calls inside `cmd_launch` (or, for the resize hook, inside
+    `_draw_panels`), and all three were measured to be unreadable there: `util.warn` for
+    an unimplemented slot lands 86 bytes before tmux's own `\\x1b[?1049h`, so the
+    operator's terminal switches to the alternate screen milliseconds later and the line
+    is restored to view only when the frame EXITS. A warning printed where it cannot be
+    read is worse than silence, because it creates a record that the operator was told.
+    None is per-launch news anyway — a tmux below the floor, a tmux below
+    `tmuxctl.RESIZE_HOOK_FLOOR`, and a configured slot with no renderer are true on every
+    launch on this machine and this plane until something changes. They are capability
+    ceilings, so they belong to the two surfaces built to report ceilings on demand: this
+    one, and `doctor.check_frame`. (A per-launch notice mechanism, for conditions specific
+    to one launch, is deliberately NOT built here.)
+
+    **`RESIZE_HOOK_FLOOR` is the third, and it was the one this function did not know
+    about (#387).** It sits ABOVE `FLOOR`, so an operator on tmux 3.2 passed the floor
+    cleanly, saw a green tick on both ceiling surfaces, and silently had no resize
+    recovery at all — the gap being closed here. It does NOT change this function's exit
+    code, for the same reason the other two do not: `cmd_launch` draws the frame
+    regardless, and a probe stricter than the launcher lies about the launcher.
 
     Two callers share this, both read-only for the same reason `charter/news.py`
     requires of a `check:` (reads, never acts; and this module's own tmux calls all go
@@ -318,6 +326,8 @@ def frame_ready() -> tuple[int, str, str]:
     ceilings = []
     if v < tmuxctl.FLOOR:
         ceilings.append(tmuxctl.below_floor_message(v))
+    if v < tmuxctl.RESIZE_HOOK_FLOOR:
+        ceilings.append(tmuxctl.below_resize_hook_message(v))
     missing = frame_slots.unimplemented(config.FRAME["slots"])
     if missing:
         ceilings.append(no_renderer_message(missing))
@@ -883,8 +893,15 @@ def _wait_for_harness(socket: str, harness_pane: str) -> int | None:
         time.sleep(_POLL_SECONDS)
 
 
-def _drawable_slots(cols: int, rows: int) -> list[str]:
+def _drawable_slots(cols: int, rows: int, configured: list[str] | None = None) -> list[str]:
     """Which configured slots this frame will actually draw, at *cols* x *rows*.
+
+    *configured* defaults to `config.FRAME["slots"]` — already the density-resolved list
+    (`instance.frame_of` expands a declared `[frame] density` into it, so nothing here
+    knows presets exist). `cmd_density` passes the list a level expands to instead, so a
+    frame re-laid-out by the hotkey menu goes through exactly the same two filters a launch
+    does rather than a second copy of them: below the size floors it drops the same slots
+    in the same order, and a slot with no renderer is skipped for the same reason.
 
     `[frame] slots` can accept a slot (`instance.FRAME_SLOTS`, sized by
     `layout.SLOT_SIZE`) that `frame.slots.SLOTS` — the RENDERER registry — has no
@@ -905,8 +922,8 @@ def _drawable_slots(cols: int, rows: int) -> list[str]:
     demand instead, from the same `frame_slots.unimplemented` this filters on.
     """
     frame = config.FRAME
-    slots = layout.visible_slots(frame["slots"], cols, rows,
-                                 frame["min_cols"], frame["min_rows"])
+    slots = layout.visible_slots(frame["slots"] if configured is None else configured,
+                                 cols, rows, frame["min_cols"], frame["min_rows"])
     unimplemented = frame_slots.unimplemented(slots)
     return [s for s in slots if s not in unimplemented]
 
@@ -1095,12 +1112,18 @@ def _launch_in_operator_tmux(socket: str, session: str, *, fid: str, argv: list[
     # version: whatever is under the id predates the frame now claiming it.
     state.clear_exit(fid)
     gather.discard(fid)
+    state.clear_shape(fid)
     # Rewritten, not merely written: an adopted directory may name the OTHER server, and
     # a stale marker would make `reap` there skip a frame that is now genuinely ours.
     state.record_server(fid, socket)
     state.bump(fid)
 
     env = _frame_env(fid, h)
+    # The launcher is the only process that knows this frame's own charter identity —
+    # see `state.record_identity` for the measurement. Written before any pane exists,
+    # because a `run-shell` child fired later reads the SERVER's environment, not this
+    # one's.
+    state.record_identity(fid, _frame_identity_env(env))
     cwd = os.getcwd()
     opened = tmuxctl.run(
         "opening a window for the frame",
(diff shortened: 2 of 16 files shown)


diazoxide:
Rebased onto `origin/main` @ `e407852`, which now carries #418 — the `RealPlaneWrite` tripwire for #402. That is the guard the brief warned this work must not add to, so it is worth stating what it says about this branch rather than leaving it to the CI tick:

- **Full suite green under the tripwire: `Ran 4192 tests … OK`.** `tests/test_frame_density.py` derives every state-touching case from `PersonaIso`; the three classes that do not (`ShippedDefaultsAgree`, `DensityResolves`, `DensityIsWiredIntoTheCli`) touch no filesystem at all — they read the `FRAME_DENSITY` table, call `frame_of` on literal dicts, and build a parser.
- **Mutation suite re-run on the rebased tree: 44 mutants, 0 survivors**, restored suite green.

Nothing in the diff changed; this is the same commit rebased.

The #386 reconciliation is unchanged and still pending: #412 is open, and when it lands this branch needs the two adjacent-line conflicts resolved (`FRAME_FIELDS`, `docs/frame.md`) and the shipped `density` flipped `normal` → `full`. `ShippedDefaultsAgree` is what forces that flip rather than leaving it to be remembered — it goes red on the merged tree until the word changes, in both directions.

diazoxide:
Rebased onto `a1b8905` (#412 merged). **`ShippedDefaultsAgree` fired**, as designed — red until `density` moved `normal` → `full`, and red in both directions if it had not. The mechanism did its job; nothing was found green that should have been red.

It was **not** the two-conflict resolution I rehearsed. Four files conflicted, and checking your four points against the merged tree turned up two duplications and three unfalsifiable guards of my own.

**1. `_FRAME_IDENTITY` (five names now).** My `_pane_env_for` was a parallel construction; deleted. New `_pane_identity_env(env, v)` delegates to `_frame_identity_env` and gates on `tmuxctl.PANE_ENV_FLOOR`, and `cmd_launch`'s own private-path call now goes through it too — so the sixth name reaches every pane charter creates, not the ones somebody remembered. The test asserts against `commands_frame._FRAME_IDENTITY` itself rather than a list of its own, for exactly the reason the tuple already grew from four to five.

Related, and the reason this needed care rather than a rename: **my `panes` file was keeping a second copy of the harness pane id**, which #412 made authoritative in `record_harness_pane` for `is_live`. Two records of one fact, written at different moments, free to disagree. `record_panes` now stores slots only and `cmd_density` reads `state.harness_pane(fid)`.

**2. Fourth suppression condition.** Nothing in my diff enumerates the suppression conditions — I checked rather than assumed (`grep` over the diff for suppress/blank/quiet). The only new sentence near it is a `check_frame` docstring line saying `_statusline_suppressed_note` rides on the hint but is deliberately *not* a ceiling: a ceiling is a capability this machine lacks, a blanked status line is one working as designed. `check_frame`'s body is now your `quiet` note composed with my `ceilings` list.

**3. `is_live`'s new checks — and yes, a mutant went unfailable.** Not through `is_live`, through the default flip:
`test_no_density_at_all_leaves_the_shipped_slots_untouched` sabotaged the level `"normal"` by name. Once the shipped density became `full`, patching `normal` changed nothing on the default path, and the test passed against unconditional expansion too — the guard that keeps #412's own `slots` default from becoming dead code. It reads `FRAME_DEFAULTS["density"]` now. **Caught by the mutation run, not by reading it.**

**4. `respawn_argv` overlay.** Nothing of mine relied on "replaces"; `_relayout` only splits.

**Three unfalsifiable guards found and deleted rather than documented** (all mine, all introduced by the collapse above): a client-env split in the re-layout that no test could tell from its absence — a pane's environment comes from the server and `-e`, never from the tmux client, once the server is up; a `_frame_env` strip credited with protecting new panes when `_frame_identity_env`'s five-name filter already excluded `TMUX`/`TMUX_PANE`/`COLUMNS`/`LINES`; and the test asserting that strip, which could not fail while `test_a_new_pane_carries_nothing_beyond_that_identity` passed.

**`commands_frame.py:1129`/`:1170` untouched** — `git diff a1b8905` shows no `pane_env=env` line added or removed. New code does not reproduce #446 either: a re-laid-out pane carries the five identity names on both servers, pinned with an `AWS_SECRET_ACCESS_KEY` sentinel that could only arrive if the whole environment were handed over.

**Verification on the rebased tree:** full suite `Ran 4221 tests … OK`; **44 mutants, 0 survivors, no bad anchors**, restored suite green, `__pycache__` cleared between every run.

Ready for you to merge.

diazoxide:
All five blocking items fixed, plus both extra survivors and every non-blocking note. `efe3b49`. **Full suite `Ran 4234 … OK`; mutation suite now 58 mutants, 0 survivors, no bad anchors.** The review was right on all five — B1 in particular is a real bug I introduced and my own test could not see.

**B1 — fixed, and the diagnosis was exactly right.** I confirmed the four `set-environment` calls cover only `CHARTER_SESSION_ID`, `CHARTER_FRAME_EXIT`, `CHARTER_PY` and `PYTHONSAFEPATH` — no identity name among them — so a `run-shell` child on the shared private socket reads the starting launcher's plane.

I took the **state-directory** option rather than extending `_session_id_env_argv`, because the session-scoped route cannot be taken on both paths: `_launch_in_operator_tmux` may not write a session option at all (ADR 0018), so that fix would have worked on one server and left the other silently wrong. New `state.record_identity`/`state.identity`, written by the launcher on both paths beside `panes`/`density`, read by `_relayout_pane_env`.

An unrecorded identity states the four unknown names **empty rather than omitting them** — omitting is what lets the pane inherit the server's, which is the bug; empty is what charter's own readers already treat as absent, so the pane falls back to its inherited cwd. `CHARTER_SESSION_ID` is forced to `fid` rather than read back. Pinned by a test that patches `os.environ` to a *deliberately wrong* identity in all four names and asserts the recorded one wins, plus a launcher-side test with a sentinel `$CHARTER_WORKSPACE` (that one is what killed "a launch stops recording the identity", which survived my first pass).

**B2 — fixed, one line plus a registry.** `slots.ANIMATED = frozenset({"bottom"})`; `_watch` computes `animates = slot in slots.ANIMATED` and the `and` **short-circuits**, so a `top`/`left`/`right` panel no longer pays even the `stat`. Not asserted by reading the set: `OnlyTheAnimatedSlotAnimates` renders every slot twice at two clock readings with a record in flight and asserts the moving set **equals** `ANIMATED` — red in both directions, so a renderer that gains or loses a spinner without updating the set is caught. Your 4 816µs `right` figure is in the constant's docstring.

**B3 — widened.** It counts `os.stat`, `builtins.open` and `subprocess.run` at the bottom now, where every higher-level spelling must arrive. I added your M9 and M10 as mutants; both are killed, and the interpreter-dependence is gone.

**B4 — one sentence each**, in `docs/frame.md` and the news entry, naming clone/`gl-refresh` explicitly, why the dispatch tracker cannot simply absorb them (the overlap nudge would announce a clone as a peer agent), and **#420**.

**B5 — fixed, but not as a one-liner.** Unconditional unset broke two existing launch tests that correctly assert no `window-resized` command is issued when no panel came up — a launch's window is new and has nothing armed. `_install_resize_hook` takes `replacing: bool`; a launch with no panels issues nothing, a re-layout that dropped every slot issues `set-hook -w -u`. Both directions are mutants.

**M17** — pinned by reading `state.density` from inside a patched `_relayout`, which is the only moment the order is observable. **M23** — the tautology is gone: the test asserts the literal `"normal"` and separately asserts `DEFAULT_VERBOSITY == "normal"` with a reason. `DEFAULT_VERBOSITY = "terse"` is now a killed mutant.

**Non-blocking, all done.** `_verbosity` -> `verbosity`. The `ShippedDefaultsAgree` narrative now says the flip **happened** rather than predicting it (and records that the invariant needed three parts, not one). `_DENSITY_MARK` is `*` — East-Asian Narrow, the marker `git branch` uses — with the reason written down under the rule it would have crossed. `cmd_density` records no menu inside an operator's tmux, matching `cmd_launch`, with both directions tested. `_relayout`'s docstring now says what its ordering does buy (each step's invariant holds when it returns) and what it does not (it is not a transaction; tmux cannot make it one).

Still not merged, as asked.

Issue #376, opened 2026-08-22
`charter guard` is not all-or-nothing across harnesses, so one command can leave a rule in force under one and absent under another


diazoxide:
Split out of #369, which fixed the *claim* and deliberately left the behaviour.

`cmd_guard_ask` and `cmd_guard_allow` iterate `registry.all()` and apply the rule harness by harness. The refusal to touch a malformed file is per-harness, so one invocation can half-succeed:

```
$ charter guard ask 'foo *'
✗ claude-code: …/.claude/settings.json is not valid — left untouched.
•   Fix it by hand, then re-run. charter never repairs these files.
✓ opencode: asking for ('bash', 'foo *') → …/opencode.json
•   codex: codex has no command-pattern permissions — charter's own guard still applies.
!   The rule landed UNEVENLY — it is in force under the harnesses ticked above and NOT under the one refused, from this one command.
```

The Claude Code refusal is correct and the file is left byte-identical. But the plane now holds the rule under one harness and not the other, and a re-run after the fix makes opencode's copy a duplicate write rather than a first one.

**What #369 already did.** The last line above is new: the output names the uneven landing at the moment it happens, and `add_permission_rule`'s docstring no longer claims charter "reports it and stops". An operator is no longer misled. Nothing about the write changed.

**Why the behaviour was not changed there.** All-or-nothing means a two-phase apply — validate every harness's file, then commit — across three different file formats, with no rollback primitive and no obvious answer for a harness that fails during the commit phase rather than the validate phase. That is a design decision with real alternatives, not a defect with one right answer.

**What bounds it today.** It needs a malformed or unwritable config file to trigger at all, and the output now says what happened. The residue is a plane in a state one command did not intend.

**Reachability, honestly.** Ordinary use, once any harness's config file is malformed — which arrives from a teammate's machine or a bad merge, the same way `vaults.json` did in #363.

**Suggested direction, none chosen here.** (1) Two-phase: ask every harness to validate, and write nothing anywhere if any refuses. (2) Stay per-harness and keep the message, on the grounds that a rule under two harnesses of three is better than a rule under none. (3) Make re-running exactly idempotent and lean on that instead, so the remedy is always "fix the file and re-run" with no duplicate-write concern.

Issue #90, opened 2026-08-13
Vault guard is Bash-only: Read/Grep/Glob walk straight past it


diazoxide:
PreToolUse in `hooks/hooks.json` registers exactly two matchers:

```
matcher: Bash
matcher: Task|Agent
```

`hooks.pretooluse()` reads `tool_input["command"]` (`charter/hooks.py:478`) and `_leak_reason` walks the shlex argv. A `Read(file_path=".charter/vaults/devops.json")` carries no `command` key, matches no registered matcher, and never reaches any of it.

**Why it bites:** the bash denial *names the path it refused*. So the agent runs `cat .charter/vaults/devops.json`, reads a well-worded denial containing the path, and its obvious next move — `Read` on that same path — succeeds and dumps every plaintext credential into the transcript. The guard effectively hands over the target.

**This contradicts the docs.** `docs/secrets.md:53-54` states the guard "denies reading a vault file directly (`cat .charter/vaults/…`)". True for Bash, false for every file-reading tool the harness offers.

**Scope:** not a privilege escalation — anyone who can run the agent can already read the file. It is a hole in the one mitigation charter documents: keeping the value out of the model context and the transcript.

**Fix sketch:** add a `Read|Glob|Grep` PreToolUse matcher checking `file_path` against the vault path regex. A `permissions.deny` entry written by `init` would be belt-and-braces, but that reverses the deliberate "charter has no business touching permissions" stance at `commands.py:604` — worth discussing separately.

Found during a codebase audit; verified by hand against `hooks/hooks.json` and `charter/hooks.py`.

PR #445, opened 2026-08-24
opencode dispatches every guard charter ships (#433)

### LINKED ISSUE #433: opencode never dispatches `pretooluse-read`, so the vault-read guard is absent on that harness
Two dispatch names exist (`charter/hooks.py:3019-3020`). `pretooluse()` never looks at `tool_name` (`:1276-1285`) — it reads `tool_input["command"]` and runs the Bash guards. The file-reading guard is `pretooluse_read()` (`:1162`), gated on `_CONTENT_TOOLS = {"Read","Grep"}` (`:1156`).

Claude Code dispatches both (`hooks/hooks.json` registers `Bash → pretooluse` and `Read|Grep → pretooluse-read`), and Codex installs the same plugin (`charter/harness/codex.py:68-79`).

opencode does not. `charter/harness/opencode.py:192` builds one payload and `:206` calls one handler:

```js
const res = await $`charter hook pretooluse < ${...}`
```

`grep -rn 'pretooluse-read\|pretooluse_read' charter/harness/` → nothing. The generated `~/.config/opencode/plugin/charter.ts:48` confirms it: one call, no second.

Verified with the shim's own payload shape:
- `{"tool_name":"Read","tool_input":{"file_path":".charter/vaults/x.json"}} | charter hook pretooluse` → **empty, exit 0**
- same payload `| charter hook pretooluse-read` → `permissionDecision: "deny"`
- `{"tool_name":"Bash","tool_input":{"command":"cat .charter/vaults/x.json"}} | charter hook pretooluse` → deny

So under opencode the Bash denial fires **and names the refused path**, while the `read` tool on that same path is allowed. That is #90 verbatim (`tests/test_vault_read_guard.py:7-12`: *"the guard handed over the target"*).

`README.md:280-282` and `docs/harnesses.md:5-8` claim parity unqualified, including *"the secret-leak check"*. `OpenCodeHarness.deficits` (`opencode.py:383-390`) declares only `status-bar` and `prompt-hook`, so `charter harness list` — the mechanism `README.md:284` offers for exactly this — does not print it. `SECURITY.md:43` is the one page that scopes correctly.

Why it shipped: the shim forwards every tool to one entry point on the theory that *"every decision stays in Python, where it has tests"* (`opencode.py:176-178`), but that entry point only handles Bash — and `tests/test_vault_read_guard.py:100-113`, which pins the guard as "actually wired," reads `hooks/hooks.json` only.

#376/#407 did not cover this: `git show 062a735 -- charter/harness/opencode.py | grep -i 'pretooluse\|read'` returns nothing. #376 is about `guard ask/allow` writing operator-authored permission rules; the vault-read guard is a built-in hook.

**Fix:** route by tool in `_SHIM_TEMPLATE` (`opencode.py:167`) from a Python-side table beside `TOOL_NAMES` — `Read`/`Grep` → `pretooluse-read`, `Task` → `pretooluse-dispatch`, `Write`/`Edit` → `pretooluse-edit`, else `pretooluse`. Bump the shim stamp so `stale_wiring` (`:395`) moves existing installs.

**Verify first (UNVERIFIED):** opencode's `read` tool may name its argument `filePath` rather than `file_path`, in which case `hooks._PATH_KEYS` (`:1158`) needs it too. I could not extract the schema from the stripped binary.

**Interim, until wired:** add `Deficit("vault-read-guard", …)` to `OpenCodeHarness` so `charter harness list` and `doctor` say it, and qualify `README.md:281` / `docs/harnesses.md:6`. `SECURITY.md:43` has the wording to copy.

**Test:** `tests/test_vault_read_guard.py::TestEveryHarnessIsWired` over the **generated shim text** — assert every handler `hooks/hooks.json` dispatches is reachable from opencode.

---

diazoxide:
Closes #433

## The finding, reproduced

The generated opencode plugin forwarded **every** tool call to `charter hook pretooluse`. That handler is the *Bash* guard — it reads `tool_input["command"]` and never looks at `tool_name` — so `pretooluse_read`, the guard that refuses a file-reading tool pointed at a vault, was simply **absent** on that harness. The Bash denial still fired and still **named the path it refused**, which made `read` on that same path the obvious next move. #90 verbatim, one harness over.

Reproduced before touching anything, under `PersonaIso`, with the shim's own payload shape:

```
pretooluse-read  {"file_path": ".charter/vaults/x.json"}   -> deny
pretooluse-read  {"filePath":  ".charter/vaults/x.json"}   -> ALLOW
pretooluse       {"filePath":  ".charter/vaults/x.json"}   -> ALLOW
pretooluse       {"command": "cat .charter/vaults/x.json"} -> deny, and names the path
```

## The assessment's UNVERIFIED item is verified — and it was load-bearing

> whether opencode's `read` tool names its argument `filePath` rather than `file_path`

It does. Booted `opencode serve` from opencode 1.18.21 and read `GET /experimental/tool`:

| tool | arguments |
| --- | --- |
| `read` | `filePath`, `offset`, `limit` |
| `write` | `content`, `filePath` |
| `edit` | `filePath`, `oldString`, `newString`, `replaceAll` |
| `grep` | `pattern`, `path`, `include` |
| `task` | `description`, `prompt`, `subagent_type`, `task_id`, `command` |
| `skill` | `name` |
| `bash` | `command`, `timeout`, `workdir` |

Routing `read` to `pretooluse_read` and stopping there would have shipped a guard that ran, found no key it recognised, and allowed every real read — the exact shape of the defect it was fixing.

## The full manifest diff the coordinator asked for

`hooks/hooks.json` against what opencode dispatched, before this PR:

| event | matcher | handler | opencode, before |
| --- | --- | --- | --- |
| PreToolUse | `Bash` | `pretooluse` | ✅ (and every other tool too) |
| PreToolUse | `Read\|Grep` | `pretooluse-read` | ❌ **the finding** |
| PreToolUse | `Task\|Agent` | `pretooluse-dispatch` | ❌ |
| PreToolUse | `Write\|Edit\|MultiEdit` | `pretooluse-edit` | ❌ |
| PostToolUse | `Bash` | `posttooluse-bash` | ❌ (went to `posttooluse`) |
| PostToolUse | `Write\|Edit\|MultiEdit` | `posttooluse` | ⚠️ dispatched, but read `file_path` only — a no-op |
| PostToolUse | `Skill` | `posttooluse-skill` | ❌ never dispatched (not in `EFFECTFUL_TOOLS`) |
| PostToolUse | `Task\|Agent` | `posttooluse-dispatch` | ❌ never dispatched |
| PostToolUse | `SendMessage` | `posttooluse-message` | n/a — opencode 1.18.21 has no such tool |
| SessionStart | — | `sessionstart` | n/a — no event; charter writes `CONTEXT_PATH` instead |
| UserPromptSubmit | — | `userpromptsubmit` | n/a — the declared `prompt-hook` deficit |

Eight of eleven, three of which were security-relevant (the vault read guard; the "you just wrote a secret into committed memory" scan; the routing/edit ask).

### On #376

Checked, as asked. `git show 062a735 -- charter/harness/opencode.py` touches `_apply_rule` and `dry_run` only — zero hits for `pretooluse` or `tool.execute`. #376 made *operator-authored permission rules* all-or-nothing across harnesses; the vault-read guard is a built-in hook and was never in its path.

## The fix

**Routing.** `PRE_HOOKS`, `POST_HOOKS` and `DEFAULT_PRE_HOOK` sit beside `TOOL_NAMES` and are generated into the shim, which now calls `charter hook ${hook}`. One table, in the language that has both halves.

**Argument names.** `hooks._PATH_KEYS` gains `filePath`/`notebookPath` — **additive**, so Claude Code's spelling is untouched and a fix that merely moved the guard from one harness's spelling to the other's would fail its own test. `posttooluse` and `posttooluse_skill` read both spellings for the same reason.

**Append vs dispatch became two questions.** `EFFECTFUL_TOOLS` gated the PostToolUse *call*, which is why `skill` and `task` were never tallied. It now gates only whether charter's answer may be written into the tool's output. A `read` result is still never rewritten; neither is a skill's or a task's.

**One ceiling declared rather than left implicit.** `tool.execute.before` can allow or throw — it cannot ask. Denials are unaffected (throwing *is* denial, which is what the vault guard, the one-credential rule and the containment rule use), but charter's own tool-time asks allow and are not shown. New `ask-decisions` deficit so `charter harness list` and `doctor` say it, rather than a routed handler whose answer is silently dropped — the same failure shape as this issue, one layer down. `charter guard ask` is unaffected: those become rules in `opencode.json` and opencode prompts for them itself.

## Two existing assertions changed — reported, not relaxed

`tests/test_opencode_shim.py` asserted the literals `"charter hook pretooluse"` and `"charter hook posttooluse"` were in the shim. That was the bug written down as the contract. Both were replaced with **stronger** assertions, not weaker ones:

- the shim must interpolate the routing table (`charter hook ${hook}`), and every entry of `PRE_HOOKS`/`POST_HOOKS` must appear in the generated text;
- `pretooluse` must still appear — as `DEFAULT_PRE_HOOK`, which is the honest place for it;
- `test_running_a_post_handler_and_appending_its_answer_are_two_questions` pins the new split.

`tests/test_harness.py::test_opencode_names_the_capabilities_it_cannot_carry` gained `ask-decisions`, plus a new sibling pinning that the deficit stays narrow ("Denials are unaffected").

## New test, aimed at the class

`tests/test_opencode_dispatches_every_hook.py` reads `hooks/hooks.json` and the **generated shim text** and diffs them. Anything genuinely unreachable must be listed in `_UNREACHABLE` *with a reason*, and a stale excuse fails too. It also drives the real handlers with the payloads opencode really sends — `filePath`, `name`, `subagent_type` — rather than asserting a fixture agrees with itself.

The test that should have caught this (`TestItIsActuallyWired`) read `hooks/hooks.json` only, which is one harness's answer.

## Verification

- Full suite: `Ran 4057 tests ... OK`.
- The generated shim was `node --check`ed and then **driven end to end** in node against a stand-in for Bun's `$` — 20 assertions covering every route, the payload shape, that a `deny` on the read route throws, and that a skill/task result is never rewritten. Not part of the suite (`dependencies = []`, stdlib only), so the committed test asserts over the shim text.
- **Seven mutations, each RED then GREEN with `__pycache__` cleared between runs:** `PRE_HOOKS` emptied; `filePath` dropped from `_PATH_KEYS`; `posttooluse-skill` dropped from `POST_HOOKS`; `pretooluse-dispatch` dropped from `PRE_HOOKS`; `posttooluse` back to one spelling; `posttooluse_skill` back to one spelling; the `ask-decisions` deficit un-declared.

## One thing for the release

The shim is stamped with `__version__`, so `refresh_shim` moves an install only when the version differs. No version was bumped here (releases are cut separately) — which means **this reaches existing installs when the release that carries it bumps the version**, and not before. An operator sitting on a checkout at the same version as their installed shim keeps the old plugin until then; `charter reinit` after moving it aside is the manual path. Worth a glance at release time.


diazoxide:
Superseded by #499. Both round-four asks verified after the merge onto 0.52.0: the call sites are pinned twice — as text in `test_opencode_shim.py` so the property holds where no JS runtime is installed, and by driving the generated shim under node through charter's real handlers — and the prototype-walk class is covered by `own()` (own key + string value) with the runtime test generating its tool ids from the runtime's own `Object.prototype` rather than from the four names found by hand. 47 runtime tests ran with 0 skipped.

Merge reconciliation in three places: `SECURITY.md` keeps 0.52.0's much longer guard section with this branch's two additions folded in, and `commands_update.py`'s corrections moved into `_move_harness()` — which `main` factored out for the dev channel — so both channels get them.

diazoxide:

PR #447, opened 2026-08-24
Say what the vault guarantees, not what it does not (#444)

### LINKED ISSUE #444: Docs: the vault claim gets stronger as the audience gets less able to check it
One coordinated edit; the exact replacement wording is in the audit's section 2.

The gradient, measured:
- **`docs/secrets.md:368-371`** — correct: *"Redaction covers what comes back, not what the child does with it… `--exec` and `--stream` capture nothing by design, and therefore redact nothing."* `charter/secrets/base.py:180` likewise calls redaction *"a defence-in-depth net."*
- **`SECURITY.md:34`** — drops the qualifier: *"The model names the secret; it never sees it."* (The qualifying paragraph is two lines below, so this one is nearly fine.)
- **`README.md:201-202`** — drops it further: *"the model never sees the value"*, plus `README.md:191`'s mermaid note *"no step here ever put the value in a context window"* — true of the drawn path, while step 2 of the same diagram is the model choosing the command. `README.md:275-276` repeats it.
- **`skills/secrets/SKILL.md:54-55`** — the text loaded into the **model's** context, the one reader that cannot go check, and the strongest of the four: *"In every case the value is injected into the subprocess and **redacted from its output**, so a command that echoes it still cannot leak it into the transcript."* "In every case" is false for `--exec`/`--stream` (which this file never mentions), and "cannot leak" is false for any transform — verified: `printf %s "$T" | base64` returned the value unredacted through the default capturing path.

Also: **`docs/hooks.md:37`** states a semantic property the implementation does not have (*"A command whose argv would put a vault's contents into the transcript"*) — the guard is a 16-name allowlist. And **`docs/secrets.md:53`** says the hook *"denies `--reveal` outright"*, which is false for a wrapped invocation.

A deliberate documented trade-off is not a vulnerability, and `docs/secrets.md` gets it right. The defect is that a reader who never opens that page ends up with a materially wrong model of what the vault protects — and the model *is* such a reader.

**Fix:** apply the replacement wording for `SECURITY.md:30-34`, `README.md:191/194-195/201-202/275-276`, `skills/secrets/SKILL.md:54-55` plus two new hard rules, `docs/hooks.md:37`, and `docs/secrets.md:53`.

**Test:** `tests/test_claims.py::test_the_vault_claim_is_qualified_everywhere_it_appears` — grep for the unqualified phrases across `README.md`, `SECURITY.md`, `skills/`, `docs/`, and fail if any appears without the qualifying clause within N lines. Crude, but it is the check that would have caught this.

---

diazoxide:
Closes #444

## The finding, reproduced

charter's central claim — *"The model names the secret; it never sees it"* (SECURITY.md:34), *"the model never sees the value"* (README.md:201) — is false in a reachable default configuration, by charter's own documented commands. Redaction is `str.replace` over **captured** output, so a transform gets through the ordinary redacting path. Run in a throwaway plane under `$CHARTER_ROOT` with a fabricated value:

```
$ charter secret exec demo --env T=T -- sh -c 'echo "$T"'
***
$ charter secret exec demo --env T=T -- sh -c 'printf %s "$T" | base64'
RkFLRS1TRUtSSVQtYWJjMTIzLW5vdC1yZWFs        # the fabricated value, unredacted
```

`--exec` and `--stream` capture nothing and therefore redact nothing, which `docs/secrets.md:368-371` and `charter/secrets/base.py:180` ("a defence-in-depth net") have always said.

**The defect was never the mechanism.** It was that the strength of the claim rose as the reader's ability to check it fell — correct on the page you reach after going looking, unqualified on the front page, and strongest of all in `skills/secrets/SKILL.md`, the file loaded into the **model's** context: *"In every case the value is injected into the subprocess and redacted from its output, so a command that echoes it still cannot leak it into the transcript."*

## What changed

The audit's exact replacement wording (section 2 of the assessment), applied at every site it names, plus three more of the same class found while writing the test.

| File | Was | Is |
|---|---|---|
| `SECURITY.md:30-34` | "redacts every occurrence… it never sees it" | the promise, plus **"What that depends on, and it is not a footnote"** |
| `README.md:191` (mermaid) | "no step here ever put the value in a context window" | "charter put the value in no context window — step 2 chose the command" |
| `README.md:194-195` | "denies both that flag and `cat`-ing a vault file outright" | "Those close the accidental paths; they are not a boundary against a command chosen on purpose" |
| `README.md:201-202` | "*the model never sees the value*" | "***charter never prints the value into the conversation***. What the command you hand it to does with it is that command's business." |
| `README.md:275-276` | "the value never enters an agent's context" | "**charter never puts** the value in an agent's context or transcript — the command charter hands it to still can" |
| `skills/secrets/SKILL.md:54-55` | "in every case… still cannot leak" | net, not boundary; names `--exec`/`--stream`, which the file never mentioned |
| `skills/secrets/SKILL.md` hard rules | — | **+2**: you choose the command charter trusts; never `secret cp` to `/dev/stdout` |
| `docs/hooks.md:37` | "a command whose argv would put a vault's contents into the transcript" | a name-based check on the argv it can see |
| `docs/secrets.md:53` | hook "denies `--reveal` outright" | denies it on a charter invocation it can recognise |

Same class, found by the test rather than by the issue: `skills/browser/SKILL.md` (frontmatter description, and "it never reaches the transcript" about Playwright substitution — the other model-facing skill), `docs/mcp.md:54`, `docs/secrets.md:321`, `docs/git-policy.md:50`.

Nothing was softened back into a promise, and no feature is talked out of: the vault does exactly what it did, and the honest bits already there (`SECURITY.md:39` "not a password manager", `:43-46` "Guard rails, not guarantees") are untouched.

## The test

`tests/test_claims.py` — 8 tests, stdlib `unittest`, no plane state so no `PersonaIso` needed.

- Reads prose as **blocks, then sentences**: fenced code dropped, mermaid labels kept (the README's worst line was a diagram label), quoted spans removed so a news entry can print the sentence it retracts.
- **The unit is the sentence, not the audit's "within N lines".** The corrected paragraph six lines above README:201 would have rescued the false sentence below it — and that sentence is not unqualified, it is false.
- The rule: an absolute claim that a value does not reach a reader must **name charter as the actor**, or carry its bound in the same sentence. A redaction claim quantified over every output, or asserting a leak is impossible, fails regardless of who is named.
- `TestTheDetectorFires` feeds the **six sentences that actually shipped** back through the detector and fails if it stops flagging them, so a rule loosened until nothing trips it stops passing.

## Verification

Full suite: `Ran 4047 tests … OK`.

Mutation-tested, `__pycache__` cleared between runs — each site restored to its pre-fix wording individually, all RED, all GREEN on restore:

```
RED  README.md       *the model never sees the value*
RED  README.md       Note over M,A: no step here ever put the value in a context window
RED  README.md       the value never enters an agent's context or transcript
RED  SECURITY.md     The model names the secret; it never sees it
RED  SKILL.md        In every case … still cannot leak it into the transcript
RED  browser/SKILL   so the password never enters the conversation
RED  browser/SKILL   even though it never reached your transcript
RED  docs/mcp.md     the values never enter a context window, a transcript, or a summary
RED  test_claims.py  ← the detector itself, loosened with an over-broad qualifier
```

That last one is the anti-vacuity check working. It also caught a real one during development: `"still can"` in the qualifier list (from README's corrected *"the command charter hands it to still can"*) is a prefix of `"still cannot"`, and silently exempted the worst sentence in the repo. Removed, with a comment saying why.

No credential appears in any fixture or assertion; the reproduction ran under a throwaway `$CHARTER_ROOT` with an invented value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,9 +200,13 @@ accidental paths; they are not a boundary against a command chosen on purpose �
 ship today — `plain_file`, `reference` (point at a value that lives elsewhere) and
 **`1password`** — and more are coming. Read [docs/secrets.md](docs/secrets.md) before
 storing anything real: **`plain_file` is plaintext at mode 0600, with no encryption at
-rest.** What every provider buys you is the same and it is the point — ***charter never
-prints the value into the conversation***. What the command you hand it to does with it is
-that command's business. What only a real backend buys you is encryption. The vault is not
+rest.** What every provider buys you is the same and it is the point — on the paths that
+consume it (`secret exec`, `--dotenv`, MCP) ***charter never prints the value into the
+conversation, and everywhere else prints it only into a destination you named***. What the
+command you hand it to does with it is that command's business — and mind what you name:
+`secret cp <vault> <key> /dev/stdout` is a destination charter will take, and it puts the
+credential in this transcript ([#421](https://github.com/diazoxide/charter/issues/421),
+open). What only a real backend buys you is encryption. The vault is not
 a password manager; 1Password is, and charter will read from it.
 
 **A browser login: charter hands Playwright the password by name, so nobody types it into
@@ -276,8 +280,10 @@ you use. The browser lane additionally shells out to `npx`. That is the whole li
   and it **defaults to `local`**.
 - **Vault** — where a persona's credentials live. The provider decides the storage
   guarantee; the boundary is the same for all of them, and it is that **charter never puts
-  the value in an agent's context or transcript** — the command charter hands it to still
-  can.
+  the value in an agent's context or transcript on the paths that consume it** —
+  `secret exec`, `--dotenv`, MCP — while `secret get --reveal` and `secret cp <dest>` print
+  it into the destination you named, `/dev/stdout` included. The command charter hands it
+  to still can.
 
 ## Also in the box
 
--- a/README.md
+++ b/README.md
@@ -202,12 +202,12 @@ ship today — `plain_file`, `reference` (point at a value that lives elsewhere)
 storing anything real: **`plain_file` is plaintext at mode 0600, with no encryption at
 rest.** What every provider buys you is the same and it is the point — on the paths that
 consume it (`secret exec`, `--dotenv`, MCP) ***charter never prints the value into the
-conversation, and everywhere else prints it only into a destination you named***. What the
-command you hand it to does with it is that command's business — and mind what you name:
-`secret cp <vault> <key> /dev/stdout` is a destination charter will take, and it puts the
-credential in this transcript ([#421](https://github.com/diazoxide/charter/issues/421),
-open). What only a real backend buys you is encryption. The vault is not
-a password manager; 1Password is, and charter will read from it.
+conversation, and everywhere else prints it only where you asked for it yourself***. What
+the command you hand it to does with it is that command's business — and `secret get
+--reveal` prints to your terminal, while `secret cp` writes a real file it creates and
+refuses any destination that turns out to be one of charter's own streams, `/dev/stdout`
+included ([#449](https://github.com/diazoxide/charter/pull/449)). What only a real
+backend buys you is encryption. The vault is not a password manager; 1Password is, and charter will read from it.
 
 **A browser login: charter hands Playwright the password by name, so nobody types it into
 the conversation.** `charter browser install` generates *Playwright's own* driving pages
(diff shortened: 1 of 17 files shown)


diazoxide:
Superseded by #499. The round-four bypass is closed and re-verified after the merge: pasting `charter` into a subordinate clause, a trailing aside or a parenthetical no longer exempts a sentence, and only charter-as-subject with a bound in the same sentence passes.

Widening the corpus to 0.52.0's news entries and the two security branches merged beside this one turned up eleven more sites. Six were the detector's own fault, in three shapes — a period inside a word ending a sentence, a promise assembled across a quotation mark, and an instruction read as a promise — and all three were GREEN under mutation before `TestTheUnitIsWhatSomebodyWrote` was added for them. The other five are fixed in the prose, which is what `_QUOTED`'s docstring already prescribes.

diazoxide:

Issue #423, opened 2026-08-23
secret cp` output is outside every guard — the guard's own suggested remedy is the bypass


diazoxide:
`_VAULT_PATH_RE` (`charter/hooks.py:301`) matches only `.charter/(vaults/|browser|active-)`, and both the Bash guard (`:461`) and the Read/Grep guard (`:1192`) use it. A file written by `secret cp` to any other path is invisible to both.

Verified against the real hooks with a fabricated secret:
- `charter hook pretooluse-read` with `Read(file_path=<tmp>/leak.txt)` → exit 0, no output
- `charter hook pretooluse` with `cat <tmp>/leak.txt` → exit 0, no output

So `secret cp` + `cat` is a two-command, fully in-policy read of any vault value — and the `--reveal` denial at `hooks.py:457-459` names `cp` as the remedy.

**Fix:** record every path `secret cp` writes in a 0600 ledger under `.charter/`, and have `pretooluse`/`pretooluse-read` deny reads of any path in it.

**Test:** `tests/test_vault_read_guard.py::test_a_materialised_secret_is_covered_by_the_read_guard` — `secret cp` to a temp path, then assert both handlers deny a read of it.

---
*Filed from the 2026-08-24 security audit (30-agent adversarial review). Full assessment: `docs/superpowers/specs/2026-08-24-security-assessment.md`.*

PR #449, opened 2026-08-24
secrets: secret cp writes a real file or nothing; secret exec stops handing out other vaults' identities (#421 #422 #425 #434 #435)

### LINKED ISSUE #421: secret cp` writes plaintext to a non-regular destination — `/dev/stdout` prints the credential, then prints "Value not shown."

### LINKED ISSUE #422: secret cp` destination is unconstrained — it creates arbitrary directories, truncates any writable file, and downgrades its mode
`charter/commands_secrets.py:470-476` does no containment check:

```python
dest = Path(args.dest).expanduser()                      # :470  no check
dest.parent.mkdir(parents=True, exist_ok=True)           # :471  arbitrary dirs
fd = os.open(str(dest), O_WRONLY|O_CREAT|O_TRUNC, 0o600) # :472  clobbers
os.chmod(dest, 0o600)                                    # :475  mode-changes the victim
```

Reproduced: a pre-existing `0644` file containing `original-config` was silently replaced by the secret at `0600`. No warning, no `--force`.

charter will therefore truncate `~/.ssh/config`, a CI file, or anything else the user can write, and silently change its mode.

`docs/secrets.md:41-42` and `skills/secrets/SKILL.md:43` present `cp` as one of the two *safe* consumption paths and describe only its 0600 mode. Neither says the destination is unconstrained or that an existing file is destroyed.

**Fix:** refuse a path outside the plane (or outside an explicitly configured materialisation directory); open with `O_EXCL` by default and require `--force` to overwrite.

**Test:** `tests/test_secret_cp_destination.py::test_an_existing_file_is_not_clobbered_without_force`.

---

### LINKED ISSUE #425: secret exec` hands the child every other vault's declared identity variable
`charter/commands_secrets.py:533` is `env = dict(os.environ)`, passed unmodified to `execvpe` (`:622`), `--stream`'s `subprocess.run` (`:646`) and the capturing `subprocess.run` (`:652`). Nothing between `:533` and those calls filters anything.

Measured with fabricated env vars in a throwaway plane:

```
$ charter secret exec tmpaudit --env TOKEN=FAKE_TOKEN -- /usr/bin/env
OP_SERVICE_ACCOUNT_TOKEN=FABRICATED-op-sa-token-AAAA
ANTHROPIC_API_KEY=FABRICATED-anthropic-key-BBBB
TOKEN=***
```

The one secret the model named comes back redacted; every other credential comes back in the clear, into the caller's context. `base.redact` only knows the values it resolved this call.

Names only, from this machine's harness environment (no values read): all four `OP_*_SERVICE_ACCOUNT_TOKEN`s, `NPM_AUTH_TOKEN`, `CLAUDE_CODE_MESSAGING_TOKEN`, `SSH_AUTH_SOCK` reach the child — confirmed by name through the fabricated vault. So `charter secret exec qa …` gives its child the devops **and** marketing **and** personal 1Password service-account tokens.

`charter/secrets/base.py:66-77` sells `--token-env` as least-privilege — *"without this the mapping lives in every caller's shell… which is the property the vault abstraction otherwise removes."* This line puts it back.

Not documented anywhere: `grep -rn 'inherits.*environment|os.environ' docs/ README.md SECURITY.md` returns only unrelated hits, and `docs/secrets.md`'s "Deliberate properties" lists never mention it.

**Fix:** `env = {k: v for k, v in os.environ.items() if k not in _declared_identity_vars()}`, where `_declared_identity_vars()` collects every source name from `registry.vaults()[*]['config']['env'].values()` across all vaults, minus the ones the vault being read declares. ~8 lines, no new config. A child was never meant to hold another vault's identity, so this cannot break a working setup.

**Test:** `tests/test_secret_exec_env.py::test_another_vaults_token_env_is_not_inherited`.

---

### LINKED ISSUE #434: --stream` temp files survive SIGTERM and SIGHUP; the docs name only SIGKILL
The cleanup itself is sound — `charter/commands_secrets.py:665-667` is a single `finally` covering every `--file`/`--dotenv` path, with paths registered before the write (`:563-565`, `:605-607`). Verified clean after normal exit and after SIGINT.

But the documented limit names only SIGKILL:
- `commands_secrets.py:636-639`: *"a SIGKILLed parent runs no cleanup, so the 0600 file survives"*
- `cli.py:940-941`: same
- `tests/test_secret_stream.py:17` and `:153` (`assertIn("SIGKILL", src)`) lock the wording in

Measured, `secret exec <v> --stream --file F=K -- sh -c 'sleep 30'`, signal at t+2s:

| signal | result |
|---|---|
| SIGINT | TMPDIR empty |
| **SIGTERM** | survivor `charter-<v>-<k>-qpbw6ago`, `-rw-------`, containing the fabricated value |
| **SIGHUP** | survivor `charter-<v>-<k>-lngyve1b`, same |

Python installs no handler for SIGTERM/SIGHUP, so the default action terminates without unwinding the `finally`. SIGTERM is the *ordinary* termination — a supervisor, a `kill`, a harness killing a hung tool call — and `--stream` exists for long-running children, which are exactly what gets SIGTERMed at shutdown.

"Only if someone SIGKILLs us" reads as vanishingly rare. "Any SIGTERM" is routine.

**Fix:** around the `--stream` path, `for s in (signal.SIGTERM, signal.SIGHUP): signal.signal(s, lambda n, f: sys.exit(128 + n))`. `SystemExit` unwinds. Update the wording at `commands_secrets.py:637`, `cli.py:940`, `tests/test_secret_stream.py:17,153`.

**Test:** `tests/test_secret_stream.py::test_a_sigtermed_parent_still_cleans_up` — spawn, SIGTERM, assert TMPDIR empty. Mutation-tests cleanly by removing the handler.

---

### LINKED ISSUE #435: _safe_unlink` is a bare `os.unlink` but is called "shred" in six places
`charter/commands_secrets.py:670-674`:

```python
def _safe_unlink(path: str) -> None:
    try: os.unlink(path)
    except OSError: pass
```

No overwrite pass. Called "shred" at `charter/cli.py:936`, `charter/commands_secrets.py:509`, `:529`, `:633`, `charter/persona.py:444`, and `skills/browser/SKILL.md:62,78`.

An overwrite pass would be meaningless on APFS/ext4 copy-on-write, so **do not add one**. Fix the word.

**Fix:** replace "shred"/"shreds" with "delete"/"deletes" in all seven places.

**Test:** `tests/test_wording.py::test_no_doc_or_docstring_claims_to_shred` — grep assertion over `charter/` and `skills/`.

---

diazoxide:
The `secret cp` / `secret exec` group from the 2026-08-24 security assessment.

Closes #421
Closes #422
Closes #425
Closes #434
Closes #435

## The headline one

```
$ charter secret cp tv API_TOKEN /dev/stdout 2>&1 | cat
FABRICATED-…✓ Wrote 'tv/API_TOKEN' to /dev/stdout (0600). Value not shown.
```

The success line was false on its own output. `docs/secrets.md` said `cp` "prints only
the path, never the contents"; that was true of the command and false of the
destination, which is what actually decided where the plaintext went. After:

```
✗ Refusing to write a secret: /dev/stdout is a symlink, and charter will not follow one
  to write a credential — the link decides where the plaintext lands, not you. Name the
  real path.
```

## What changed

**`secret cp` (#421, #422).** The destination is checked *before* the vault is read, so
a refused path never resolves the value at all.

| input | before | after |
|---|---|---|
| `/dev/stdout`, `/dev/stderr`, `/dev/fd/N` | plaintext on the agent's pipe | refused |
| a FIFO | blocks, holding the resolved value | refused |
| a directory, socket, block/char device | `OSError` or worse | refused, named by kind |
| a symlink (to a file, a device, or dangling) | written through to the target | refused, and `O_NOFOLLOW` on the open |
| an existing file | `O_TRUNC` + `chmod 0600`, silent | refused; `--force` overwrites and says so |
| a missing directory tree | `mkdir -p`, arbitrary depth | one level, at 0700, else refused |
| a tracked path inside the plane | committed by the next `charter save` | refused, the rule `vault add` already applies |

The mode is now set with `fchmod` on the opened descriptor rather than `chmod` on the
path, and the default open is `O_EXCL`.

**Note where I differed from the issue.** #422's sketch says "refuse a path outside the
plane". I implemented the inverse: outside the plane is the documented case (`cp`'s own
example is a kubeconfig under `~`), and it is a path *inside* the plane that git would
track which gets the credential pushed to the forge on the next `charter save` — the
hazard `cmd_vault_add` already refuses and `cp` did not (assessment §"folds into item
8"). Same helper, same message shape.

**`secret exec` (#425).** `env = dict(os.environ)` handed the child every credential on
the machine: the one secret the model named came back `***`, every other vault's
service-account token came back in the clear. The child now gets the caller's
environment minus every identity variable declared by a vault *other* than the one being
read — both halves of each binding, since the target (`OP_SERVICE_ACCOUNT_TOKEN`) is what
the CLI authenticates with and the source is where this machine carries it. The vault
being read keeps its own names, so `charter secret exec devops -- charter secret get
devops K` still works, and `PATH`/locale/everything unrelated is untouched. Applied at
`env` construction, so `--exec`, `--stream` and the capturing path all get it.

**`--stream` cleanup (#434).** The cleanup is a `finally`, and a default-action signal
unwinds nothing — Python installs a handler for exactly one of the three (SIGINT).
Measured live, `--stream --file F=K -- sh -c 'sleep 20'`, signal at t+1s, `TMPDIR`
observable:

| | before | after |
|---|---|---|
| SIGTERM | survivor at `-rw-------` | 0 survivors, rc 143 |
| SIGHUP | survivor at `-rw-------` | 0 survivors, rc 129 |

SIGTERM and SIGHUP now raise `SystemExit(128+N)` for the duration of the command (all
modes that can hold a temp file, not just `--stream`), and the handlers are restored on
the way out so charter-as-a-library does not change how its host dies. The wording in
`commands_secrets.py`, `cli.py` and `tests/test_secret_stream.py` said SIGKILL when it
meant "any terminating signal"; SIGKILL is now the whole of the stated limit rather than
the politest third of it.

**"shred" (#435).** `_safe_unlink` is a bare `os.unlink` and should stay one — an
overwrite pass is meaningless on a copy-on-write filesystem. Six places called it
shredding; the word changed, not the code.

## Tests

`tests/test_secret_cp_destination.py` (22), `tests/test_secret_exec_env.py` (8),
`tests/test_wording.py` (3), and four cases added to `tests/test_secret_stream.py`.
Written against the class rather than the string in the issue: every non-regular kind,
all three spellings of the agent's own pipe, symlinks to files / to devices / dangling /
under `--force`, and the two plants that only `O_EXCL` and `O_NOFOLLOW` can refuse (the
check is stubbed out so the *open* has to answer, which a single-threaded test can
otherwise never reach).

No real credential appears in any fixture or assertion message: the exec test's probe
prints variable NAMES only, and every comparison is a filtered projection rather than a
whole environment.

Full suite: `Ran 4153 tests … OK`.

**Mutation-tested, 14 for 14 RED**, `__pycache__` cleared between runs: non-regular arm
removed · symlink arm removed · exists-check removed · `O_EXCL`→`O_TRUNC` ·
`O_NOFOLLOW` dropped · git-containment refusal removed · `mkdir` one level→`makedirs` ·
value resolved before the check · env filter removed · only the source half of a binding
collected · malformed-registry guard removed · termination handlers not installed ·
handlers never restored · one "shreds" put back.

The SIGTERM cases signal the test process itself, so the class installs a recording
no-op handler first: with the fix charter's handler wins, and without it the case fails
on its assertions instead of killing the runner.

## Not in scope here

`hooks.py` is untouched — the `secret cp` read-guard ledger and the `--reveal` denial
text that names `cp` are #423's, and belong to whoever holds that file.

### DIFF
--- a/charter/cli.py
+++ b/charter/cli.py
@@ -943,8 +943,10 @@ def _sa_exec(p):
                         "GOOGLE_APPLICATION_CREDENTIALS takes a path, not a value) — the "
                         "case --exec cannot serve, because exec leaves nothing alive to "
                         "clean up. Output is NOT redacted (nothing is captured). Cleanup "
-                        "survives SIGINT, SIGTERM and SIGHUP; only a SIGKILLed charter "
-                        "runs no cleanup, and then the 0600 file survives.")
+                        "survives every terminating signal charter may catch — SIGINT, "
+                        "SIGTERM, SIGHUP, SIGQUIT and the rest. It does not survive "
+                        "SIGKILL, which cannot be caught, or a fault (SIGSEGV, SIGABRT), "
+                        "which charter does not intercept; then the 0600 file survives.")
     p.add_argument("--exec", dest="exec_mode", action="store_true",
                    help="Replace this process with the command (os.exec) instead of capturing "
                         "it, so stdio streams through — required for a long-running child such "
--- a/charter/cli.py
+++ b/charter/cli.py
@@ -918,7 +918,12 @@ def _sa_key(p):
 
 def _sa_cp(p):
     p.add_argument("key")
-    p.add_argument("dest")
+    p.add_argument("dest",
+                   help="Path of a REAL FILE to create. A device, FIFO, directory or "
+                        "symlink is refused — /dev/stdout is this conversation.")
+    p.add_argument("--force", action="store_true",
+                   help="Overwrite an existing file (destroying its contents and "
+                        "setting it to 0600). Refused without this.")
 
 
 def _sa_exec(p):
@@ -933,12 +938,13 @@ def _sa_exec(p):
                         "dotenv secrets file, e.g. PLAYWRIGHT_MCP_SECRETS_FILE.")
     p.add_argument("--stream", dest="stream_mode", action="store_true",
                    help="Run the command as a CHILD with stdio inherited, wait for it, then "
-                        "shred any --file/--dotenv temp files. For a long-running child "
+                        "delete any --file/--dotenv temp files. For a long-running child "
                         "whose credential must be a FILE (Google ADC's "
                         "GOOGLE_APPLICATION_CREDENTIALS takes a path, not a value) — the "
                         "case --exec cannot serve, because exec leaves nothing alive to "
-                        "clean up. Output is NOT redacted (nothing is captured). Note: a "
-                        "SIGKILLed charter runs no cleanup and the 0600 file survives.")
+                        "clean up. Output is NOT redacted (nothing is captured). Cleanup "
+                        "survives SIGINT, SIGTERM and SIGHUP; only a SIGKILLed charter "
+                        "runs no cleanup, and then the 0600 file survives.")
     p.add_argument("--exec", dest="exec_mode", action="store_true",
                    help="Replace this process with the command (os.exec) instead of capturing "
                         "it, so stdio streams through — required for a long-running child such "
--- a/charter/commands_secrets.py
+++ b/charter/commands_secrets.py
@@ -477,6 +477,86 @@ _DEST_KINDS = ((stat.S_ISDIR, "a directory"), (stat.S_ISFIFO, "a FIFO"),
                (stat.S_ISSOCK, "a socket"), (stat.S_ISCHR, "a character device"),
                (stat.S_ISBLK, "a block device"))
 
+#: charter's own three streams, by descriptor, for the refusal message.
+_OWN_STREAMS = ((0, "standard input"), (1, "standard output"), (2, "standard error"))
+
+
+def _own_stream_identities() -> dict[tuple[int, int], str]:
+    """``(st_dev, st_ino) -> "standard output"`` for this process's own three streams.
+
+    IDENTITY, not name. The first version of this guard asked what a path was *called*
+    and `/dev/fd/1` walked straight through it: on macOS `/dev/fd/N` is neither a symlink
+    nor a device but the underlying object re-opened, so `os.lstat` reported a plain
+    regular file and the "already exists" arm took over — an arm `--force` switches off.
+    `charter secret cp v k /dev/fd/1 --force` then wrote the credential into charter's own
+    captured stdout and printed "Value not shown." on top of it. That is issue #421's
+    symptom with a different spelling, and no list of spellings closes it: `/dev/stdout`,
+    `/dev/fd/1`, `/proc/self/fd/1`, the path `readlink` gives for the transcript log, and
+    any hardlink to it are five names for one inode.
+
+    So the question asked here is "is this object the same object as one of my streams",
+    which has exactly one answer per object however it is spelled.
+
+    A stream charter cannot stat (a closed descriptor) contributes nothing rather than
+    raising: the caller's other arms still apply.
+    """
+    out: dict[tuple[int, int], str] = {}
+    for fd, name in _OWN_STREAMS:
+        try:
+            st = os.fstat(fd)
+        except OSError:
+            continue
+        out.setdefault((st.st_dev, st.st_ino), name)
+    return out
+
+
+def _own_stream_refusal(dest, which: str) -> str:
+    """The refusal for "that destination IS one of charter's own streams".
+
+    Deliberately names no flag. The refusal this replaced ended "Pass --force to
+    overwrite it deliberately" — and `--force` was the bypass, which is the exact
+    pattern #421 filed against hooks.py's `--reveal` text. A guard that prints its own
+    way around itself is a signpost, not a guard.
+    """
+    return (f"{dest} is charter's own {which} — the channel this conversation is read "
+            f"from, whatever it is called here. Writing a credential to it puts the "
+            f"plaintext straight into the transcript, which is the leak `secret cp` "
+            f"exists to avoid. Name a real file that is not one of these streams.")
+
+
+def _identify_dest(raw: str):
+    """`fstat` of *raw* opened for writing without creating or truncating anything.
+
+    `os.lstat` is not enough: on macOS it answers about the devfs entry for `/dev/fd/N`,
+    whose `st_dev` is devfs's and NOT the underlying file's, so a `(st_dev, st_ino)`
+    comparison against `os.fstat(1)` does not match. Measured on this platform::
+
+        lstat /dev/fd/1  dev=18446744071623019954 ino=202112957
+        fstat(1)         dev=16777234             ino=202112957
+        fstat(open('/dev/fd/1', O_WRONLY))  dev=16777234 ino=202112957
+
+    Only the descriptor tells the truth, so this opens one. The open is deliberately
+    harmless — no ``O_CREAT``, no ``O_TRUNC``, ``O_NOFOLLOW`` so a symlink is not
+    traversed, ``O_NONBLOCK`` so a FIFO planted since the `lstat` cannot block the
+    process with the answer half-computed. Callers must have refused every non-regular
+    kind *before* calling: opening a device is not free of side effects, and this must
+    never be the thing that opens one.
+
+    Returns ``None`` when the destination cannot be opened at all (it does not exist, or
+    is not writable) — nothing to compare, and the real open below will say so.
+    """
+    flags = os.O_WRONLY | getattr(os, "O_NOFOLLOW", 0) | getattr(os, "O_NONBLOCK", 0)
+    try:
+        fd = os.open(raw, flags)
+    except OSError:
+        return None
+    try:
+        return os.fstat(fd)
+    except OSError:
+        return None
+    finally:
+        os.close(fd)
+
 
 def _cp_dest_refusal(dest: Path, force: bool) -> str | None:
     """Why *dest* is not somewhere a secret may be materialised — or ``None``.
@@ -501,7 +581,9 @@ def _cp_dest_refusal(dest: Path, force: bool) -> str | None:
       now `--force`, and the default is ``O_EXCL``.
 
     Ordered so the *most specific* thing true of the path is what gets said: a symlink to
-    a character device reads as a symlink, not as a device.
+    a character device reads as a symlink, not as a device. The stream-identity arm sits
+    AHEAD of the "already exists" arm on purpose — the exists arm names `--force`, and
+    `--force` must never be the suggested next step for one of charter's own streams.
     """
     raw = str(dest)
     if not raw:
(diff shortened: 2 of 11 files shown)


diazoxide:
## Round two — both blocking findings closed

### 1. BYPASS: `secret cp <vault> <key> /dev/fd/1 --force` — **closed**

Reproduced first, against this branch, in an isolated plane with charter's stdout on a
regular file (this harness's own condition):

```
rc=0
stdout file: FABRICATED-probe-value-zzz
stderr: ! Overwrote /dev/fd/1 and set it to 0600.
        ✓ Wrote 'v/k' to /dev/fd/1 (0600). Value not shown.
fd 1 mode after: 0o100600      # the fchmod landed on the transcript
```

Without `--force` it printed `… Pass --force to overwrite it deliberately` — the guard
naming its own bypass, the #421 pattern.

Root cause was the name/identity mistake: `_cp_dest_refusal` asked `os.lstat` about the
*path*, and on macOS `/dev/fd/N` is the underlying object, not a symlink and not a
device. It read as an ordinary existing file and landed in the one arm `--force`
switches off.

Fixed by identity, `(st_dev, st_ino)` against `os.fstat(0/1/2)`. Worth noting for anyone
implementing the same check elsewhere: **`lstat` on the path is not sufficient**, measured
on this platform —

```
lstat /dev/fd/1                      dev=18446744071623019954 ino=202112957
fstat(1)                             dev=16777234             ino=202112957
fstat(open('/dev/fd/1', O_WRONLY))   dev=16777234             ino=202112957
```

devfs reports its own `st_dev`. Only a descriptor answers correctly, so the check opens
one — non-destructively (`O_WRONLY|O_NOFOLLOW|O_NONBLOCK`, no `O_CREAT`, no `O_TRUNC`),
and only after the path has already been refused for kind, so it never opens a device.
The `lstat` pair is kept alongside it because it catches a second *name* for the inode
(a hardlink, a `readlink`'d log path) without opening anything.

Both sit ahead of the exists/`--force` arm, so no stream refusal can suggest `--force`,
and the message names no flag at all. The real open is `fstat`ed too — before the vault
is read, before anything is written — which is the half a race cannot beat. `O_TRUNC` is
gone from the open (it is an `ftruncate` after identification), `fchmod` moved after it,
and the abandon path unlinks only what the call itself created.

After:

```
rc=2
stdout file: (empty)
stderr: ✗ Refusing to write a secret: /dev/fd/1 is charter's own standard output — the
        channel this conversation is read from, whatever it is called here. …
fd 1 mode after: 0o100644      # unchanged
```

Same answer for the transcript's real path, for a hardlink to it, for `/dev/fd/2`, with
and without `--force`, and with the path check stubbed out entirely.

**@#447 — this is closed, so "charter never prints the value into the conversation" is
true again as far as `secret cp` is concerned. The two can land together.**

### 2. OVERCLAIM: SIGQUIT leaves the 0600 file — **fixed by widening**

Reproduced as a failing test before changing anything (SIGQUIT and SIGUSR1 both
terminated without unwinding). Rather than adding two more names to a list, the handled
set is now derived by exclusion from `signal.Signals`: minus what cannot be caught
(SIGKILL, SIGSTOP), what does not terminate (SIGCHLD, SIGCONT, SIGURG, SIGWINCH), what
suspends the process (SIGTSTP, SIGTTIN, SIGTTOU — Ctrl-Z must background a command, not
kill it), SIGPIPE and SIGINT (already handled by Python), and the faults.

Both wrong versions of that sentence were wrong the same way, so the stated limit is now
a category rather than examples: SIGKILL, which nothing can catch, and a fault (SIGSEGV,
SIGBUS, SIGABRT), which charter deliberately does not intercept — a handler on an
already-broken process can turn a crash into a hang. Updated in the `--stream` help, the
code comment and the news entry, and pinned by tests on SIGQUIT, on SIGUSR1, on the
derived set's contents and exclusions, and on the help text naming both SIGQUIT and the
faults.

### Mutation testing

Eight mutants, each killed by a named case:

| mutant | killed by |
|---|---|
| drop pre-open identity arm | `test_a_name_is_recognised_without_opening_anything`, `test_the_refusal_never_names_a_flag_that_would_get_past_it` ×2 |
| drop post-open identity check | `test_the_open_refuses_even_when_the_path_check_is_bypassed` ×2 |
| probe open returns nothing | `test_the_refusal_never_names_a_flag_that_would_get_past_it` (`/dev/fd/1`) |
| `O_TRUNC` back at open | `test_the_open_refuses_even_when_the_path_check_is_bypassed` (real path) |
| unlink unconditionally | `test_the_open_refuses_even_when_the_path_check_is_bypassed` (real path) |
| drop post-open kind check | `test_a_device_that_appears_after_the_check_is_refused_by_the_descriptor`, `test_a_fifo_with_a_reader_is_refused_by_the_descriptor` |
| signal set back to two | `test_a_sigquit_parent_still_cleans_up`, `test_a_sigusr1_parent_still_cleans_up`, `test_the_catchable_terminating_signals_are_all_handled`, `test_the_handlers_are_removed_again_afterwards` |
| revert the `--stream` help | `test_the_help_does_not_blame_sigkill_for_what_sigquit_does` |

One mutant started out **surviving** and is worth recording: `O_TRUNC` at open lived
while the only race case used `/dev/fd/1`, because macOS devfs silently ignores
truncation through that name — the file behind it was untouched and the test passed. The
case now runs both spellings, and the real path kills it.

Rebased onto `origin/main` (`a1b8905`). Full suite: `Ran 4198 tests … OK`.

PR #451, opened 2026-08-24
Close five parsing bypasses in the vault and git guards (#429 #430 #431 #423 #438)

### LINKED ISSUE #423: secret cp` output is outside every guard — the guard's own suggested remedy is the bypass

### LINKED ISSUE #429: _segment_argv`'s unparseable-quote fallback silently drops every guard after the first invocation
`charter/hooks.py:637-642` — on `shlex.ValueError`, `_segment_argv` returns `[(cmd or "").split()]`: the whole command as one whitespace-split segment. Every guard then reads only `toks[0]` as the program, so any invocation after the first is invisible.

The docstring at `:626-628` asserts the opposite: *"the leak guard scans the entire text and stays **fail-closed** — not printing a secret is a safety invariant."* It does not scan text; `_leak_reason` iterates invocations (`:452-463`).

`echo $'it\'s fine' ; cat .charter/vaults/x.json` is valid bash (`bash -n -c` → rc=0) and trips shlex. Against the shipped hook:

```
$ charter hook pretooluse --plugin-version 0.51.0 < payload.json
(exit 0, no output)          ← ALLOW
$ bash -c "$CMD"
it's fine
{"K":"FABRICATED-VALUE-777"}  ← vault plaintext into the transcript
```

The same prefix flips three more guards from DENY to ALLOW: `git clone git@github.com:o/r.git`, `git commit -S -m x`, `GIT_SSH_COMMAND=x git fetch`, and (under `bypassPermissions`) `git tag v9.9.9`.

The regression test that names this property is scoped so it cannot see it: `tests/test_guard_parsing.py:121-127` uses `"charter secret get v k --reveal 'unbalanced"` — the offending program is token 0, the one arrangement where the collapse is harmless. It passes while the property it asserts is false.

`SECURITY.md:57-59` puts this in scope: *"anything that bypasses the vault guard or the git policy guard."*

**Fix:** re-segment the whitespace split on `_OPERATORS` so later invocations stay visible. For the leak guard specifically, on the unparseable path also run a raw substring scan for `--reveal` and `_VAULT_PATH_RE` over the whole string — which is what the docstring already claims happens. Correct the docstring.

**Test:** rescope `tests/test_guard_parsing.py:121-127` and add `test_an_unparseable_quote_does_not_hide_a_later_invocation` with the bash-valid case plus the three git variants.

---

### LINKED ISSUE #430: Guards take `prog` from token 0, so a wrapper prefix or shell keyword hides the real program
`_split_env` (`charter/hooks.py:655-661`) strips only `VAR=value` prefixes, and `_OPERATORS` (`:602`) contains no grouping tokens. So `prog` is `env`/`command`/`then`/`(`, never the real program, and neither `_is_charter` (`:311-319`) nor the `_READERS` test (`:460`) matches.

Verified ALLOW from the real functions, DENY for the unwrapped form in each case:

```
env charter secret get v k --reveal --force     (and it printed the value)
/usr/bin/env charter secret get v k --reveal --force
command cat .charter/vaults/x.json              (live: printed the file)
time / nohup / exec / sudo cat .charter/vaults/x.json
{ cat .charter/vaults/x.json; }   ( cat .charter/vaults/x.json )
echo $(cat .charter/vaults/x.json)
if true; then cat .charter/vaults/x.json; fi
cd .charter/vaults && cat x.json
```

`_single_credential_hit` (`:890`) has the same shape and **no second gate**:

```
GIT_SSH_COMMAND=/tmp/k git push        → HIT
env GIT_SSH_COMMAND=/tmp/k git push    → None
git push git@github.com:o/r.git        → HIT
/usr/bin/env git push git@…            → None
```

That half is the sharper one: unlike `--reveal`, nothing downstream re-checks, so an injected model can route git through an attacker-supplied SSH transport — the thing golden rule 0 exists to stop.

Internal inconsistency: `hooks.py:793-801` explicitly follows a `cd` in an earlier segment for the plane-root guard (*"A `cd` earlier in the SAME command moves where the later segments run"*). `_leak_reason` does not.

The project has accepted this argument before — `charter/cli.py:37-40`, on `--rev` abbreviating to `--reveal`: *"A guard that a three-character abbreviation walks past is not a guard."*

**Fix:** add `(`, `)`, `{`, `}` to `_OPERATORS`; strip a leading run of `env`, `command`, `builtin`, `exec`, `nice`, `nohup`, `stdbuf`, `timeout`, `sudo`, `doas`, `if`, `then`, `else`, `elif`, `do`, `while`, `until`, `for`, `!` before taking `prog`. Fixes A2/A3/A4 at the same time.

**Not in scope for this issue:** re-parsing `sh -c '<string>'`. That is pinned as a deliberate limit at `tests/test_leak_guard_readers_that_write.py:105-113` and should be documented rather than changed here.

**Test:** `tests/test_guard_parsing.py::test_a_wrapper_prefix_does_not_hide_the_program` — table of the ten commands above, each asserted DENY.

---

### LINKED ISSUE #431: _VAULT_PATH_RE` is matched without normalizing `//`, `/./` or case
`charter/hooks.py:301` is a literal, case-sensitive substring match, used by both the Bash guard (`:461`) and the Read/Grep guard (`:1191-1192`).

Bash side, verified live with a fabricated vault:
- `cat .charter/vaults/x.json` → DENY
- `cat .charter//vaults/x.json` → ALLOW, printed the value
- `cat .charter/./vaults/x.json` → ALLOW, printed the value

Read side: Claude Code normalises `tool_input.file_path` before the hook, so `//`, `/./` and `x/../` are all correctly denied through the real Read tool. **But case variance survives normalisation** — Read of `.Charter/vaults/x.json` returned the fabricated plaintext, because the regex is case-sensitive and APFS is not. (Would not resolve on case-sensitive ext4.)

charter already knows the normalisation idiom and applies it 900 lines later: `hooks.py:2089`, `norm = ("/" + fp.replace("\\","/")).replace("//","/")`. The two vault guards, which `hooks.py:1170-1173` argues must never disagree, both skip it.

**Fix:** collapse `/+` and `/./` and fold case before `.search` at `hooks.py:461` and `hooks.py:1191`.

**Test:** `tests/test_vault_read_guard.py::test_path_spelling_variants_are_denied` — `//`, `/./`, `/x/../`, `.CHARTER/`, through both handlers.

---

### LINKED ISSUE #438: pretooluse_read` swallows its own deny in a bare `except Exception
`charter/hooks.py:1200-1201` — `except Exception: return 0` wraps the entire body **including** the `_deny(...)` call. A `BrokenPipeError` from `_deny`'s `print` is an `Exception` and would silently allow.

`pretooluse` (Bash) deliberately has no such wrapper, so the two sibling vault guards fail in opposite directions — in a module that argues at `:1170-1173` that they must never disagree.

Reachability is low (the body is dict gets, `str()` and a regex), but the direction is wrong.

**Fix:** narrow the `except` to wrap only the parsing, leaving `_deny` outside it.

**Test:** `tests/test_vault_read_guard.py::test_a_broken_pipe_does_not_turn_a_deny_into_an_allow` — patch `print` to raise, assert non-zero exit rather than 0.

---

diazoxide:
Five confirmed SECURITY findings from the 2026-08-24 audit, all in `charter/hooks.py`
and all the same shape: **the guards are bypassed by parsing, not by cleverness.** Every
guard reads token 0 of a segment as the program, and both the segmenting and the path
matching had holes. Each case below was run against the shipped hook with a fabricated
vault and printed the value.

Closes #429
Closes #430
Closes #431
Closes #423
Closes #438

## What was wrong, and what changed

**#429 — an unparseable quote dropped every guard after the first invocation.**
`_segment_argv`'s `shlex.ValueError` fallback returned the whole command as ONE
whitespace-split segment, so only the first program existed.
`echo $'it\'s fine' ; cat .charter/vaults/x.json` is valid bash (`bash -n` rc=0), trips
shlex, and flipped the leak guard, `git clone git@…`, `git commit -S`,
`GIT_SSH_COMMAND=… git fetch` and an unattended `git tag` from deny to allow. The
docstring asserted the opposite — that this path kept the leak guard *fail-closed*.

The fallback now re-segments on the operators, including operators glued to their
neighbours (`echo 'x;cat …` is two commands to a shell and one token to `str.split`), and
`_leak_reason` adds a raw scan of the string when parsing failed — which is what the
docstring already promised. `_plane_root_branch_reason` opts out of the fallback in one
explicit line and keeps failing OPEN; it is the guard whose failure mode is annoyance, and
a phantom `git checkout` conjured out of a broken quote must not stop a turn.

**#430 — one word in front of the program hid it.** `env`, `command`, `sudo`, `time`,
`nohup`, `exec`, `xargs`, `{ … }`, `( … )`, `$( … )`, `if …; then …; fi`, and a `cd` in an
earlier segment. `_split_env` now strips the wrapper run before naming the program, so all
four guards get one answer from one place — including the sharper half, where
`env GIT_SSH_COMMAND=/tmp/k git push` walked past the one-credential rule with nothing
downstream to re-check it. The value-flag table is **per wrapper** (`env -i` takes no
value; `xargs -I` does), because a flat table swallows the program on `env -i cat <vault>`
— fail-open on the exact input the table exists to catch. `env -S '<cmd>'` is expanded
rather than skipped, for the same reason. `sh -c '<string>'` stays out of scope and is now
pinned as a limit rather than left ambiguous.

**#431 — `//`, `/./` and case were three different paths to the guard and one file to the
filesystem.** `_VAULT_PATH_RE` was a literal, case-sensitive substring match. The first two
were allowed through Bash; `.Charter/vaults/x.json` was allowed through **both** guards,
because APFS opens it and the regex did not. One predicate, `_vault_path_hit`, now answers
for the Bash guard and the Read/Grep guard — this module has argued for a while that the
two must never disagree while each open-coded the match. It also covers the bare
`.charter/vaults` directory, which the Read guard handled and the Bash guard did not, and
`_is_charter` is folded the same way (`CHARTER secret get --reveal` runs the same binary).

**#423 — `secret cp` wrote plaintext where no guard was looking, and the denial pointed at
it.** `charter secret cp v k /tmp/x && cat /tmp/x` was a two-command, fully in-policy read
of any vault value, and the `--reveal` refusal said *"Use `charter … secret exec`/`cp`"*.
New `charter/materialized.py` records every destination (0600; paths and key names, never
values, self-pruning on write) and both guards refuse to read anything in it — by absolute
path, or relative to the session cwd or an earlier `cd`. **The denial text is fixed as well
as the hole:** neither reason names `cp` as an alternative to anything any more.

**#438 — the read guard swallowed its own deny.** `pretooluse_read` wrapped its whole body
in `except Exception: return 0`, `_deny` included, so a `BrokenPipeError` from the denial's
own `print` returned an allow — while the Bash sibling has no such wrapper, leaving the two
vault guards failing in opposite directions. Only the parse is inside the handler now; the
trace keeps its own, so a full disk cannot un-deny a read.

## Verification

`python3 -m unittest discover -s tests` → **Ran 4074 tests, OK** (4039 before; +35).

Every guard added was mutation-tested, `__pycache__` cleared between runs, RED confirmed on
the named tests and GREEN on restore:

| mutation | RED |
| --- | --- |
| fallback returns one segment again | `test_an_unparseable_quote_does_not_hide_a_later_invocation` (3 git variants), 2 tokenizer tests |
| `_split_env` stops stripping wrappers | 36 failures across `TestAWrapperDoesNotHideTheProgram` + `TestTheProgramIsWhatRuns` |
| `(`/`)`/`{`/`}` out of `_OPERATORS` | `$(cat <vault>)`, `test_grouping_tokens_end_a_segment` |
| `(?i)` off `_VAULT_PATH_RE` | `.CHARTER/…` and `.Charter/…` through both handlers |
| `_norm_path_text` returns its input | 10 `//` and `/./` cases through both handlers |
| `..` resolution disabled | `test_a_traversal_INTO_the_vault_is_resolved` |
| `materialized.covers` returns False | 3 materialised-secret tests |
| `cmd_secret_cp` stops recording | same 3, plus the ledger test |
| deny moved back inside the blanket `except` | `test_a_broken_pipe_does_not_turn_a_deny_into_an_allow` |
| deny/trace order swapped | `test_a_failing_trace_still_leaves_the_deny_standing` |
| raw scan removed from the unparseable path | the two flag-value/`--reveal` cases |
| directory-slash append removed | `grep -r . .charter/vaults` through both handlers |
| `_is_charter` case-sensitive again | `test_the_program_name_is_case_folded_too` |

No existing test was weakened. `test_the_leak_guard_still_fails_CLOSED` was **rescoped**:
its fixture put the offending program at token 0 — the one arrangement where collapsing to
a single segment is harmless — so it passed green for as long as the property it named was
false. The original spelling is kept as a second assertion. No fixture holds a real
credential, and no assertion prints one.

## Deliberate non-changes

* `sh -c '<string>'` is not re-parsed (out of scope per #430, now pinned by a test).
* A broad search rooted at a directory that happens to contain a materialised file is not
  refused — the same posture the vault guard already documents for `.charter/vaults/`.
* A file `secret cp` wrote and something else later moved is not tracked; covering that
  needs content scanning on every read.


diazoxide:
## Round two — the three findings, reproduced then closed

All three reproduced first, through `python3 -m charter hook pretooluse` with a fabricated
vault, and were re-run after the fix. Rebased onto current `origin/main` (a1b8905).

### 1 · REGRESSION — `(`/`)` in `_OPERATORS` stranded the operand

Confirmed: DENY on `origin/main`, ALLOW on round one, for `cat $(echo .charter/vaults/x.json)`,
`head $(ls …)`, `tail -n 5 $(echo …)`, `grep -q secret $(echo …)` and
`git push $(echo git@github.com:o/r.git)`.

Not a revert — the parens are load-bearing the other way. A `$( … )` run now yields an
**additional inner segment while the enclosing segment keeps accumulating**, so
`echo $(cat <vault>)` (substitution is the reader) and `cat $(echo <vault>)` (substitution
is the operand) are both seen. Segmenting is one function, `_segment_tokens`, shared by the
parsed and the unparseable paths.

Next spellings, each verified ALLOW before the fix and denied now: backticks (normalised to
`$( … )` before tokenizing — `cat \`echo .charter/vaults/x.json\``), process substitution
(`cat <(cat <vault>)`), a substitution spliced into a longer word
(`cat $(echo .charter)/vaults/x.json`, whose halves shlex hands back as separate operands),
and `( true );cat <vault>` — shlex emits `);` as ONE token, which matched no operator, so
the `cat` was swallowed into the argv before it.

### 2 · Wrapper chdir, and `pushd`

`env -C`, `env --chdir=`, `sudo -D`, `sudo --chdir=` and `pushd` all reproduced as ALLOW.
The flag is now read **once**, giving both its name (does the next token start the program)
and its value (where the program runs) — two readings is how the value came to be dropped.
The value feeds the same `here` the `cd` branch sets. `pushd` counts as `cd`; `popd`
deliberately does not, because forgetting `here` there is the fail-open direction. A wrapper
chdir applies to its own program and does **not** move the shell for later segments
(pinned: `env -C .charter/vaults ls && cat x.json` stays allowed).

### 3 · The case fold that landed everywhere except two places

`_single_credential_hit` and `_release_floor_reason` each kept their own
`prog.rsplit("/", 1)[-1]`. `GIT push git@github.com:o/r.git`,
`env GIT_SSH_COMMAND=/tmp/k GIT push`, `/usr/bin/GIT clone …` and `GIT tag v1` were allowed
while the lowercase spellings denied. Both ask `os.path.basename(prog).lower()` now.

### Text corrected

The news entry claimed the program-name fold was everywhere and implied the substitution
class was covered; `docs/hooks.md` listed the argv invariants without them. Both fixed —
correcting text a reviewer proved false is part of the fix. No denial message in this group
names a route around itself (checked: `_REVEAL_REASON`, `_READ_REASON`, the materialized
refusal, the one-credential prose, the release floor).

### Verification

Suite: **4198 tests, OK** on top of `origin/main`.

Eleven mutants, each RED on the named tests and GREEN on restore, `__pycache__` cleared
between runs: `(` back to a plain boundary · substitution emits no inner segment · outer
segment does not accumulate across one · no punctuation-run split · no backtick
normalisation · chdir value discarded again · `pushd` off `_CHDIR_BUILTINS` · chdir made
sticky past its segment · case fold removed from the one-credential guard · same from the
release floor · spliced operands not re-joined.

### Still open (not a regression — allowed on `main` too)

`xargs -a .charter/vaults/x.json echo` prints the vault. `xargs` is a wrapper, so `-a`'s
value is consumed as a wrapper flag value and never reaches the operand check. It is a
different family from the three findings (a wrapper flag whose value names a file to *read*,
not a directory), and closing it means widening the `_split_env` contract again — better as
its own change than smuggled into a parsing fix.

diazoxide:
Superseded by #499. Three of the five issues here are merged onto 0.52.0; two are not, and that is the report rather than an omission.

**#431 is refuted against current `main`.** `_names_a_vault_path` (#462) already unions `os.path.normpath(operand)` with the raw operand and `_VAULT_PATH_RE` already carries `IGNORECASE`, and both guards already call that one predicate. This branch's `_vault_path_hit` is the older, weaker pattern — no `.edm`, no `fingerprint.key`, no bare `.charter`, `vaults` unanchored to a path segment — so merging it would have been the (B) case. It is dropped; every call site added here asks `_names_a_vault_path`.

**#423 and #438 were closed on `main` in 0.52.0.** `charter/materialized.py` is dropped with it: `main` rejected that ledger explicitly, in `_leak_reason`'s own docstring, because it matches a spelling.

**The differential corpus was stale.** It was recorded against 060d584; re-recording against 18a8afc added or widened 798 of 2,444 rows, so it would have passed on 1,646 rows while saying nothing about the rest. Re-recorded, and re-verified: 0 of the 834 inputs current `main` denies are allowed here.

diazoxide:

PR #462, opened 2026-08-24
The vault guards say what they catch, and the denial stops naming the bypass (#423)

### LINKED ISSUE #423: secret cp` output is outside every guard — the guard's own suggested remedy is the bypass

diazoxide:
Closes #423

The 2026-08-24 review's **DOCUMENT THE LIMIT** tier (items 21–25) plus the message half of
#423. Four rounds of adversarial review established these cannot be closed in code — every
added parser lost to a new spelling and one attempt shipped a regression — so this PR closes
none of them in code. It fixes the sentences that claimed otherwise, and pins the behaviour
so the sentences cannot drift back.

## The model-facing text was the worst of it

`skills/secrets/SKILL.md` told the agent — the one reader who cannot go and check — that the
value is *"redacted from its output, so a command that echoes it still cannot leak it into
the transcript."* Redaction is `str.replace` over the bytes that came back
(`charter/secrets/base.py:177-186`). A transform beats it through the **ordinary, redacting**
path:

```
charter secret exec v --env T=K -- sh -c 'printf %s "$T" | base64'   # value, in full
```

and `--exec`/`--stream` capture nothing and therefore redact nothing, which that file never
mentioned. It now says *net, not boundary*, and adds two hard rules an agent can act on: you
choose the command and charter trusts your choice; `secret cp` hands a **path** to a tool
that needs one and is not a way to get at the value.

## The same claim, weaker, in four more places

`SECURITY.md`, `README.md`, `docs/hooks.md`, `docs/secrets.md` — including the mermaid note
reading *"no step here ever put the value in a context window"* sitting directly under the
step where the model picks the command. Each now separates the guarantee (**charter never
prints the value into the conversation**) from what depends on the command chosen, and states
the guard's ceiling in concrete spellings rather than implying completeness:

| Limit | What now says so |
|---|---|
| 21 · redaction is a substring net | `SECURITY.md`, `README.md`, `SKILL.md`, `docs/secrets.md` (top of doc *and* the reference section) |
| 22 · `--exec`/`--stream` redact nothing | same four; already correct in `cli.py` help, now pinned by test |
| 23 · `_READERS` is a name allowlist with a ceiling | `SECURITY.md`, `docs/hooks.md`, and a `#:` block at the definition explaining why it is not widened |
| 24 · `sh -c '<string>'` is not re-parsed | `docs/hooks.md`, `docs/secrets.md` — was only a test docstring |
| 25 · a vault outside `.charter/` is unguarded | `docs/secrets.md`, next to the `--file` remedy that creates the situation |

All verified against the shipped code first — `python3 -c`, `base64`, `cp`, `jq`, `cut`,
`dd`, `tr <`, `git show HEAD:<path>` and `sh -c 'cat …'` are each `allow` today, while `cat`
on a vault file denies.

## #423: the `cp` hole is fixed; the message was not

The `/dev/stdout` half landed in #449. What remains true is the two-step: `secret cp <v> <k>
/tmp/x` then `cat /tmp/x` — both allowed — and the `--reveal` denial read *"Use `charter …
secret exec`/`cp`"*, pointing the agent at that door. Both leak denials, the `Read`/`Grep`
denial, and `secret get`'s masked hint and non-interactive refusal now name `secret exec` as
the covered route and say plainly what `cp` is for and that reading a materialised copy back
is the same leak.

**The ledger #423 proposes is deliberately not built**, and the reason is in the
`_leak_reason` docstring: it matches a path spelling, so `/tmp/./x`, a hardlink, a copy or
`python3 -c open(...)` walks past it — the same shape as the guard it would be patching — at
the price of a lookup on a path that runs on every Bash tool call.

The **first 70 characters of every denial are byte-identical** to before, because `_trace`
records `reason[:70]` as a tally key. A test now pins that too.

## Tests

`tests/test_documented_limits.py`, 21 tests. Every assertion is an `allow`, which is the
unusual direction for a security suite and is the point: each says "the docs promise nothing
here", and a failure means go make the named paragraph true again. Each class carries a
**positive control**, because an all-`allow` file passes just as well against a guard that
was deleted.

Mutation tested, `__pycache__` cleared between runs, green restored after each:

| Mutation | Red |
|---|---|
| add `python3 base64 jq cut dd git` to `_READERS` | 6 failures |
| make `redact` scrub base64 + reversed forms | 2 |
| widen `_VAULT_PATH_RE` to the cp destination and an outside vault | 2 |
| reword the denial's first 70 chars (`Use` → `Prefer`) | 1 |
| drop "NOT redacted" from `--exec` help | 1 |
| re-parse `sh -c '<string>'` in `_leak_reason` | 1 |

Full suite: `Ran 4549 tests … OK`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,18 +188,21 @@ sequenceDiagram
     K-->>C: output, may echo the value
     C-->>A: output, value redacted
     A-->>M: "rollout 3/3 ready"
-    Note over M,A: no step here ever put the value in a context window
+    Note over M,A: charter put the value in no context window — step 2 chose the command
 ```
 
 Reads are masked by default, `--reveal` refuses a non-interactive stdout, and the plugin's
-guard denies both that flag and `cat`-ing a vault file outright.
+guard denies that flag and known reader programs pointed at a vault file. Those close the
+accidental paths; they are not a boundary against a command chosen on purpose — see
+[SECURITY.md](SECURITY.md).
 
 **Vaults are pluggable, and the provider is where the storage guarantee comes from.** Three
 ship today — `plain_file`, `reference` (point at a value that lives elsewhere) and
 **`1password`** — and more are coming. Read [docs/secrets.md](docs/secrets.md) before
 storing anything real: **`plain_file` is plaintext at mode 0600, with no encryption at
-rest.** What every provider buys you is the same and it is the point — *the model never
-sees the value*. What only a real backend buys you is encryption. The vault is not a
+rest.** What every provider buys you is the same and it is the point — ***charter never
+prints the value into the conversation***; what the command you hand it to does with it is
+that command's business. What only a real backend buys you is encryption. The vault is not a
 password manager; 1Password is, and charter will read from it.
 
 **A browser login, with the password never typed into the conversation.** `charter browser
@@ -272,8 +275,9 @@ you use. The browser lane additionally shells out to `npx`. That is the whole li
   travels — disk only, committed, or pushed to the team — is one setting, `[memory].share`,
   and it **defaults to `local`**.
 - **Vault** — where a persona's credentials live. The provider decides the storage
-  guarantee; the boundary is the same for all of them, and it is that the value never
-  enters an agent's context or transcript.
+  guarantee; the boundary is the same for all of them, and it is that **charter never puts
+  the value in an agent's context or transcript** — the command charter hands it to still
+  can.
 
 ## Also in the box
 
--- a/README.md
+++ b/README.md
@@ -192,9 +192,10 @@ sequenceDiagram
 ```
 
 Reads are masked by default, `--reveal` refuses a non-interactive stdout, and the plugin's
-guard denies that flag and known reader programs pointed at a vault file. Those close the
-accidental paths; they are not a boundary against a command chosen on purpose — see
-[SECURITY.md](SECURITY.md).
+guard denies that flag and known reader programs whose argument spells out a vault path.
+Those close the accidental paths; they are not a boundary against a command chosen on
+purpose — a glob, a shell variable or an unlisted program walks past, by design and not by
+oversight — see [SECURITY.md](SECURITY.md).
 
 **Vaults are pluggable, and the provider is where the storage guarantee comes from.** Three
 ship today — `plain_file`, `reference` (point at a value that lives elsewhere) and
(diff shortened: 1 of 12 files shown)


diazoxide:
## Follow-up: writing the limit down found a real hole (e3a96cb)

Asked of my own new sentence in `SECURITY.md` — *"denies a known file-reading program
pointed at a path under `.charter/`"* — **what is the next spelling of this that still gets
through?** The answer was not exotic:

```
cat .charter/vaults/db.json     → deny
cat .charter//vaults/db.json    → ALLOW      ← same file, same program, one keystroke
cat .charter/./vaults/db.json   → ALLOW
```

Not the `_READERS` name ceiling and not a path the caller chose — the guard missing the one
path it claims to cover. Both guards (Bash leak + `Read`/`Grep`) shared the pattern and the
hole.

`_names_a_vault_path` now tests the operand **and** its `os.path.normpath`, and restores the
trailing slash on the normalised form when the operand carried one — without that,
`grep -r . .charter//vaults//` (the operand that walks *every* vault file) normalises into a
form the pattern rejects. Still not a resolver: `realpath` would stat every operand of every
Bash call, so a symlink at a path you chose stays the documented limit, and the docs now say
*canonicalised, not resolved*.

`tests/test_vault_path_spellings.py` **generates** equivalent spellings rather than listing
bad strings, and asserts each lands where the canonical form lands — denied for a vault file,
**allowed** for `.charter/vaults.json` (the registry: config and paths, never values), which
is the false positive a wider pattern would bring back.

Mutation tested: dropping the normalised arm → 12 red; dropping the trailing-slash
restoration → 1 red; dropping the **raw** arm → survived on the first pass, so a test was
added for the case it covers (`cat .charter/vaults/../../elsewhere`, a denial that existed
before this change and may not be silently traded away) → now 1 red. Green restored after
each, `__pycache__` cleared between runs.

Full suite: `Ran 4557 tests … OK`.

PR #464, opened 2026-08-24
Pin what CI runs, pin the run site, and stop refusing the half that is safe (#443, #455, #456)

### LINKED ISSUE #443: Release workflow's publish job runs unpinned third-party actions while holding `id-token: write
`.github/workflows/release.yml:98-105`:

```yaml
permissions:
  id-token: write            # :99  — the sole holder
...
- uses: actions/download-artifact@v4                # :101
- uses: pypa/gh-action-pypi-publish@release/v1      # :105
```

Neither is pinned to a commit SHA. `release/v1` is a **branch head**, not a tag:

```
$ git ls-remote --heads https://github.com/pypa/gh-action-pypi-publish.git
dc37677b2e1c63e2034f94d8a5b11f265b73ba33  refs/heads/release/v1
```

and there is no `refs/tags/release/v1`. All 13 `uses:` in the repo are floating refs; zero SHA pins. `actions/checkout@v4` and `download-artifact@v4` are floating major tags — GitHub's immutable releases cover exact version tags, not those.

A force-push to that branch executes attacker code inside a job where `ACTIONS_ID_TOKEN_REQUEST_URL`/`_TOKEN` are ambient. The code mints an OIDC token and PyPI's Trusted Publishing verifies it as charter's legitimate publisher, because the claim (repo + workflow + environment) is genuine. This is the tj-actions/changed-files mechanism (CVE-2025-30066); SHA pinning is what separated victims from non-victims.

Same exposure, smaller radius, in `announce`, which holds `contents: write` (`:111-114`) while running `actions/checkout@v4` and `actions/setup-python@v5`.

**Blast radius, corrected in both directions:** `hooks._autosync_version_lock` (`hooks.py:1770-1796`) installs the plane's **pinned exact** version and refuses downgrades, so existing planes are not swept by a malicious `latest` — but once a pin is bumped it runs unattended at every teammate's session start (`docs/control-plane.md:367-369`). Fresh `uv tool install charter-cp` and `charter update` (`commands_update.py:37`) do take latest. Then it executes on every session via `hooks/hooks.json` SessionStart and on every Bash/Read/Write/Task tool call.

**Note:** `release.yml:2-5` is **not** false — it claims only that no API token exists to leak, which is true. Optionally narrow it to "no *PyPI* credential."

**Mitigating context:** `pypa/gh-action-pypi-publish@release/v1` is the verbatim invocation PyPA's and PyPI's own Trusted Publishing docs prescribe. This is an ecosystem default, not a charter-specific misconfiguration. That bears on blame, not on risk.

**Fix:** pin every `uses:` to a full commit SHA with the tag in a trailing comment — `pypa/gh-action-pypi-publish@76f52bc884231f62b9a034ebfe128415bbaabdfc # v1.12.4` — starting with the two steps standing next to `id-token: write`. Add a Dependabot/Renovate entry so the SHA moves deliberately.

**Test:** `tests/test_workflows.py::test_every_action_is_pinned_to_a_sha` — parse both workflow YAMLs, assert every `uses:` matches `@[0-9a-f]{40}`.

---

### LINKED ISSUE #455: dev-install: the no-interpolation guard is pinned at the build site but not the run site
Recorded so a partially-closed guard is not carried forward as closed. Found by the independent reviewer of PR #454, and correctly reported rather than left implicit.

## What is pinned

`dev_install_argv`'s docstring claims *"there is no format call on this path."* A canary test now pins that at the **build** site — a brace-bearing value proves a `.format()` there would be observable.

## What is not

The reviewer put `.format()` back in `_sync_dev` — the site that actually **runs** the argv — and it **survives all 4415 tests**. So the claim is pinned where the argv is constructed, not where it is executed.

## Why this is not urgent

The exposure is unreachable without a second, independently-tested change. Interpolation only matters if `DEV_SPEC` contains a `{…}`, and adding one immediately reddens two existing tests:

- `test_the_spec_carries_no_interpolation_point_of_any_kind`
- `test_the_spec_is_exactly_what_the_docs_tell_you_to_type`

So the composite invariant — *a committed value cannot reach an argv through interpolation on the dev-install path* — cannot be broken without a red test somewhere. That is why it did not block the merge.

## Why it is still worth closing

The protection is currently *emergent* from two separate tests rather than stated at the site that would be exploited. Someone refactoring `DEV_SPEC` into something templated could satisfy both existing tests with a new constant and reintroduce the exposure at `_sync_dev` with nothing red.

One line in the canary test, extended to cover the run site as well as the build site, closes it.

## Context

This matters because charter has now been bitten three times by a committed value crossing into a structured context: `[frame] hotkey` reaching tmux config text where a newline achieved code execution at launch; a committed `mcp.json` key injecting YAML into a generated agent (#453, still open, CRITICAL); and the general class that drove the whole `channel`-as-closed-enum design in #454. The discipline is worth holding at every site, not most of them.

### LINKED ISSUE #456: charter update refuses on a checkout, so the dev channel's plugin refresh is unreachable where it is most wanted
Found immediately on first real use of the dev channel (#454), switching the maintainer's own plane over.

## What happens

```
$ charter doctor
  !  plugin files   installed 18cdb73, marketplace 38c6671 — skills/browser/SKILL.md, skills/secrets/SKILL.md
        → this plane tracks the dev channel and its plugin does not. … Run: charter update

$ charter update
✗ this is a charter checkout — refusing to install over the tree you are editing.
•   what you probably want:  charter version
```

`doctor` names `charter update` as the fix. `charter update` refuses. There is no flag to run only the plugin half, so the recommended remedy is unreachable and the operator is left to work out `marketplace update` → `uninstall` → `install` by hand.

## Why the refusal is right, and still wrong here

Refusing to install a wheel over the checkout someone is editing is correct and must stay — that guard exists for good reason.

But `cmd_update` does two independent things: it moves the **CLI**, and it force-refreshes the **plugin**. Only the first is unsafe on a checkout. The plugin is a separate artifact in `~/.claude/plugins/`, entirely outside the tree, and refreshing it over a checkout is harmless.

## Why it bites exactly the wrong person

A charter checkout is where a maintainer works, and a maintainer is the most likely person to want the dev channel at all. So the guard blocks the remedy in precisely the case the feature was built for.

## Suggested shape

Either:

1. On a checkout, skip the CLI install, still refresh the plugin, and say so — "the CLI is your checkout; refreshed the plugin only." The operator's CLI is already whatever they have checked out, so there is nothing to move.
2. Or add `charter update --plugin-only`, and have `doctor`'s hint name that command when it detects a checkout.

(1) is better: it needs no new flag, and it makes `doctor`'s existing hint true rather than adding a second thing to know.

Either way, `doctor`'s hint must name a command that actually works from where the operator is standing. A remedy that refuses when followed is worse than no remedy, because it costs the reader their trust in the next hint too.

## Verified working by hand

The three steps `force_refresh` performs do work on a checkout — I ran them:

```
claude plugin marketplace update charter
claude plugin uninstall charter@charter --scope project
claude plugin install charter@charter --scope project -y
```

Result: `✓ plugin files  matches the marketplace clone (38c6671)`. So nothing about the plugin refresh is unsafe here; only the CLI half needed to be skipped.

diazoxide:
Closes #443
Closes #455
Closes #456

Three issues from the 2026-08-24 audit and from the first real use of the dev channel. Full
suite green: **4548 tests, OK**, stdlib `unittest`, no network.

## #443 — every action CI runs is pinned to a commit

`release.yml`'s `publish` job holds `id-token: write`. Trusted Publishing means there is no
PyPI token to steal — and that moves the question rather than answering it: the OIDC token
is minted on demand and PyPI verifies a claim that is entirely genuine, so **whatever code
runs in that job can publish charter**. It named that code with floating refs:

- `pypa/gh-action-pypi-publish@release/v1` — `refs/heads/release/v1`, a **branch head**, no
  such tag exists;
- `actions/download-artifact@v4` — a tag its owner can retarget.

All 13 `uses:` in both workflows now name a full commit SHA with the tag it resolved to in
a trailing comment, e.g. `pypa/gh-action-pypi-publish@dc37677… # v1.14.2`. Resolved from
each floating ref's current target, so **no version changed** — only the ref did.

`.github/dependabot.yml` (new) opens the pull request that moves a pin. A pin freezes a
security fix out as effectively as it freezes an attacker out; Dependabot opens it, a human
still merges it.

**Permissions audit.** `release.yml` already carried workflow-level `contents: read` with
`announce` narrowing to `contents: write` on that job alone. What was missing was the
statement that `publish`'s job-level block **replaces** the workflow's rather than adding
to it — so it holds `id-token: write` and *nothing else*, including no `contents`, which is
correct because it never checks out and `download-artifact` reads same-run artifacts with
the runtime token. That is now a comment where someone would otherwise "fix" it.

### The test asks the property, not the spelling

`tests/test_workflows.py`: *given only this ref, can the bytes it resolves to change
without a commit landing in this repository?* A full commit SHA, an image digest, or an
in-repo path — everything else is a third party's promise, however spelled.

- It **walks** every YAML under `.github/` plus every `action.yml`/`action.yaml` in the
  tree, not the two workflow files that exist today. A local composite action is
  `uses: ./tools/thing`, correctly immutable, and whatever *that* file runs executes in the
  calling job with the calling job's token.
- It **fails closed**: `uses:` tokens are counted at the byte level and the test refuses to
  pass unless a ref was parsed for every one, so a flow mapping or a folded scalar fails the
  suite instead of being silently skipped.
- It also requires each SHA to carry a `# <tag>` comment, and requires something to be
  watching the pins.

**Named, not claimed:** it cannot check that a SHA is a real commit of the repo beside it or
that the tag comment is honest — both need the network. A wrong-but-well-formed SHA fails
in CI loudly on the next run. It says nothing about a `run:` step piping a script off the
internet; there is none.

`release.yml`'s header used to end *"a leaked secret cannot be used to publish charter"* —
true, and answering the smaller question. Rewritten to answer the real one. The dev-channel
news entry and `test.yml`'s comment, both of which cite #443 as an open exposure and both
of which ship in this same unreleased set, are updated to say what is now true.

## #455 — the run site of the no-interpolation rule

`dev_install_argv`'s docstring claims *"there is no format call on this path"*, pinned by a
canary at the **build** site. #454's reviewer put `.format()` back into `_sync_dev` — the
site that actually runs the argv — and it survived all 4415 tests.

Now stated as a property of the boundary: **the argv charter hands the operating system on
this path is, element for element, the argv `dev_install_argv` returned.** One equality
forbids interpolation, appending, re-splitting and shell-joining together, including the
spellings nobody has thought of — which is the difference between this and a list of
forbidden characters.

The recorder sits at `subprocess.run`, with `subprocess.Popen` and the `os.exec*`/`spawn`
family trapped to raise. A counter on `util.run` would pin `_sync_dev`'s *current spelling*,
not the property — the same mistake as the `Path.stat` counter that missed an `open()`.
Residual named in the docstring: a `ctypes`-level spawn is not visible to any test here.

## #456 — the remedy that refused when followed

```
$ charter doctor
  !  plugin files  … Run: charter update
$ charter update
✗ this is a charter checkout — refusing to install over the tree you are editing.
```

`cmd_update` does two independent things and only one is unsafe on a checkout. The plugin
lives in `~/.claude/plugins/`, outside the tree, refreshed from a clone that is not the tree
either. On a **dev-channel checkout** `update` now skips the CLI install, says so, and
refreshes the plugin:

```
!  this is a charter checkout — the CLI here is the tree you are editing, so nothing was
   installed over it.
   the charter you run is this checkout, moved by git:  charter version
   the plugin lives outside this tree, so that half still runs:
✓  plugin: … — it loads on the NEXT session.
```

Option (1) from the issue: no new flag, and it makes doctor's existing hint true instead of
adding a second thing to know.

**The CLI refusal is not weakened.** It is asserted on both channels, with and without
`--to`, by name (`_sync_dev`/`_sync_to` never entered) *and* at the argv layer (no command
ran at all). Still refused outright, with the original message and exit code: a stable
checkout, and `--to X.Y.Z` anywhere in a checkout — that names a published CLI, which is
exactly the half that cannot happen here. Also not done: `_move_harness`, because it writes
into the plane root, which on a charter checkout is the tree being protected. And with no
`claude` on PATH it says there is nothing to do rather than reporting a refresh that did not
happen.

One test spans `doctor`'s hint text and `cmd_update`'s behaviour, because apart they were
both green while the pair was broken. `skills/update/SKILL.md`'s sentence "`update`
refuses" is rewritten in the same commit — a doc promising what the code no longer does is
itself the bug.

## Mutation testing

13 mutations, each applied → confirmed RED → restored → confirmed GREEN, `__pycache__`
cleared between runs:

| # | Mutation | Killed by |
|---|---|---|
| M1 | one action back to `@v4` | `test_no_workflow_runs_a_ref_its_owner_can_move` |
| M2 | a NEW workflow file with `@main` | same (proves the walk, not a hardcoded pair) |
| M3 | `- {uses: …}` flow mapping | `test_every_uses_token_produced_a_ref` (fail-closed) |
| M4 | `dependabot.yml` deleted | `test_something_is_watching_the_pins_for_updates` |
| M5 | a SHA with no `# tag` comment | `test_every_pin_says_which_version_it_is` |
| M6 | composite action under `.github/actions/` | pinning test |
| M14 | local composite action in `tools/` | pinning test (the hop outside `.github/`) |
| M7 | `.format()` at the run site (**the exact #455 mutation**) | run-site canary |
| M8 | an element appended to the argv | run-site canary ×3 |
| M9 | argv joined into one string | run-site canary ×3 |
| M10 | blanket checkout refusal restored | dev-checkout test **and** the doctor-hint test |
| M11 | `_move_harness()` on the checkout path | `test_a_dev_checkout_still_gets_the_plugin_half` |
| M12 | `--to` also taking the plugin-only path | `test_an_explicit_target_…_refused_outright` |
| M13 | CLI install reinstated on the checkout path | `test_a_charter_checkout_cannot_install_over_itself` (both channels) |

## What is the next spelling that still gets through?

- **#443:** a ref that *is* 40 hex characters but is a branch named after one — theoretical,
  GitHub resolves it as a commit, and it is named in the module docstring rather than
  claimed away. A `run:` step that curls a script is not covered by pinning at all.
- **#455:** an exec that never reaches `subprocess` or `os` — a `ctypes` spawn. Named, not
  covered.
- **#456:** none found; the parse of doctor's hint (`split("Run:")`) fails loudly if the
  hint's wording changes, which is the coupling this fix wants.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,20 @@
+version: 2
+# What makes a SHA pin maintainable instead of merely frozen.
+#
+# Every `uses:` in `.github/workflows/` is pinned to a full commit SHA (#443), which is the
+# only ref a third party cannot move under charter's feet. The cost of that is that a
+# security fix upstream does not arrive on its own — so this opens the pull request that
+# moves each pin, and a human reads the diff and merges it. Deliberate, reviewed, and not
+# automatic: that is the same bargain `charter update` makes with the operator.
+#
+# `github-actions` only. charter's runtime has zero dependencies (`pyproject.toml` says
+# `dependencies = []`, and `tests/test_packaging.py` holds it there), so there is no `pip`
+# ecosystem here to watch.
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 5
+    commit-message:
+      prefix: "ci"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,7 +14,9 @@ name: release
 #
 # Bumping a pin: read the diff for the range, then put the new SHA and its tag
 # here. Dependabot opens that pull request (`.github/dependabot.yml`); it does
-# not merge it. `tests/test_workflows.py` fails on any ref that is not 40 hex.
+# not merge it. `tests/test_workflows.py` reads these files as YAML — not as lines of
+# text — and fails on any ref that is not a full commit SHA, an image digest, or a file
+# tracked in this repository.
 #
 # Trigger: push a version tag.
 #   git tag v0.1.0 && git push origin v0.1.0
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,8 +15,10 @@ name: release
 # Bumping a pin: read the diff for the range, then put the new SHA and its tag
 # here. Dependabot opens that pull request (`.github/dependabot.yml`); it does
 # not merge it. `tests/test_workflows.py` reads these files as YAML — not as lines of
-# text — and fails on any ref that is not a full commit SHA, an image digest, or a file
-# tracked in this repository.
+# text — and fails on any ref written here that is not a full commit SHA, an image digest,
+# or a file tracked in this repository. `container:`/`services:` count as refs, in both
+# shapes GitHub writes an image in: the `image:` mapping and the bare scalar shorthand.
+# What it does NOT reach is what a pinned action's own tree then names — see #473.
 #
 # Trigger: push a version tag.
 #   git tag v0.1.0 && git push origin v0.1.0
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,8 +1,20 @@
 name: release
-# Publishes to PyPI via Trusted Publishing (OIDC) — there is no API token
+# Publishes to PyPI via Trusted Publishing (OIDC) — there is no PyPI API token
 # anywhere: not in the repo, not in Actions secrets, not on a laptop. PyPI
 # verifies GitHub's signed identity claim for this exact repo + workflow +
-# environment, so a leaked secret cannot be used to publish charter.
+# environment, so there is no publishing credential to leak.
+#
+# That moves the question rather than answering it, and this file has to answer
+# the moved one: what CAN publish charter is code running inside the `publish`
+# job, because the OIDC token is minted there on demand and the claim it carries
+# is genuine. Every `uses:` below is therefore pinned to a full commit SHA, with
+# the tag it resolved to in a trailing comment. A floating ref — `@v4`, or
+# `@release/v1`, which is a BRANCH head and not a tag — is a promise by whoever
+# can move it, and moving it is the whole of CVE-2025-30066 (#443).
+#
+# Bumping a pin: read the diff for the range, then put the new SHA and its tag
+# here. Dependabot opens that pull request (`.github/dependabot.yml`); it does
+# not merge it. `tests/test_workflows.py` fails on any ref that is not 40 hex.
 #
 # Trigger: push a version tag.
 #   git tag v0.1.0 && git push origin v0.1.0
@@ -23,10 +35,10 @@ jobs:
     name: Verify the tag matches the packaged version, and that it has notes
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
       # Pinned rather than inherited: both steps below run charter itself, which needs
       # 3.11+ for tomllib. The runner's default `python` is a version nobody here chose.
-      - uses: actions/setup-python@v5
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
         with:
           python-version: "3.12"
       - name: Compare the git tag to pyproject.toml
@@ -63,8 +75,8 @@ jobs:
       matrix:
         python-version: ["3.11", "3.12", "3.13", "3.14"]
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
         with:
           python-version: ${{ matrix.python-version }}
       - name: Run the suite
@@ -75,15 +87,15 @@ jobs:
     needs: test
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
         with:
           python-version: "3.12"
       - name: Build
         run: pipx run build
       - name: Validate metadata renders on PyPI
         run: pipx run twine check dist/*
-      - uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
           name: dist
           path: dist/
@@ -96,13 +108,18 @@ jobs:
       name: pypi
       url: https://pypi.org/p/charter-cp
     permissions:
+      # A job-level `permissions:` REPLACES the workflow's, it does not add to it — so
+      # this line is the whole grant: `contents` is `none` here, not the `read` above.
+      # That is deliberate and must stay. This job never checks the repository out; it
+      # downloads the artifact `build` produced, which `actions/download-artifact` reads
+      # from the same run with the runtime token, needing no `GITHUB_TOKEN` scope at all.
       id-token: write        # REQUIRED: mints the OIDC token PyPI verifies
     steps:
-      - uses: actions/download-artifact@v4
+      - uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
         with:
           name: dist
           path: dist/
-      - uses: pypa/gh-action-pypi-publish@release/v1
+      - uses: pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33 # v1.14.2
 
   announce:
     name: Create the GitHub Release from the shipped news entry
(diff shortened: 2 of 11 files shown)


diazoxide:
Round 3: the `container:` shorthand hole is closed, and the claim that overreached is narrowed in the same commit (540fd57).

**Finding 1 — (A) the reported bypass reproduced, and is fixed.** `_CODE_KEYS = ("uses", "image")` knew `jobs.<id>.container.image` and not `jobs.<id>.container`, the string shorthand. `container: evil/img:latest` on the `publish` job left all 30 workflow tests and the 4616-test suite green. `services:` had the same two shapes and the same hole.

Fixed by reading both shapes, and by giving `Ref` a `kind` (`ACTION`/`IMAGE`) so the rule tests select on what a reference *is* rather than on the key name that spelled it. Ten spellings injected into release.yml's `publish` job — plain, single- and double-quoted, `docker://`, a hex-escaped key, the mapping form, the wrapped-scalar form, and the three `services:` variants — are all RED now, by `test_no_job_runs_a_container_image_its_publisher_can_move` or by `Unparsed` naming the construct. Twelve mutants applied, all RED, restored GREEN, `__pycache__` cleared.

**Finding 2 — (C) claim narrowed.** The four sentences that said the rule covers container images are now true rather than trimmed, because the code went the rest of the way. What was *added* is the limit nobody had stated: what is enforced is *every reference charter writes*, which is narrower than *every byte that runs in the publishing job*.

**(D) filed as #473.** `uses: owner/action@<sha>` pins that action's tree; it does not pin what that tree then names — a Docker action's `Dockerfile` `FROM` tag, or a composite action's own `uses:`. A **local** `uses: ./x` is followed into its `action.yml`; a remote one is not followed at all, because reading it needs the network and no test here makes a network call. #473 has the reproduction (identical action content, refused locally, invisible remotely) and the three options. Referenced from the news entry as the known remaining case; deliberately not widened into this PR.

**Not weaker than main:** `origin/main` has no `tests/test_workflows.py` and ships `@v4` / `@release/v1`. Every input main allows here, this branch denies.

Full suite: `Ran 4618 tests … OK`.

Issue #331, opened 2026-08-21
A committed `vaults.json` can point a vault at any file on the machine, and `doctor` chmods it at session start


diazoxide:
`vaults.json` at the plane root is the **committed** half of the vault registry (`charter/config.py:250`, and the module docstring of `charter/secrets/registry.py` explains why it is committed: the wiring has to travel). `load_registry` (`:70-90`) layers the local half over it per field, so a vault present only in the shared half is taken entirely from the committed file.

A plain-file vault's `file` is resolved by `VaultProvider.file_path` (`charter/secrets/base.py:100-118`):

```python
# charter/secrets/base.py:115-118
p = self.config.get("file")
p = Path(p).expanduser()
return p if p.is_absolute() else (_config.ROOT / p)
```

Absolute is accepted deliberately — the docstring says so, for "a vault deliberately kept outside the plane". Combined with the committed half, that means a file in git decides which path on the machine charter treats as a vault. There is no containment check and no confirmation.

**The part that surprised me is that it fires unattended.** `check_vaults` (`charter/doctor.py:1104-1130`) calls `health()` on every registered vault, and `charter doctor` runs from the SessionStart hook (`hooks/hooks.json`, 20s budget). `PlainFileProvider.health()` reaches `_load()`, which calls `_tighten()` (`charter/secrets/plain_file.py:44-52`, `:84-88`) — a chmod to 0600 on any file with group or other bits set.

**Verified on 0.47.2**, with a committed-shaped `vaults.json` naming a path outside the plane, in a throwaway plane:

```
BEFORE: -rw-r--r--
$ charter doctor
  ✓  vaults           1 reachable (references not resolved)
AFTER : -rw-------
```

`doctor` reported green while changing the mode of a file outside the control plane, with nobody watching. `charter/statusline.py:1004` calls the same `health()` behind a TTL cache, so the status line reaches it too.

**What bounds it today.** `_tighten` only ever tightens, never loosens — so this is an integrity and availability problem, not a privilege escalation: a file that some other process needs group-readable silently stops being so. The larger reach is not the chmod but the identity: whatever that path is, `get()` will read values out of it and `set()`/`_save()` will overwrite it whole at 0600 (`plain_file.py:54-62`). Both need an operator to run a `secret` command against a vault name they did not register.

[...]

PR #465, opened 2026-08-24
secrets: a fingerprint you cannot check a guess against, and a vault that is 0600 before the plaintext (#436, #437)

### LINKED ISSUE #436: Masked `secret get` prints an unsalted sha256 prefix plus exact byte length
`charter/commands_secrets.py:435-436`:

```python
digest = hashlib.sha256(value.encode()).hexdigest()[:12]
print(f"{args.vault}/{args.key}: present · {len(value)} bytes · sha256:{digest}")
```

Unsalted, un-keyed, 48 bits, plus the exact byte count. `:381` also prints exact length on write.

Verified with a fabricated weak value — stored `Summer2024!`, then to a non-tty stdout with no guard firing:

```
audit2/WEAK: present · 11 bytes · sha256:323725e8eff4
```

Offline confirmation with no further access to charter:

```
Password1    len= 9  19513fdc9da4
Summer2023!  len=11  935cfa60fb49
Summer2024!  len=11  323725e8eff4   ← match
```

The length prefilters, the digest confirms. Decisive against a human-chosen password; irrelevant against a 40-char random token.

`docs/secrets.md:44-45` documents exactly what is printed — but as a *reassurance* (*"never the value"*), never as an offline verification oracle, and `skills/secrets/SKILL.md:57-60` instructs the model to run it. No guard denies `secret get`.

**Fix:** `hmac.new(plane_key, value.encode(), 'sha256').hexdigest()[:12]`, with `plane_key` 32 random bytes generated at `charter init` and stored 0600 under `.charter/`. The fingerprint stays stable and comparable within a plane — which is all it is for ("is this the same value as before") — and becomes useless as a wordlist oracle. Optionally bucket the length. Say so at `docs/secrets.md:44`.

**Test:** `tests/test_secret_get_masked.py::test_the_fingerprint_is_not_a_raw_sha256_of_the_value` — assert the printed digest ≠ `sha256(value).hexdigest()[:12]`, and that two planes print different digests for the same value.

---

### LINKED ISSUE #437: PlainFileProvider._save` writes plaintext into a pre-existing file at that file's existing mode
`charter/secrets/plain_file.py:56-64` opens the vault in place with `O_CREAT|O_TRUNC`. For an **existing** inode the mode argument is ignored, so the plaintext is written at whatever mode the file already had; the `chmod` lands only after `json.dump` returns.

Two docstrings assert the opposite, and that false premise is the stated reason `_tighten` is skipped on the write path:
- `:59` — *"Create with 0600 from the start so the plaintext is never briefly world-readable"*
- `:83-84` — *"`set`/`delete` need no call — `_save` recreates the file at 0600 with `O_CREAT` and chmods it again"*

Measured with an instrumented `json.dump`, fabricated secret, temp dir:

```
pre-existing mode:                0o644
mode while plaintext on disk:     0o644
mode after set:                   0o600
inode changed?                    False
```

Impact is small — milliseconds, requires an already-loose file — but it is a confident docstring that is wrong about *why* something is safe.

Same in-place pattern at `registry.py:174-179` and `config.py:184-185`; those carry paths and account pins rather than values, so they are cosmetic by comparison.

**Fix:** write to a sibling temp created 0600 and `os.replace` into position (also makes the update atomic), or call `_tighten(p)` before opening. Correct both docstrings — `O_CREAT` does not recreate an existing inode. Optionally `mkdir(mode=0o700)` for `VAULTS_DIR`, since `.charter/vaults/` is currently `drwxr-xr-x` and vault names are enumerable by other local users.

**Test:** `tests/test_plain_file.py::test_a_preexisting_loose_vault_is_tightened_before_the_write` — chmod 0644, patch `json.dump` to stat mid-write, assert 0600.

---

diazoxide:
Closes #436
Closes #437

Both reproduced first, exactly as filed.

## #436 — the masked line was an offline check against a guess

```
audit2/WEAK: present · 11 bytes · sha256:323725e8eff4
```

The length prefilters a wordlist and the 48-bit digest confirms the hit, with no
further access to charter. It is a hygiene defect rather than an escalation —
anyone who can run `secret exec` already holds a better oracle, and the
assessment's own second pass says so — but this line travels where the value
never does: an agent's transcript, a pasted terminal, a screenshot in a ticket.

**Now:**

```
audit2/WEAK: present · 1–15 bytes · fp:9c41a0b7e5d2
```

- `HMAC-SHA256(plane key, value)[:12]`, the key 32 random bytes generated on
  first use and kept 0600 under `.charter/`. Stable and comparable inside one
  plane — the only comparison anyone makes with it — and not computable by
  anyone without the key.
- Power-of-two size band instead of an exact count, counted in UTF-8 bytes
  (`len()` on a `str` counted characters while the label said bytes).
  `secret set` bands too: with `--from-file` that line lands in the transcript of
  an agent that has never seen the value.
- No key possible → **no fingerprint**, never an unkeyed one.
- `.charter/fingerprint.key` joins `.charter/vaults/` behind the read guard.
  It was not there, and the guard's own denial names the path it refused; for a
  1Password vault — nothing on disk to deny — the key was the only readable
  thing between the printed line and the value.

### The property, not the spelling

"assert the digest ≠ `sha256(value)[:12]`" is a spelling. `blake2b(b"charter" +
value)[12:24]` walks past it and is exactly as checkable offline. The test that
carries the weight runs one value through two planes whose only difference is
the key file and requires the fingerprints to differ — which no pure function of
the value can do, including constructions nobody has thought of. Stability
within a plane is asserted alongside it so "print random bytes" does not pass
either. The filed regression is kept as a backstop, widened to every hash
`hashlib` guarantees over three constant-salt spellings.

Length is tested as indistinguishability (all lengths in a band print one
label), paired with a boundary assertion so a constant label fails too.

## #437 — the vault was at its old mode while the plaintext went in

`O_CREAT|O_TRUNC, 0o600` ignores the mode argument for an inode that already
exists. Measured on a 0644 vault: 0644 while the plaintext was on disk, 0600
after, same inode. Two docstrings asserted the opposite, and that false premise
was the stated reason `_tighten` is skipped on the write path.

The mode is now settled on the **descriptor**: open without `O_TRUNC`, `fchmod`,
`fstat` the same fd and **read the mode back**, then truncate and write. Reading
it back is the point — a chmod that returns 0 is not evidence the bits moved on
a mount with fixed permissions (exFAT, many network shares). If the mode is
still loose charter writes nothing and says so, and the previous contents
survive; warning and writing anyway leaves the credential world-readable with
the warning scrolled off the top of a log.

Both docstrings corrected in the same commit. A vault directory charter creates
is 0700. `registry._write` — named in the issue, carries paths and env variable
NAMES rather than values — takes the same descriptor-first path, without the
refusal, because its shared half is deliberately 0644 and committed.

**Not touched:** the plain-file vault stores plaintext at rest. Documented,
accepted, and not the defect.

### Instrumented at a layer every spelling reaches

A spy on `json.dump` (what the issue suggested) is blind to `Path.write_text`, a
bare `os.write`, and the deletion of the guard it pins. The test wraps `io.open`
— which `os.fdopen`, `open()` and `Path.open` all reach — and `os.write`,
records the mode from an `fstat` of the descriptor being written, and keys
records by `(st_dev, st_ino)` so a temp-file-plus-`os.replace` implementation is
measured on the right file. It then asserts the watch **actually saw** the
plaintext, so an implementation writing through a path it does not observe goes
RED asking to be extended rather than passing on having seen nothing.

## Mutation testing

12 mutants, `__pycache__` cleared between every run, restore verified green each
time.

| # | Mutation | Expected | Got |
|---|---|---|---|
| M1 | `_save` reverted to the filed defect | RED | RED (5 failures) |
| M2 | chmod the PATH before open, no read-back | RED | RED (2) |
| M3 | same guard, plaintext via `Path.write_text` | **GREEN** | GREEN |
| M4 | vault directory left at the umask default | RED | RED (1) |
| M5 | fingerprint back to `sha256(value)[:12]` | RED | RED (7) |
| M6 | `blake2b(salt+value)[12:24]` — the next spelling | RED | RED (6) |
| M7 | unkeyed digest as the no-key fallback | RED | RED (2) |
| M8 | size band back to the exact count | RED | RED (2) |
| M9 | size band collapsed to one constant | RED | RED (2) |
| M10 | key file written without reading its mode back | RED | RED (1) |
| M11 | `registry._write` reverted | RED | RED (1) |
| M12 | `fingerprint.key` dropped from the read guard | RED | RED (3) |

M3 is the one that had to stay green: it is what shows the watch is not pinned
to `json.dump` or `os.fdopen`.

## What is the next spelling of this that still gets through?

Stated in `docs/secrets.md` and in the module rather than left to be found:

- **Whoever holds the key holds the oracle.** The read guard denies it to
  file-reading tools; a shell running as you reads it, as it reads everything
  else you own. `SECURITY.md`'s framing is unchanged.
- **Within one plane the fingerprint is still an equality oracle.** Anyone who
  can `secret set` a guess can compare — and can also `--reveal --force`, so it
  is not a step up for them. It is the reason the key is per-plane rather than
  per-vault: per-vault salting would close it and break the one comparison the
  fingerprint exists for.
- **`.charter/vaults/` on planes that already exist keeps its old mode.** Only a
  directory charter creates is 0700; silently chmod-ing a directory charter did
  not create is the #331 mistake.
- **`config.py`'s state-dir migration** has the same in-place pattern, on a file
  holding paths. Left alone deliberately — it is one line in a migration path
  another change is likely touching.

## Suite

`python3 -m unittest discover -s tests` → **Ran 4557 tests … OK**. stdlib only,
`PersonaIso` for anything touching plane state, no network, no real credential
in any fixture or assertion message.

### DIFF
--- a/charter/commands_secrets.py
+++ b/charter/commands_secrets.py
@@ -15,7 +15,6 @@ which command got which one has the audit half of the story missing.
 
 from __future__ import annotations
 
-import hashlib
 import os
 import re
 import signal
@@ -26,7 +25,7 @@ import tempfile
 from pathlib import Path
 
 from . import config, util
-from .secrets import base, registry
+from .secrets import base, fingerprint, registry
 
 
 # --------------------------------------------------------------------------- #
@@ -393,7 +392,11 @@ def cmd_secret_set(args) -> int:
     except base.VaultError as e:
         util.err(str(e))
         return 1
-    util.ok(f"Set '{args.key}' in vault '{args.vault}' ({len(value)} bytes). Value not shown.")
+    # Banded, not counted, for the same reason `get` is (#436): this line is printed into
+    # whatever ran the command — an agent's transcript for `--from-file`, where the length
+    # of a value nobody in the conversation has seen is the one thing it would disclose.
+    util.ok(f"Set '{args.key}' in vault '{args.vault}' "
+            f"({fingerprint.size_band(value)}). Value not shown.")
     return 0
 
 
@@ -507,8 +510,10 @@ def cmd_secret_get(args) -> int:
         return 1
 
     if not args.reveal:
-        digest = hashlib.sha256(value.encode()).hexdigest()[:12]
-        print(f"{args.vault}/{args.key}: present · {len(value)} bytes · sha256:{digest}")
+        # NOT a hash of the value. `sha256(value)[:12]` plus the exact byte count turned
+        # this line into an offline verification oracle for a guessed value — see
+        # `secrets/fingerprint.py` (#436).
+        print(f"{args.vault}/{args.key}: present · {fingerprint.masked(value)}")
         print("(value hidden — use `charter secret exec`/`cp` to consume it, or --reveal to print)")
         return 0
 
--- a/charter/hooks.py
+++ b/charter/hooks.py
@@ -307,10 +307,22 @@ _READERS = frozenset("cat less more head tail bat nl tac xxd od strings grep rg 
 #: `.edm` is the pre-rename spelling, kept for the reason :data:`_CHARTER_PROGS` keeps the
 #: old binary name.
 #:
+#: `fingerprint.key` is here because reading it un-does `secret get`'s masking (#436).
+#: The masked line carries `HMAC(plane key, value)` rather than a hash of the value, so a
+#: guess cannot be checked against it — *unless the reader also holds the key*, at which
+#: point the offline wordlist check is back exactly as it was. That file is the whole
+#: strength of the fingerprint, so it belongs on the same side of this guard as the vaults
+#: themselves. It matters most for the providers that have no vault file at all: a
+#: 1Password-backed secret has nothing here to deny, and the key would have been the one
+#: readable thing standing between a fingerprint and the value it describes.
+#:
 #: Known limit, and the reason the tool gate does NOT reuse this as its only answer: this
 #: is a name, and a plane with `$CHARTER_HOME` set keeps its vaults somewhere this pattern
-#: cannot spell. `toolgate._resolves_into` asks the filesystem instead.
-_VAULT_PATH_RE = re.compile(r"\.(?:charter|edm)(?:/(?:vaults/|browser|active-)|/?$)")
+#: cannot spell. `toolgate._resolves_into` asks the filesystem instead — and `_control_roots`
+#: already derives the state directory from `config.STATE_DIR`, so the key file is inside
+#: the tool gate's surface on such a plane without being named there.
+_VAULT_PATH_RE = re.compile(
+    r"\.(?:charter|edm)(?:/(?:vaults/|browser|active-|fingerprint)|/?$)")
 
 
 #: `edm` is charter's pre-rename name. Kept because this is a security guard and the cost
--- a/charter/secrets/base.py
+++ b/charter/secrets/base.py
@@ -40,6 +40,147 @@ def vault_file_path(configured):
     return p if p.is_absolute() else (Path(_config.ROOT) / p)
 
 
+#: Every bit that lets an account other than the owner reach a directory. The guard is
+#: written as this mask rather than as a comparison against a list of bad modes (0755,
+#: 0750, 0775, …) on purpose: a literal set of bad values is a spelling, and the next mode
+#: nobody listed walks straight past it. ``mode & _OTHERS`` is the property.
+#:
+#: **What the mask still cannot see, said out loud rather than left to be discovered.**
+#: A POSIX mode is not the whole access-control story on either platform charter runs on.
+#: macOS extended ACLs (``chmod +a "user:bob allow read,list"``) and Linux POSIX.1e ACLs
+#: (``setfacl -m u:bob:rx``) grant another account full access while ``st_mode`` still
+#: reads 0700 — so a directory this function calls private can be readable by someone
+#: else, and `loose_dirs` will not report it. That is a real hole and it is not closed
+#: here: reading an ACL means `acl_get_file`, which is not in the standard library on
+#: either platform, and shelling out to `ls -le`/`getfacl` from a health check that
+#: `doctor` runs at every session start is a worse trade than the gap. SECURITY.md's
+#: framing applies unchanged — guard rails, not guarantees.
+_OTHERS = 0o077
+
+
+def make_private_dir(p) -> None:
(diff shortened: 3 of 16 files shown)


diazoxide:
**Round three.** Both remaining findings were claim overreach, not code. Narrowed in `0d49e5f`; no code path changed (the only `charter/` edit is a docstring).

**1. "Every directory charter creates on the way to a vault file is 0700."** The 0700 walk is `base.make_private_dir`, and only the three secrets writers call it. On the default flow they are not what creates `.charter/` — `charter vault add` writes the local registry first, through `registry._write` → `path.parent.mkdir(parents=True, exist_ok=True)` with no mode. Measured on a plane built from nothing under `umask 022`:

```
.charter               755   <-- registry._write
.charter/vaults        700
.charter/vaults/team   700
```

Under `umask 077`, `.charter` is 700. The umask decides that one level. The provider-level tests could not see it because every case let the writer create `.charter/` itself. Sentence scoped to the vault writers; `.charter/` named as the remaining case; filed as #470.

**2. "…on the vault's health line, and therefore in `charter doctor` and `charter vault list`."** `check_vaults` does `healthy, _ = prov.health()` and drops the detail; `_loose_dir_note` never sets `healthy` False. Verified with `.charter/vaults` at 0755: `vault list` names it, `doctor` prints a green `vaults 1 reachable`, `doctor --json` has no occurrence. Sentence scoped to the `STATUS` column of `charter vault list`; filed as #471.

**New tests.** `TheOrderTheCliActuallyUses` drives `registry.add_vault` first — the ordering the CLI produces and none of the existing cases reproduce — then the writer, and pins three things: the levels the writers create are 0700 under umask 000/022/077 *with the registry having gone first*; `.charter/` is decided by the umask (asserted as "two umasks, two modes", so a hardcoded 755 would not satisfy it); and the loose `.charter` appears in `cmd_vault_list`'s output and not in `doctor.check_vaults().render()`. The last two are defects pinned on purpose — if #470 or #471 lands they go RED naming the documents to update.

**Mutations** (RED → restore → GREEN, `__pycache__` cleared each time): `make_private_dir` → `mkdir(parents=True, mode=0o700)`; `registry._write` → `make_private_dir` (#470 "fixed"); `check_vaults` carrying the detail into `notes` (#471 "fixed"); `loose_dirs` dropping the stop directory; each doc overclaim restored.

Full suite: 4636 tests, OK.

PR #466, opened 2026-08-24
hooks: a denial charter cannot write is not an allow, and a file restore is not a branch move (#438, #461)

### LINKED ISSUE #438: pretooluse_read` swallows its own deny in a bare `except Exception

### LINKED ISSUE #461: the plane-root guard reads 'git checkout <file>' as a branch switch and refuses a file restore
Hit while restoring a single file in the plane root.

## What happens

```
$ git checkout charter.toml
charter guard: would switch to 'charter.toml' in the PLANE ROOT, which is one working
tree every session shares — two agents here silently clobber each other's branches …
```

`charter.toml` is a **file that exists in the working tree**, not a branch. `git checkout <path>` restores a path; `git checkout <branch>` moves HEAD. The guard treats the first as the second.

`git restore charter.toml` — the modern, unambiguous spelling of the same operation — is correctly allowed, which is what makes this a false positive rather than a policy: charter permits the operation, just not one of the two ways git spells it.

## Why it matters more than the inconvenience

The denial is confident and detailed, and it is **wrong about what the command does**. An operator following its advice would go create a workspace clone to restore one file. Worse, the message ends with "Wrong about this case? … Run it yourself, in your own terminal" — which is the right escape hatch for a guard that is *correctly* refusing an agent, and the wrong outcome for a guard that has simply misread the command.

This is adjacent to #401 (the guard stopped a branch switch but not the `git reset --hard` that destroyed eleven memory commits): the same guard is both too narrow on destructive operations and too broad on harmless ones, and both come from matching command *shape* rather than resolving what the argument actually is.

## The distinguishing rule

git itself disambiguates by whether the operand resolves as a revision, and by the `--` separator. So:

- `git checkout <path>` where the operand exists as a path in the tree and does not resolve as a revision → a file restore, allow.
- `git checkout -- <paths>` → always a file restore, allow.
- `git checkout <rev>` / `-b` / `-B` / `--detach` → a ref move, keep guarding.
- Genuinely ambiguous (a file and a branch share a name) → keep guarding; that is the case where refusing is right, and the message should say *that*, not assume a branch.

`git switch` exists precisely because this overload is confusing; the guard should not inherit the confusion.

## Test that would have caught it

A table of real git invocations paired with what they do, asserted against the guard's decision — including the ambiguous case, asserted as *still denied* so the fix cannot be over-applied. `_plane_root_reset_reason` in #401 already established this shape for `reset`.

diazoxide:
Two guards in `charter/hooks.py` that failed the same way the last four rounds failed: they matched a **spelling** instead of testing the **property**.

Closes #438
Closes #461

---

## #438 — a denial charter cannot write was an allow

`_deny` refuses by printing one JSON object on stdout. A hook that cannot print has said nothing, and a `PreToolUse` hook that says nothing is an **allow**. Both spellings of that were in the tree at once, on the same invariant:

| guard | what a `BrokenPipeError` out of `print` did | result |
|---|---|---|
| `pretooluse_read` (Read/Grep) | caught by `except Exception: return 0` around the whole body, `_deny` included | exit 0, vault read |
| `pretooluse` (Bash) | propagated to `cli.main`, which correctly turns it into 141 for `charter … \| head` | exit 141 — a *non-blocking* hook error; call proceeds |

Two sibling guards on one rule failing in opposite directions, in a module that argues at `:1170` that they must never disagree.

**The property, named:** *whenever a guard decides to deny and the JSON channel is unusable, the hook process refuses.* Not "this body does not raise".

- `_deny` returns the exit status the handler owes: `0` on the normal JSON path, `DENY_EXIT = 2` when the write fails. 2 is the harness's other refusal channel (blocks the call, hands stderr to the model); **every other non-zero status is a non-blocking error and the call goes ahead**, which is exactly how the 141 above let one through. The tests assert the number, not "non-zero".
- On the fallback path stdout is redirected to `/dev/null` at the fd level, because a failed shutdown flush *replaces* the exit status with 120 — another number in the "proceeds" bucket. Same move, and same guard, as `cli.main` already makes.
- The `except` in `pretooluse_read` was **narrowed, not deleted**: a payload charter cannot parse is still an allow, still silent, still not a broken turn. Deciding is fallible; refusing is not.
- An undelivered denial is also recorded, and `dispatch` refuses on it. That is the part that covers the guard nobody has written yet — a fail-open is invisible from the outside, so it is closed at the one place every guard passes through, not at each call site.

The coordinator asked for a sweep of the other entry points: `pretooluse_read` was the only one hiding a deny inside a `try`. `pretooluse_dispatch`'s `except Exception` wraps a *nudge* (`_ask`), which is the deliberate posture and stays. Two AST tests now pin the shape — no bare `_deny(...)` expression statement, and no `_deny` call inside a `try` — so the next one fails here rather than in an audit.

## #461 — `git checkout <file>` is a file restore, not a branch switch

`git restore charter.toml` was allowed the whole time. So charter was never holding a policy against restoring a file in the plane root; it was refusing **one of the two ways git spells an operation it permits** — and the denial was confident, detailed, and wrong about what the command does.

**The property:** git disambiguates by whether the operand resolves as a revision and by `--`. So charter asks git the same questions, in the same repo, instead of matching command shape:

- `rev-parse --verify --quiet <op>^{commit}` — does it resolve to a commit?
- `ls-files --error-unmatch -- <op>` — does git *track* a path by that name? (Not `os.path.exists`: an untracked file is not restorable, while `.`, `src/*.py` and `:/` are pathspecs that match plenty and exist as no single file.)

Only `path` — cannot be read as a commit, **can** be read as a tracked path — opens the gate. That is an affirmative demonstration, not the absence of a bad spelling, so everything unresolvable stays refused: a remote-only branch (git's DWIM checks it out here as a *new local branch* — verified), a name git has never heard of, `git checkout "$BR"`, and a git that times out.

| invocation | measured effect (git 2.50) | verdict |
|---|---|---|
| `git checkout README` | Updated 1 path from the index | allow |
| `git checkout .` / `notes` / `--ours README` | restore, HEAD unchanged | allow |
| `git checkout feature README` | Updated 1 path from `<sha>`, HEAD unchanged | allow |
| `git checkout feature` / `-` / `HEAD~0` | Switched to branch / detached | **deny** |
| `git checkout ambig` (file **and** branch) | "Switched to branch 'ambig'" — ref wins | **deny, as ambiguous** |
| `git checkout nosuch` / `"$BR"` | unresolvable → DWIM or error | **deny** |
| `git checkout --detach` / `git switch --detach` | "HEAD is now at …" | **deny (was allowed)** |

**The ambiguous case stays denied and now says so.** The message no longer asserts a branch; it says the command is ambiguous, that git breaks the tie in favour of the ref, and names the two spellings that are not ambiguous — `git restore <name>` and `git checkout -- <name>`, both of which are asserted allowed in the same test class, so the advice cannot rot.

**Two things the same misreading was hiding, found while fixing it:**

1. **Too narrow.** `--detach` takes the root off its branch with no operand at all, and the guard required an operand. Both `checkout --detach` and `switch --detach` walked past it.
2. **The next bypass, closed before it shipped.** charter's tokeniser flattens `git checkout $(echo feature)` into `['git','checkout','$','(','echo','feature',')']`. A rule that read "more than one operand ⇒ a restore" — which is what git's behaviour suggests — would have handed **every command substitution** a free pass at the branch guard. Trailing operands are resolved against git instead, bounded by `_MAX_CHECKOUT_OPERANDS`; `(` is not a tracked path. There is a test named for it.

Also fixed on the way past: `_plane_default_branch` could raise `ProcTimeout` straight out of the `PreToolUse` handler — a broken turn rather than a verdict.

## Overclaiming text, changed in the same commit

- `docs/hooks.md` → *When a hook fails* said flatly "Hooks swallow their exceptions … none of this is load-bearing." A denial is load-bearing, and the section now says which half is excused (deciding) and which is not (refusing), with the exit-2 contract and why the number cannot be "any failure".
- `docs/hooks.md` → *The guards* now states that a restore runs in the plane root, that `--detach` counts, and what happens on a shared name.
- Two `docs/news/unreleased-*.md` entries.

## Verification

- `python3 -m unittest discover -s tests` → **Ran 4574 tests … OK** (46 of them new; stdlib `unittest`, `PersonaIso`, no network, no non-stdlib import).
- Every git behaviour in the #461 tests was **measured** in a scratch repo with HEAD recorded before and after, including the DWIM and file-plus-branch cases — not inferred from the docs.
- **Mutation-tested, 11 mutants, all killed**, `__pycache__` cleared between runs, GREEN restored after each:

| mutant | killed by |
|---|---|
| `_deny` loses its exit-status fallback | 1 |
| `pretooluse_read` swallows its own deny again | 1 (the AST test — the `dispatch` backstop keeps behaviour correct, which is the redundancy working) |
| `dispatch` drops the undelivered-deny backstop | 1 |
| `DENY_EXIT` becomes 141 | 1 |
| any single operand counts as a restore | 8 |
| trailing operands assumed to be paths | 5 |
| `--detach` stops counting as a ref move | 3 |
| an unaskable git reads as a path | 2 |
| the ambiguous case described as a plain switch | 1 |
| an ambiguous operand counts as a restore | 3 |
| the restore carve-out reverted entirely | 6 |

## What is the next spelling that still gets through?

Stated rather than left for the next audit, and consistent with `SECURITY.md`'s "guard rails, not guarantees":

- **A command charter's tokeniser cannot see as a `git` invocation at all** — `sh -c 'git checkout feature'`, an alias, a script. That residual is older and wider than this guard and is unchanged by this PR.
- **A git alias defined on the command line** — `git -c alias.z='checkout feature' z` really does switch the branch (verified: *"Switched to branch 'feature'"*), and `_plane_root_git` yields `('z', [])`, which is in neither `_BRANCH_MOVERS` nor `_RESET_TREE_MODES`. Both plane-root guards are blind to it. Pre-existing and not reachable through anything this PR adds — the same shape `docs/hooks.md` already documents for the tool-gate (`git -c alias.z=clean z -xfd`) — so it is filed separately rather than guessed at here.
- **TOCTOU** — the guard asks git now and the command runs a moment later. An agent that creates a branch named after a tracked file between the two turns a `path` into a `both`; the window buys it a denial, not an allow, so it fails in the safe direction.

None of these are made reachable by this change; the set of allowed commands here strictly shrinks except for the file-restore forms.

### DIFF
--- a/charter/hooks.py
+++ b/charter/hooks.py
@@ -30,6 +30,7 @@ from __future__ import annotations
 import json
 import os
 import re
+import shlex
 import sys
 import time
 from pathlib import Path
@@ -132,26 +133,55 @@ def _deny(event: str, reason: str) -> int:
             "permissionDecision": "deny",
             "permissionDecisionReason": text,
         }})
+        # A WRITE IS NOT A DELIVERY, and this line is the whole of #438's second half.
+        # `print` to a PIPE block-buffers: the JSON lands in an 8KiB userspace buffer, the
+        # call returns cleanly, and the `except` below never runs — so the first version of
+        # this fix returned 0 on a real broken pipe and the tool call proceeded, byte for
+        # byte what it replaced. The failure surfaced only when the interpreter flushed at
+        # shutdown, too late for any handler to answer and worth exactly 120: a NON-blocking
+        # status. Only an unbuffered stdout — a terminal, or a test that stubs `print` —
+        # raised where the code expected it, which made the buffered pipe (the real hook's
+        # real stdout, one end of a harness pipe) the one shape nothing covered.
+        #
+        # Flushing here moves the answer to "did the harness get this?" back inside the
+        # `try`, where the guard can still act on it.
+        sys.stdout.flush()
     except Exception:  # noqa: BLE001 - the channel, not the verdict
         _undelivered_deny.append(text)
-        # stderr is what exit 2 hands to the model, so the reason still arrives.
+        # stderr is what exit 2 hands to the model, so the reason still arrives — and it is
+        # flushed for the reason above: buffered is not delivered, and a reason discovered
+        # to be undeliverable at shutdown is a refusal with no reason.
         try:
             print(text, file=sys.stderr)
+            sys.stderr.flush()
         except Exception:  # noqa: BLE001 - best effort; the exit status is the guard
-            pass
+            _mute(sys.stderr)
         # And stop the interpreter trying to flush the dead stream on the way out: a failed
         # shutdown flush REPLACES the exit status with 120, which is in the "tool call
         # proceeds" bucket — the same fd-level move, and the same guard, as `cli.main`.
-        try:
-            fd = os.open(os.devnull, os.O_WRONLY)
-            try:
-                os.dup2(fd, sys.stdout.fileno())
-            finally:
-                os.close(fd)
-        except Exception:  # noqa: BLE001 - no real pipe under test; nothing to suppress
-            pass
+        _mute(sys.stdout)
         return DENY_EXIT
     return 0
+
+
+def _mute(stream) -> None:
+    """Point *stream*'s file descriptor at ``/dev/null`` so a later flush cannot fail.
+
+    A `BrokenPipeError` raised while the interpreter flushes on the way out REPLACES the
+    process's exit status with 120 — a non-blocking hook error, i.e. an allow. So the last
+    thing :func:`_deny` does with a dead channel is make sure nothing tries to use it again:
+    the exit status is the only thing left carrying the refusal, and it has to survive
+    shutdown. Best effort by construction — under a test's `StringIO` there is no fd to
+    replace and nothing to suppress.
+    """
+    try:
+        fd = os.open(os.devnull, os.O_WRONLY)
+        try:
+            os.dup2(fd, stream.fileno())
+        finally:
+            os.close(fd)
+    except Exception:  # noqa: BLE001 - no real pipe here; nothing to suppress
+        pass
 
 
 #: The one host permission mode that means NOBODY IS WATCHING. Deliberately not a set that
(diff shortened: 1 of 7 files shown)


diazoxide:
## Round three — `e78e19f`

**The reported bypass is closed.** `git checkout --detach main` and its six siblings are
denied; verified against real git in the fixture (each answers "HEAD is now at <sha>" and
HEAD leaves `main`).

Sorted the way the review asked:

**(B) — the branch is not, and was not, weaker than `origin/main`.** Measured rather than
argued. `tests/…::TestNeverWeakerThanTheGuardThatShips` loads `charter/hooks.py` as it
stands on `origin/main` and runs the same corpus past both. Over the 281-row corpus plus the
88 route rows: 49 inputs this branch denies that main allows, 3 the other way — all three
`git checkout README`, none of them moving HEAD, which is #461's false positive being removed
on purpose. `origin/main` allows `git checkout --detach main` too; the hole is real and now
closed, but it is not a regression this branch introduced.

**(A) — fixed, with the reproduction each came from:**

1. The remedy carve-out gated on the operand's name. `restore_shaped` gates it now — the
   property "every option present is placed as restore-only", which is what makes a command
   leave HEAD *attached* to the default branch.
2. A relative `git -C` resolved against the hook process's cwd, so `git -C ../../.. checkout
   feature` from a clone moved the plane root's HEAD and was allowed. It now joins onto the
   cwd the walk is already tracking, which also gives git's own rule for repeated `-C`s.
3. `git switch -- feature` switched branches and was read as a restore. `--` introduces
   paths on a `checkout`; `switch` has no path half for it to introduce. **This one was
   found by the new generated corpus, not by anyone reading the code.**

**(D) — filed, not chased: #477.** `git --git-dir=<plane>/.git checkout <branch>` reaches the
root's refs from anywhere and neither plane-root guard follows it (verified end to end).
`origin/main` has it too. Named as a known remaining route in `docs/hooks.md` and the news
entry.

**(C) — narrowed in the same commit.** `docs/hooks.md`, ADR 0008 and the news entry all said
"returning to the default branch is always allowed", which was the sentence the bypass hid
behind; they now say *putting the root back **on*** its default branch, and name the
spelling. The denial message says the same thing and stops calling a detach a switch.

### The recurring failure, one level up

Every bypass so far came from matching a spelling. Round two fixed *who decides the verdict*
— real git, not a written-down answer — and left *who chooses the questions* alone: 41 rows
somebody typed. Two of the three holes above are combinations of axes that were each already
covered (`--detach` had rows, `main` had rows, `--detach main` had none). The corpus is now
the product of `subcommand × option × operand × where the --  goes`, so adding one option
adds every combination of it.

**The next spelling** is a route, not a flag: `--git-dir` / `--work-tree` / `GIT_DIR`
(#477), and after that `git worktree add` on the plane root's repository.

### Mutation testing

Six mutations, each red, each restored green, `__pycache__` cleared between:

| mutation | failures |
| --- | --- |
| carve-out gates on `not creating` again | 46 |
| relative `-C` resolves against the hook | 40 |
| the separator means paths on `switch` too | 15 |
| a detach with an operand is called a switch | 1 |
| the closing sentence promises a category | 1 |
| `--detach` is not an option class | 26 |

Two more aimed at the differential test alone, to show it is not vacuous: dropping `switch`
off `_BRANCH_MOVERS`, and making `--orphan` a restore-only option (round one's exact bypass)
— both caught there, and by that test only.

Full suite: `Ran 4661 tests … OK`.

Issue #457, opened 2026-08-24
frame: the top slot shows no dev chip, so a framed session on the dev channel says so nowhere


diazoxide:
Reported by the operator from a live frame, with a screenshot: the top slot reads `charter 0.51.0` on a plane whose `charter.toml` declares `[update] channel = "dev"`.

## What happens

`charter/frame/slots.py:_top` (installed build, line 156) renders:

```python
right = persona if verbosity(fid) == "terse" else f"{persona}  charter {__version__} "
```

`__version__` is the raw string. There is no `channel` import in that module.

## Why this is worse than a missing chip

#386 made a frame **suppress the status line** — `statusline.main` prints an empty line inside a live frame. And `statusline._brand()` is the one place the `dev` chip was added by #454.

So the two land on top of each other:

- outside a frame, on dev: `⬢ charter 0.51.0 dev` — correct
- **inside a frame, on dev: `charter 0.51.0`, and the status line that would have said `dev` is blanked**

The frame is the surface that replaced the status line, so on the dev channel a framed session has **no indication of its channel anywhere on screen**. That is exactly the state `_brand`'s own docstring argues against: it says the chip exists because the operator needs to know "whether `↑` means *a release was cut* or *main moved*", and a framed operator cannot know either.

## Fix

`_top` should render the channel the same way `_brand` does — the word `dev` beside the version, not the full `+dev (main @ sha)` label, which stays `charter --version`'s job.

Note `_top` already has a `verbosity(fid) == "terse"` branch that drops the version entirely. The chip should follow the version: shown when the version is shown, gone when it is not. Do not add it to the terse branch — terse exists to buy columns back.

## Test that would have caught it

A property test rather than a per-surface one: **every surface that renders a version renders the channel with it.** Today that is `statusline._brand` and `slots._top`; the next surface to grow a version string should fail it on the day it is added, the way `unimplemented()` catches an unregistered slot.

## Related

- #454 added the chip to `_brand` only.
- #386 made the frame own the surface.
- Neither is wrong alone; the gap is at their intersection, which is why no test in either PR could see it.

PR #492, opened 2026-08-24
The suite cannot read your update channel, and the report nudge names it (#459, #458)

### LINKED ISSUE #458: report: the staleness nudge shows charter's version but not its channel
Found while writing the property test for #457 (\"every surface that renders a version also renders the channel\").

`charter/commands_report.py:_warn_if_stale` prints:

```
util.warn(f"you are on charter {__version__}; {latest} is out — this may "
          "already be fixed. Reporting anyway is fine.")
```

This is the same idiom `statusline._brand` and `frame/slots.py:_top` had before #457 — the literal word `charter` glued directly onto the bare `__version__` — and it does not call the shared `statusline._dev_chip()` helper #457 introduced.

Whether this is worth fixing turns on the same question `_brand`'s docstring already answers for the status line: on the dev channel, "`{latest}` is out" is ambiguous between *a release was cut* and *main moved* — exactly the ambiguity the chip exists to resolve. A reporter on the dev channel reading "0.52.0 is out" may not know which of those it means, or that `charter update` on dev tracks `main` rather than the next tag.

## Fix

Same shape as #457: `util.warn(f"you are on charter {__version__}{statusline._dev_chip()}; {latest} is out — ...")`, importing `statusline` for the chip.

## Test

The property test added for #457 (`tests/test_version_shows_channel.py`) already detects this site — it is carried as a documented, named exception (`_KNOWN_GAPS`) pointing at this issue, specifically so fixing it here is a one-line change: implement the chip, delete the exception, and the property test itself proves the fix rather than needing a bespoke one.

Out of scope for #457 itself, which is specifically about the frame's top slot and the status line — filed separately per the "file the defect, don't fold it into an unrelated PR" rule.

### LINKED ISSUE #459: test: two suites don't isolate config.UPDATE, and fail for real on a dev-channel checkout
Found while verifying the suite for #457 on a clean `origin/main` worktree (`d08da07`) — these 6 failures are pre-existing and unrelated to that issue's change; confirmed by running the exact same tests against an unmodified worktree.

```
FAIL: test_brand_carries_the_indicator_only_when_newer (test_statusline_brand.UpdateIndicator)
FAIL: test_newer_version_is_reported (test_statusline_brand.UpdateIndicator)
FAIL: test_version_compare_is_numeric_not_lexical (test_statusline_brand.UpdateIndicator)
FAIL: test_a_failed_install_warns_and_never_raises (test_version_lock.AutoSync)
FAIL: test_drift_installs_the_locked_version (test_version_lock.AutoSync)
FAIL: test_the_message_says_the_running_process_is_still_the_old_build (test_version_lock.AutoSync)
```

## Root cause

Neither `test_statusline_brand.py:UpdateIndicator` nor `test_version_lock.py:AutoSync` is `PersonaIso` — they hand-roll their own isolation (`AutoSync.setUp` does `config.ROOT = self.root`, a bare `TemporaryDirectory()` with no `.git`; `UpdateIndicator` isolates its own cache file). Neither one touches `config.UPDATE`.

`config.UPDATE` is derived once, at whatever point `charter.config` first resolves a control plane, from `root.find_root()` — which, standing inside a `git worktree` with no `charter.toml` reachable some other way, redirects to the **main tree's** `charter.toml` (`root._plane_of`'s worktree-to-main-tree redirect, `root.py`). It is not re-derived when a test later does `config.ROOT = <tmpdir>`, because that line assigns only the one attribute.

So both classes silently read whatever channel the machine's REAL checkout declares — not a fixture. On a checkout whose main tree has `[update] channel = "dev"` (my own environment does — it's a live charter control plane, and this is exactly the plane #457's own bug report is about), `channel.is_dev()` is `True` where these tests assume the default `False`:

* `AutoSync`'s tests get `hooks._autosync_version_lock`'s dev-vs-pin CONFLICT message ("this control plane pins X AND declares `[update] channel = "dev"`. Those ask for two different charters...") instead of the plain drift-install message they assert on.
* `UpdateIndicator`'s tests presumably get `update.newer_than`'s dev-channel comparison (commit-based) instead of the stable one (semver-based) they assert on — not yet root-caused past `config.UPDATE` leaking; the failure is `update.newer_than("0.6.0")` returning `None` instead of `"9.9.9"`.

Reproduce: `cd <any git checkout of charter> && python3 -c "from charter import config; print(config.UPDATE)"` — on a machine whose real plane declares dev, this prints `{'channel': 'dev'}` with **zero test code involved**, before any test even runs.

## Why this was latent until now

`[update] channel = "dev"` is new (#454). Every contributor who hasn't opted their own real plane into it never triggered this — `config.UPDATE` defaulted to `{"channel": "stable"}`, which happens to match what these two classes assume, so the missing isolation was invisible. It stops being invisible the moment charter's own dogfood plane (or any contributor's) opts in — which is exactly what's happening as dev-channel adoption grows.

## Fix

`AutoSync` and `UpdateIndicator` should isolate `config.UPDATE` the way every channel-aware test in `tests/test_dev_channel.py` already does — `mock.patch.object(config, "UPDATE", {"channel": "stable"})`, or switch to `PersonaIso` (which resolves `config.ROOT` at a real tmp `charter.toml` via `config.use`, so `config.UPDATE` is genuinely re-derived rather than merely left stale).

Not fixed here — found while verifying an unrelated PR (#457) and it touches neither of these files; filing separately per the "file the defect, don't fold it into an unrelated PR" rule.

diazoxide:
## The defect

`charter update` on a charter checkout only refreshes the plugin when the plane is on the **dev** channel (`commands_update._update_dev_on_a_checkout`), so anyone working on charter has to put `[update] channel = "dev"` in their own `charter.toml`. That file is committed, and `charter save` commits it — so the moment the line lands, six tests go red for everyone. The feature 0.52.0 shipped was unusable by the people it shipped for.

Nothing is wrong with the channel. Two classes had never isolated `config.UPDATE`:

- `test_statusline_brand.UpdateIndicator` redirected `config.STATE_DIR` by hand;
- `test_version_lock.AutoSync` assigned `config.ROOT` directly.

Neither re-derives anything else, so `config.UPDATE` kept naming the channel of whatever plane the suite resolved. For as long as `[update]` has existed, `{"channel": "stable"}` was true on every machine — a fixture nobody wrote, silently agreeing with what every assertion assumed.

## The fix is structural, not per-test

Same shape as #402, which was closed by making the violation impossible (`RealPlaneWrite`) rather than by patching the one offending test.

`config.DERIVED` already covers `UPDATE`, so `PersonaIso`'s single `config.use()` call was always enough — the bug was that these classes were plain `TestCase`. So the guard is on the **read**:

- `tests/_planeguard.py` gains `RealPlaneRead`. Whenever `charter.config` points at the real plane, every name in `_GUARDED_SETTINGS` (today: `UPDATE`) holds a `dict` subclass that refuses `get`/`[]`/`in`/iteration/`keys`/`values`/`items`/`copy`/`==` — including the C fast paths `dict(x)` and `{**x}` take, which is why `__iter__` is overridden.
- **A `BaseException`**, for the reason the write half already documents: `statusline._dev_chip` wraps the read in `except Exception`, so a catchable tripwire would be reported as "not on the dev channel" and the test would pass, wrongly.
- **Armed per derivation, by ROOT**, not by a one-shot swap at import. `config.derive` is wrapped: a derivation landing back on the real plane re-arms (`config.use(config.ROOT)` is a real thing two modules do), a derivation from a tmp root does not. A guard that could be disarmed by the first isolating test would protect only the first half of a run.
- The message **names the test that made the read**, and both ways out: `PersonaIso`, or the new `tests._isolation.pin_update_channel(case)` — the companion to `isolate_state_dir` for cases that mean to run against the real plane.

`tests/test_no_test_reads_the_operators_channel.py` is the control module: it makes the refusal happen for real, checks the message, and exercises every documented way out, so "green" cannot mean "the tripwire is gone".

**Turning it on found eight more unisolated sites** — in `test_doctor_forge_visibility`, `test_doctor_mcp_launchers`, `test_doctor_personas`, `test_memory_index_health`, `test_news_cross_process`, `test_statusline_gauge` and `test_statusline_session_block` — all fixed here. That is the point of doing it this way.

## Why the six failures did not reproduce in a full run

They only failed when the two modules were run alone, which is what made this hard to believe (and what a previous look reported as "did not reproduce").

`test_secret_exec.SecretExecMode` defines a `_restore` method of its own. `PersonaIso.setUp` registered its cleanup as `addCleanup(self._restore)`, which binds by attribute lookup **on the instance** — so the subclass's method replaced it. Config was never restored, and `config.ROOT` stayed pointed at a deleted temp directory for the **1193 tests** that ran after it alphabetically. A dead plane declares no channel, which looks exactly like `stable`.

`PersonaIso`'s cleanup is name-mangled now (`_PersonaIso__restore`) so a subclass cannot shadow it; `tests/test_isolation_harness.py` pins that with a fixture that deliberately defines `_restore`. `test_secret_cp_destination` had been dodging the same collision by hand, with a comment warning the next person — that comment is now history rather than instruction.

## #458, folded in

`commands_report._warn_if_stale` glued the literal word `charter` onto the bare `__version__` with no channel beside it. On the dev channel the number after "is out" is `main`'s head commit, not a release, so `0.52.0 is out` could not tell you which. It calls `statusline._dev_chip()` now.

It was genuinely a few lines: the property test added for #457 already carried this site as a documented `_KNOWN_GAPS` entry precisely so the fix would be "implement the chip, delete the line". The entry is gone, the dict is empty (machinery kept for the next one), and `test_the_staleness_nudge_calls_the_chip_too` asserts the site by name rather than leaving it to a sweep that would also be green if the scanner stopped finding the file.

## Verification

Run from a clone whose own `charter.toml` declares `channel = "dev"` — where `config.ROOT` genuinely resolves to a dev-channel plane and the six failures reproduce — and again on stable.

**Dev channel** (`UPDATE = {'channel': 'dev'}`, `is_dev() == True`):

```
$ python3 -m unittest discover -s tests 2>&1 | grep -E "^(OK|FAILED|Ran |FAIL:|ERROR:)"
Ran 4907 tests in 218.315s
OK
```

**Stable channel** (`UPDATE = {'channel': 'stable'}`):

```
$ python3 -m unittest discover -s tests 2>&1 | grep -E "^(OK|FAILED|Ran |FAIL:|ERROR:)"
Ran 4907 tests in 202.199s
OK
```

Before the fix, on the same dev-channel tree: the two modules alone give 6 failures; the full suite gives 0, because of the `SecretExecMode` leak above. With the guard installed and the leak closed, the full suite surfaced 50 errors across 9 modules — all fixed here.

**Mutation-tested, both directions:**

1. Reverted `tests/test_version_lock.py` to `HEAD` (the unisolated `AutoSync`) and ran the module: 4 errors, each `RealPlaneRead` naming `tests.test_version_lock.AutoSync.<method>` and pointing at `PersonaIso`. Restored.
2. Set `_GUARDED_SETTINGS = ()` and ran the new control module: 16 failures. Restored — a control that cannot fail proves nothing.

stdlib only, `unittest` only, no network. No version bump, no news stamping, no tag; one `unreleased-` news entry.

Closes #459
Closes #458

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,10 +29,14 @@ command.
   that fails without the fix. The suite is plain `unittest` — no fixtures framework, no
   network, no writing outside a tempdir. That last one is enforced rather than asked for:
   a checkout inside a control plane resolves to *that* plane, so `tests/_planeguard.py`
-  refuses any write into the real `.charter/` and fails the test that tried. Derive from
-  `tests._isolation.PersonaIso` and it never comes up; a test that spawns a subprocess has
-  to hand it the throwaway plane as `$CHARTER_ROOT`, which no guard in this process can do
-  for you.
+  refuses any write into the real `.charter/` and fails the test that tried. The same file
+  refuses one kind of *read*: a setting your own `charter.toml` declares — today
+  `[update] channel` — because a test that reads it is asserting against a fixture written
+  by whoever happens to run the suite. Derive from `tests._isolation.PersonaIso` and
+  neither ever comes up; a case that runs against the real plane on purpose calls
+  `isolate_state_dir(self)` and `pin_update_channel(self)`; a test that spawns a subprocess
+  has to hand it the throwaway plane as `$CHARTER_ROOT`, which no guard in this process can
+  do for you.
 - **Comments explain *why*.** The codebase leans hard on this: a comment that restates the
   code earns nothing, one that records the failure a line prevents is worth several
   paragraphs of docs. Read a few modules before writing your first one.
--- a/charter/commands_report.py
+++ b/charter/commands_report.py
@@ -284,7 +284,13 @@ def _warn_if_stale() -> None:
         from . import statusline, update
         latest = update.newer_than(__version__)
         if latest:
-            util.warn(f"you are on charter {__version__}{statusline._dev_chip()}; {latest} "
-                      "is out — this may already be fixed. Reporting anyway is fine.")
+            # `util.color_enabled()`, not the chip's ANSI default: this is the one caller
+            # that is not a terminal surface, and `util.warn` gates its own glyph the same
+            # way. The call stays INSIDE the f-string on purpose — `test_version_shows_
+            # channel`'s AST property looks for it in the same `JoinedStr` as the version.
+            color = util.color_enabled()
+            util.warn(f"you are on charter {__version__}{statusline._dev_chip(color)}; "
+                      f"{latest} is out — this may already be fixed. Reporting anyway is "
+                      "fine.")
     except Exception:  # noqa: BLE001 - a staleness check must never block a report
         pass
--- a/charter/commands_report.py
+++ b/charter/commands_report.py
@@ -271,12 +271,20 @@ def _report_duplicates(dups: list[dict], rid: str) -> int:
 
 def _warn_if_stale() -> None:
     """Warn, never block. A bug that survives on an old charter is still worth having, and
-    "upgrade first, then report" is a reliable way to never get the report at all."""
+    "upgrade first, then report" is a reliable way to never get the report at all.
+
+    Carries the channel chip (#458). On the dev channel ``{latest}`` is not a published
+    release at all — `update.newer_than` hands off to `newer_head`, so the number beside
+    "is out" is `main`'s head commit — and "0.52.0 is out" read on a dev plane is ambiguous
+    between *a release was cut* and *main moved*. That is the exact ambiguity
+    `statusline._dev_chip` exists to resolve, and #457 already made it one function so a
+    third surface can call it rather than re-derive `channel.is_dev()` and its try/except.
+    """
     try:
-        from . import update
+        from . import statusline, update
         latest = update.newer_than(__version__)
         if latest:
-            util.warn(f"you are on charter {__version__}; {latest} is out — this may "
-                      "already be fixed. Reporting anyway is fine.")
+            util.warn(f"you are on charter {__version__}{statusline._dev_chip()}; {latest} "
+                      "is out — this may already be fixed. Reporting anyway is fine.")
     except Exception:  # noqa: BLE001 - a staleness check must never block a report
         pass
--- a/charter/statusline.py
+++ b/charter/statusline.py
@@ -1746,9 +1746,18 @@ def _session_strip(payload: dict, sid: str | None) -> str:
     return f"{_DIM} · {_R}".join(parts) if parts else ""
 
 
-def _dev_chip() -> str:
+def _dev_chip(color: bool = True) -> str:
     """`` dev``, dimmed, glued right after a version — when THIS plane is on the dev
     channel. Empty string otherwise, so a caller can interpolate it unconditionally.
+
+    *color* is for a caller that is not painting a terminal surface. The default is ANSI
+    because the two callers here are — a status line and a frame slot both write escape
+    codes on every line, and `sys.stdout.isatty()` is False for both anyway (the harness
+    reads the status line through a pipe), so gating this internally would blank the chip
+    exactly where it belongs. `commands_report._warn_if_stale` is the other shape: it
+    hands the chip to `util.warn`, which gates its own glyph on `stderr.isatty()`, and an
+    ungated chip put raw ``\x1b[2mdev\x1b[0m`` in a redirected log beside a correctly
+    plain ``!``. It passes `util.color_enabled()` so both halves decide together.
 
     Split out of `_brand` (#457) so a second surface that glues the literal word
     `charter` onto `__version__` — `frame.slots._top`, and whatever grows the idiom
@@ -1764,7 +1773,9 @@ def _dev_chip() -> str:
         dev = channel.is_dev()
     except Exception:
         dev = False
-    return f" {_DIM}dev{_R}" if dev else ""
+    if not dev:
+        return ""
+    return f" {_DIM}dev{_R}" if color else " dev"
 
 
 def _brand() -> str:
--- a/charter/util.py
+++ b/charter/util.py
@@ -19,6 +19,24 @@ def _c(code: str, text: str) -> str:
     return f"\033[{code}m{text}\033[0m" if _USE_COLOR else text
 
 
+def color_enabled() -> bool:
+    """Whether the loggers below will colour anything — i.e. stderr is a terminal.
+
+    Exported because a message BODY sometimes has to make the same choice the glyph
+    already makes. `_c` colours the glyph and nothing else, so a caller that interpolates
+    a pre-coloured fragment into `warn(...)` emits raw escape codes down a pipe while the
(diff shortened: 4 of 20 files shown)


diazoxide:
All ten addressed in `1b96049`. CI green on 3.11–3.14; suite green on **both** channels (4916 tests, `OK` on dev and on stable).

**1 — raw ANSI on a colour-stripping surface.** `_dev_chip` takes `color: bool = True`; `_warn_if_stale` passes the new `util.color_enabled()`. The default stays ANSI deliberately: `sys.stdout.isatty()` is False for *both* existing callers (the harness reads the status line through a pipe), so gating internally would blank the chip exactly where it belongs — the flag has to come from the caller that knows its own surface. The call stays inside the f-string, so `_calls_dev_chip` still sees it. Measured after:

```
! you are on charter 0.52.0 dev; 9.9.9 is out — this may already be fixed. …
```
zero escape codes, chip still `\x1b[2mdev\x1b[0m` on a tty.

**2 — "deleted" → the tmp survives.** Confirmed independently on `origin/main`: `exists: True`, `contents: ['personas']`, no `charter.toml`, `HAS_CONTROL_PLANE: False`. Five sites corrected, and each now states the mechanism (a root with no marker derives every setting to its default) rather than leaving the reader to infer it.

**3 — 1193 → 1186.** You are right and I can say where mine came from: it was the probe's count of leak *observations*, which included `SecretExecMode`'s own four and three nested probe runs. Re-measured here: 4890 discovered, indices 3700–3703, **1186 after** = 24.3%. Four sites.

**4 — the docstring overstated.** Rather than soften it, I probed every entry point: `pop`, `popitem`, `setdefault` **and `__reversed__`** all leaked. All four closed, refused *before* they mutate (new `test_a_refused_mutator_did_not_mutate_on_the_way_out`). `__repr__`/`__len__`/`bool` stay open and are now pinned as a choice, with the reason.

**5 — the silent hole in the silent-hole test.** Added `test_every_guarded_name_actually_derives_to_a_dict`, deriving from a tmp root so nothing is armed.

**6–9.** `pin_update_channel` raises `ValueError` outside `instance.UPDATE_CHANNELS` (`"deb"` previously pinned stable in silence); the nested fixture's docstring now says it is invisible *because* it is nested, and contrasts it with the module-level positive control in `test_the_suite_writes_no_trace_into_the_operators_plane` that genuinely relies on the naming rule; 113-char line reflowed; import order fixed.

**10.** Recorded in the `test_isolation_harness` module docstring, marked load-bearing: calling `_restore()` by name skipped the `enterContext(redirect_stdout(...))` exits and left `sys.stdout` bound to a `StringIO` for the rest of the process — "there is nothing to simplify back."

**Four new mutations, each red without its fix:** unconditional ANSI (3 cases), dropping `pop` (6, cascading — which is what pins the refuse-before-mutate property), guarding a `str`-valued setting (1), removing the channel validation (`"deb"` → `is_dev() == False`, silently).

Issue #408, opened 2026-08-23
frame: a panel that dies inside the operator's own tmux never comes back


diazoxide:
Found while rebasing #382 onto #381, and deliberately left unfixed there rather than invented as a side effect of a conflict resolution.

## What happens

#382 gives panels a `pane-died` hook that respawns them with backoff. #381 added a second launch path: when the operator is already inside tmux (`\$TMUX` is set), charter adds a window to *their* server instead of nesting a private one.

Those two do not meet. On the operator-tmux path, panels are drawn but **no pane-died hook is armed**, so a panel that dies there stays dead for the life of the frame — no message, no respawn, no backoff.

## Why it was not simply extended

`_panel_died_hook_argv` and `cmd_respawn` both hardcode `["tmux", "-L", SOCKET, ...]`. The operator's server is reached by socket **path** (`-S`), never by name (`-L`). Arming the hook as written would install a `run-shell` action pointing at a *different tmux server* — one that may not exist, or worse, may be some other frame's private server. That is a wrong fix that looks right in a diff, so it was backed out and `_draw_panels` left exactly as #381 wrote it.

## What the fix needs

- `cmd_respawn` has to learn which server it is talking to — `state.frame_server(fid)` already records it (#381).
- The private-server liveness check (`has-session`) has no meaning on a shared server the operator also uses; it needs a windows-based check instead.
- Both `-L name` and `-S path` argv shapes must be built from one place, so the next person cannot reintroduce the mismatch.

## Scope

Only affects `charter <harness>` launched from inside an existing tmux session. The common case — a plain terminal — takes the private-server path, where respawn works and is tested.

PR #500, opened 2026-08-24
frame: the repo table moves to a variable-height bottom, and three limits the frame carried are closed (#488, #475, #420, #413)

### LINKED ISSUE #413: frame: suppression keeps recording token usage, but nothing inside a frame renders it
Found and reported honestly by the agent that implemented #386, as a correction to that issue's own premise.

## The gap

#386's stated reasoning was that disabling `charter statusline` inside a frame "would destroy the data source for a renderer we just ported". That is **not true today**. `_context_gauge` was never actually ported into a slot — `slots._top` deliberately excludes it, and #385's docstring explicitly hands that decision on to #386, which then did not make it either.

So after #386, inside a frame:

- `record_usage` still runs on every status-line invocation and still writes the history — correctly, and it must not be unwired;
- but **no renderer anywhere in the frame reads it**.

A framed Claude Code session therefore has no `ctx`/`cache` gauge at all. #386's news entry and ADR record the honest version of this rather than the flattering half, but the capability gap is real and remains.

This also means the previously-recorded limit — "codex and opencode still get no context gauge" — understated it. Claude Code in a frame doesn't get one either.

## Why it was not simply added

It needs a design decision no issue has made. The usage file is keyed by **Claude Code's session id**, which a panel process does not know: a panel only ever sees `$CHARTER_SESSION_ID`, which the frame launcher sets to the **frame id**. The one place both ids are visible at the same moment is the suppressing `statusline.main` — so that is the only component that could write the mapping a panel would need to read.

Inventing that mapping as a side effect of #386 would have been a design decision smuggled in under a bugfix, so the implementer correctly declined and reported it instead.

## What a fix needs to decide

1. Who writes the frame-id → session-id mapping, and where it lives (it is per-session, machine-local, and must not leak between planes).
2. Which slot draws the gauge. `_top` currently rejects it with a long, correct argument about `_context_gauge({})` being structurally dead without a live payload — that argument stops applying the moment a panel has a file to read, so that docstring must be revisited rather than left contradicting the new behaviour.
3. What a panel shows when the mapping is absent or stale. **A gauge that silently reads zero is worse than no gauge** — that rule is already established in `slots._top` and must hold here.

Depends on #386 (merged). Possibly folded into #387 if it fits cleanly there; if not, this stands on its own.

### LINKED ISSUE #420: frame: the spinner only tracks dispatches, because inflight records have no kind
Reported by the agent that implemented #387, as a limit it could not deliver as written rather than one it worked around.

## What #387 promised

> "A spinner while a dispatch, **clone or `gl-refresh`** runs; complete stillness at idle."

## What shipped

Only dispatches animate. `inflight.start` has exactly **one** caller — the `Task`/`Agent` PreToolUse handler — so clones and `gl-refresh` are not tracked at all and the spinner cannot know about them.

## Why it was not simply wired up

The same records feed the **dispatch-overlap nudge** through `still_running()`. Adding a record named `clone` would make that nudge tell an operator:

> "`x` writes code and `clone` are already running"

— which is wrong, and wrong in the confident, human-readable way that is worse than silence.

Making this work needs one of:

1. a **kind field** on the inflight record, with every consumer taught to filter on it (`still_running()` wants dispatches only; the spinner wants anything live); or
2. a **second channel** for non-dispatch work, leaving inflight records meaning exactly what they mean today.

Option 1 is probably right — it keeps one file-based mechanism rather than two — but it changes a record format that other code reads, so it is a design decision, not an implementation detail. #387's implementer was explicitly told not to invent design decisions, and correctly reported this instead.

## Note

The animation mechanism itself shipped and is sound: gated so an idle tick costs exactly one `stat` (pinned by `test_an_idle_tick_costs_exactly_one_stat`, a syscall-count assertion rather than a timing one), measured at ~5,100 ns added per tick versus 17,307 ns for an ungated `live_records()` call. Widening what feeds it does not require touching any of that.

### LINKED ISSUE #475: _resize_hook_argv interpolates a disk-read pane id into a hook action with no shape check
`_resize_hook_argv` interpolates a pane id read off DISK into a hook action with no shape check — `_relayout` skips `_PANE_ID_RE` for every slot it KEEPS.

Adjacent surface to #408. That ticket's fix hardened `_panel_died_hook_argv`, which now
refuses a pane id, slot or frame id that is not tmux's own word for it before
interpolating any of them into a `pane-died` action string. The `window-resized` hook next
to it has the same construction and none of the guard, and this is present on `main`
today — it is not a regression from the #408 branch, which is why it is filed rather than
folded into that PR.

## Where

`charter/commands_frame.py`, `_relayout`:

```python
for slot, pane_id in panels.items():
    if slot in want:
        keep[slot] = pane_id        # <-- no _PANE_ID_RE check on this path
        continue
    if not _PANE_ID_RE.fullmatch(pane_id):
        # Read back off disk, so it is not tmux's own word for it any more...
        continue
```

The check guards only the slot being KILLED. Every slot the new density still wants goes
into `keep` unexamined, and `keep` is handed to `_install_resize_hook` →
`_resize_hook_argv`, which builds the action by string interpolation:

```python
action = " ; ".join(
    f"resize-pane -t {pane} {_RESIZE_FLAG[slot]} {layout.SLOT_SIZE[slot]}"
    for slot, pane in panes.items())
```

`_resize_hook_argv`'s own docstring states the assumption explicitly — "Pane ids are
tmux's own (`%<digits>`), never operator- or plane-derived, so this action is safe to
build directly" — and on the `_relayout` path that is no longer true. `panels` is
`state.panes(fid)`, read back out of the frame's directory on disk, exactly the provenance
`_panel_died_hook_argv`'s docstring names as the reason it checks every value.

## Reproduction

Through the real `_relayout`, tmux stubbed only so the issued commands can be read back:

```python
issued = []
def fake_run(what, cmd, **kw):
    issued.append(list(cmd))
    return subprocess.CompletedProcess(cmd, 0, stdout="%9", stderr="")

with mock.patch("charter.commands_frame.tmuxctl.run", side_effect=fake_run), \
     mock.patch("charter.commands_frame._relayout_pane_env", return_value=None):
    commands_frame._relayout("charter", fid="demo-1", harness_pane="%0",
                             panels={"top": "%1;kill-server"}, want=["top"], v=(3, 7))
```

The `window-resized` hook charter installs comes out as:

```
resize-pane -t %1;kill-server -y 1
```

`;` is a tmux command separator, and the action is re-parsed as a command line every time
the window is resized — so this arms `kill-server` on every resize for the life of the
window. It is also below the builder, so the same value reaches `_resize_hook_argv`
directly:

```
>>> _resize_hook_argv(socket="charter", harness_pane="%0",
...                   panes={"top": "%1 ; kill-server ; resize-pane -t %1"})[-1]
'resize-pane -t %1 ; kill-server ; resize-pane -t %1 -y 1'
```

`_PANE_ID_RE.fullmatch` refuses both values. Nothing calls it on this path.

## Reachability

`state.panes(fid)` is a file under the frame's own state directory. Whoever can write
there can put an arbitrary string in it; charter re-reads it on the next `charter frame
density …` keypress and installs it. This is the same trust boundary `_relayout`'s own
comment already names on the line below ("Read back off disk, so it is not tmux's own word
for it any more"), so the question is not whether the boundary exists but why one branch
of the same loop honours it and the other does not.

## The property, and the next spelling of it

The property is *nothing is interpolated into a tmux action string unless it has been
checked against the shape tmux itself mints*. Provenance is not the check —
`_panel_died_hook_argv`'s docstring already says "every value that reaches the text is
what decides, never where it came from" — and `_resize_hook_argv` is currently the
counter-example to its own neighbour.

The suggested fix is to hoist the existing check above the `want` branch so it guards
both, and additionally to make `_resize_hook_argv` refuse rather than trust: a builder
that documents "safe because my caller checked" grows a second caller. `_relayout` is
already the second caller that made this true for `_arm_panel_respawn`.

Next spelling to watch for: `layout.SLOT_SIZE[slot]` and `_RESIZE_FLAG[slot]` are keyed
lookups so a bad slot raises rather than interpolates, but `_install_resize_hook`'s
`harness_pane` reaches `set-hook -w -t <pane>` as argv (safe today) — a future action that
mentioned the harness pane in its TEXT would need the same treatment, and `cmd_density`'s
own `_PANE_ID_RE.fullmatch(harness_pane)` is the only thing standing there.

### LINKED ISSUE #488: frame: move the repo table to a variable-height bottom slot, and stop drawing a lesser copy in a 22-column sidebar
Requested by the operator after living with the frame: *"move repos to bottom line like now our status-line, to show full repos experience that now we have."*

## Not a bug — a layout decision

First, what is NOT broken: `slots._left` renders correctly. Probed against a workspace holding a clone, it returns `charter main ↓9 ✓` plus its worktree rows. The reported "always empty" left sidebar was the **`default` workspace, which has 0 cloned repos** — the panel was telling the truth.

The real problems are two, and both are about where repos live:

1. **A 22-column sidebar cannot show what the status line shows.** `_left`'s own docstring concedes it: `_NAME_W` (32) and `_BRANCH_W` (34) alone exceed the whole pane's width, so it recomposes a narrower thing. The operator has a rich wide table in the status line and a lesser one in the frame — and the frame is now the surface that *replaces* the status line (#386).
2. **In a workspace with no clones, `left` is 22 columns of nothing**, permanently, taking width from the harness.

## The constraint that makes this real work

```python
SLOT_SIZE = {"top": 1, "bottom": 1, "left": 22, "right": 22}
```

`bottom` is **one line**. The wide repo table is many. So this is not a matter of calling a different renderer — `bottom` has to become a **variable-height** slot, sized to its content, with a floor and a ceiling.

## What a design needs to settle

- **How tall is `bottom`?** Probably `min(rows_needed, cap)` where the cap leaves the harness a usable pane. `_pick_rows` already ranks repos so the visible subset is the useful one — reuse it rather than truncating arbitrarily.
- **What happens to `left`?** Either it goes (freeing 22 columns) or it earns its place with something narrow that is genuinely per-row. Do not keep it drawing a lesser copy of what `bottom` now draws.
- **`bottom` currently carries the alert, session news, the todo count and the plane-root warning.** Those must survive — they are the "what wants attention" line and they matter more than a repo row. The repo table joins them; it does not evict them.
- **Density presets (#387) must stay coherent.** `minimal` / `normal` / `full` expand to slot lists, and `full` is `["top","bottom","left","right"]` in an order that IS geometry (measured: 200-column bottom vs 154). A variable-height bottom changes that arithmetic and the presets have to be re-derived, not patched.
- **Resize.** `_install_resize_hook` re-asserts pane sizes on `window-resized`; a content-sized pane has to recompute its height there, not just its width.
- **The idle-cost property must hold.** #387 pinned a panel's idle tick at exactly one `stat`. A taller panel redrawing more must not turn that into per-row filesystem work — `_left` already reads only `gather.read(fid)` and never touches a repo directory, and whatever draws the table must keep that rule.

## Why this is worth doing

The frame's whole claim is that it shows the plane's live state better than a single status line could. Right now it shows *less* of the repo state than the status line it suppresses. That is the one place the frame is a downgrade.

diazoxide:
The frame's claim is that it shows the plane's live state better than the single status
line it suppresses. In one place it showed *less*, and the operator said so:

> move repos to bottom line like now our status-line, to show full repos experience that
> now we have.

Four issues, one surface.

## #488 — the repo table moves to a variable-height `bottom`, and `left` is retired

**Not a bug first, a measurement.** `slots._left` rendered correctly — probed against a
workspace holding a clone it returned `charter main ↓9 ✓` plus worktree rows. The reported
empty sidebar was the `default` workspace, which has **0 cloned repos**: the panel was
telling the truth. What was broken is that 22 columns cannot show a table whose name and
branch columns alone want 66, which `_left`'s own docstring already conceded.

`bottom` is variable-height now and draws that table at `statusline.py`'s own widths —
same four columns, same markers, same CI glyphs, same `…(+N more)` wording:

```
0 todos · ⚠ reinit: charter ws reinit needed · F2 menu
  ├─ charter ⑂3                        main↓9                              ✓ passed
  ├─ iam-service                       feat/rotate-signing-keys*↑2         ✗ failed      #412
  └─ billing-gateway ⑂1                main                                ● running
```

**How tall** (`layout.bottom_rows`, pure arithmetic): what its content wants, floored at
the one row it always was, capped so the harness keeps `HARNESS_MIN_ROWS`. The cap is not
theoretical — measured against tmux 3.7c, `resize-pane -y 40` in a 20-row window is *not
refused*: tmux grants it out of the neighbour, and the neighbour is the agent session. It
left the harness pane **1 row tall**.

**`left`'s fate: dropped**, from `FRAME_SLOTS`, `SLOT_SIZE`, `SLOTS`, `_DROP_ORDER` and
`full`. It was drawing a lesser copy of what `bottom` now draws; the 22 columns go back to
the harness. A committed `charter.toml` still naming it is filtered like any unknown slot,
which is pinned with the exact list that shipped as the default one release ago.

**The constraints held:**

- The attention row survives — alert, `_session_news`, todo count, plane-root warning — as
  line 0, always. The table joins it.
- Presets re-derived: `full` means "every edge charter draws", and `minimal`/`normal` now
  differ in how many of `bottom`'s ROWS the table gets (`_TERSE_ROWS`), because `bottom`
  is no longer one row and "one-line top and bottom" had stopped being true.
- Idle cost unchanged: a panel's idle tick is still exactly one `stat`. `bottom` is the one
  animated slot, so a table walking a directory per row would pay that back fourteen times
  over at five repaints a second — `statusline._tree_cells` is deliberately *not* called
  (it ends in a `worktree.locate`/`clone_of` pair per row), and the cost is stated: the
  frame's table has no presence column. `gather.row_count` answers the sizer from the cache
  when there is one and a directory listing when there is not, never a git sweep.
- Resize recomputes the HEIGHT — see below.

A pane narrower than the table's 95 columns draws **no** table rather than a cut one: a row
trimmed past the branch loses its CI glyph and open change, and a failing repo then reads
as clean. The shipped `min-cols` is 100.

## #475 — collapsed rather than patched

The `window-resized` hook used to carry the sizes as literal `resize-pane -t %1 -y 1` text.
A content-and-window-sized `bottom` cannot be a constant there, so the action now calls
charter back (`charter frame-resize`, `run-shell -b`, the same construction #408 hardened
for `pane-died`) and the sizes are recomputed against the window that exists.

That removes #475's class, not its instance: **no pane id reaches that text at all**. On
top of it, the two guards it was actually about:

- `_relayout`'s `_PANE_ID_RE` check is hoisted **above** the `want` branch. It guarded only
  the slot being killed, so every slot the new density KEPT went into `keep` unexamined —
  a `%1;kill-server` in the frame's own state directory armed `kill-server` on every resize
  for the life of the window.
- `_reassert_sizes` refuses a bad id itself rather than documenting "safe because my caller
  checked". #475 *was* such a builder growing a second caller; `cmd_resize` is that second
  caller, reading `state.panes` straight off disk.

Each is pinned by a test the *other* guard cannot satisfy (the first asserts the recorded
pane map, not the argv), because the first draft of both passed with either one deleted.

## #420 — records say what kind of work they are

#387 promised a spinner for "a dispatch, clone or `gl-refresh`" and shipped dispatches,
because `inflight.start` had one caller. The blocker was real: the same records feed the
dispatch-overlap nudge, which reads its answer back as a sentence — a record named `clone`
would have produced *"`x` writes code and `clone` are already running"*.

Every record carries a `kind` now, every reader says which kinds it means, **and the
default is `DISPATCH`**: the readers that put a name in front of an operator get that by
*not asking*, so the next kind cannot leak into a sentence by being forgotten at one call
site. The frame's spinner is the single caller that opts into `kind=None`.

`finish` matches on kind as well as name — the filename carries only the agent, so a clone
of a repo called `steward` and a dispatch to a persona called `steward` globbed identically
and whichever ended first retired the other's record.

## #413 — `top` gets the context gauge back

0.52.0's news named this as the one thing the frame genuinely lost. The blocker was that
the usage history is keyed by *Claude Code's* session id and a panel knows only the
*frame's*; the one process that sees both is the suppressing `statusline.main`. It writes
the mapping (`state.record_harness_session`, in the frame's own directory — per-plane,
reaped with the frame), and `slots._top` reads it back.

One figure had to start being written down: `context_window.used_percentage` exists only in
the payload, while the cache ratio and rebuild count are derivable from the token counts.
Rows grew a fourth field, readable both directions across an upgrade. Both surfaces share
`_ctx_part`/`_cache_part`, so a green 60% in a frame and in a footer mean the same thing.

**Every way of not knowing draws nothing** — no mapping, no turns yet, a history predating
the field, a recycled frame id (`clear_shape` takes the mapping with it) — because a gauge
silently reading zero is worse than no gauge, and one reading somebody else's 78% is worse
than either.

The panels are woken once per turn, not once per render: "did anything change" is asked of
the usage file's `(mtime, size)` across the record call — `_record_turn` writes only when
it appends — never of an in-process memo, which would be empty on every invocation since
each one is a whole process.

ADR 0019's consequence bullet is **amended in place**, not rewritten: it predicted exactly
this shape and stays as the record of why the decision was deferred.

## The property, and the next spelling of it

The recurring failure here is matching a spelling instead of the property, and #475 is the
seventh: the guard checked the *branch* an id took rather than the *value*. The fix names
the property twice over — the check is hoisted so it covers both branches, and the class is
removed by not putting a disk-read value into re-parsed text at all.

**The next spelling to watch for:** `layout.slot_sizes` interpolates an `int` into a
`resize-pane` argv, which is safe by type — but `cmd_resize` now reads `state.panes`'s
**keys** as well as its values, and a slot name reaches `_RESIZE_FLAG[slot]`. That is a
keyed lookup today (a bad slot raises rather than interpolates) and it is guarded by
`slot not in _RESIZE_FLAG: continue`. A future action that put a slot NAME into hook text —
a per-slot `pane-died` message, say — would need `_action_word_is_safe` on it, and nothing
would force that. The other one: `_resize_hook_argv` now trusts the frame id in action
text, so anything that widens `_FRAME_ID_RE` widens that surface with it.

## Verification

- Full suite green: `Ran 4967 tests … OK` (stdlib `unittest`, no pytest, no network).
- The real tmux integration proves both halves of the resize path against tmux 3.7c:
  `_reassert_sizes` holding sizes across four real resizes including one where the cap
  binds, and — separately — the installed hook actually firing and its `charter frame-resize`
  child actually recomputing. That one asserts on a **grow**, because on a shrink tmux's own
  redistribution makes `bottom` smaller by itself and "it got smaller" would pass with no
  hook at all.
- **41 mutations applied, all RED**, each restored with `__pycache__` cleared. Two came back
  GREEN on the first pass and both were fixed rather than explained: the #475 tests passed
  because a *different* guard caught the input, and one redundant condition turned out not to
  be a guard at all (its `try/except` already was, so the code was simplified and the
  exception mutated instead). The per-issue lists are in the commit messages.

Closes #488
Closes #475
Closes #420
Closes #413

### DIFF
--- a/charter.toml
+++ b/charter.toml
@@ -11,7 +11,12 @@ share = "local"
 default = "steward"
 
 [frame]
-# All four slots: repos down the left, persona chips down the right, the wide repo
-# table across the bottom. `left`/`right` need a terminal at least `min-cols` wide —
-# below that `layout.visible_slots` drops them rather than cramping the harness.
-slots = ["top", "left", "right", "bottom"]
+# Every slot: identity across the top, the wide repo table across the bottom, persona
+# chips down the right. `right` needs a terminal at least `min-cols` wide — below that
+# `layout.visible_slots` drops it rather than cramping the harness.
+#
+# The ORDER is the geometry (#488): `bottom` is split before `right`, so its table gets
+# the full window width. Listed after it, the same table would be inset to 177 columns
+# and start losing its own right-hand columns. There is no `left` any more — it drew a
+# 22-column copy of the table `bottom` now draws properly.
+slots = ["top", "bottom", "right"]
--- a/charter.toml
+++ b/charter.toml
@@ -16,7 +16,9 @@ default = "steward"
 # `layout.visible_slots` drops it rather than cramping the harness.
 #
 # The ORDER is the geometry (#488): `bottom` is split before `right`, so its table gets
-# the full window width. Listed after it, the same table would be inset to 177 columns
-# and start losing its own right-hand columns. There is no `left` any more — it drew a
-# 22-column copy of the table `bottom` now draws properly.
+# the full window width. Listed after it, `bottom` is split off a harness the sidebar has
+# already narrowed — 23 columns short of the window, so the table needs a 118-column
+# terminal before it is drawn at all, and below that the pane is the attention row alone
+# (#500). There is no `left` any more — it drew a 22-column copy of the table `bottom`
+# now draws properly.
 slots = ["top", "bottom", "right"]
--- a/charter/cli.py
+++ b/charter/cli.py
@@ -705,7 +705,8 @@ def _add_frame_parsers(sub) -> None:
     # naming `frame-probe` silently launching a harness instead of reading tmux's own
     # version (see `commands_frame.cmd_probe`'s own docstring).
     _core_commands = set(sub.choices) | {"frame", "panel", "frame-menu", "frame-action",
-                                         "frame-probe", "frame-respawn", "frame-density"}
+                                         "frame-probe", "frame-respawn", "frame-density",
+                                         "frame-resize"}
 
     # Which harness (by `.name`, never `.cli_name` — that's the dict key below) has
     # already claimed each word, so a SECOND harness wanting it is told who got there
@@ -791,6 +792,19 @@ def _add_frame_parsers(sub) -> None:
     rs.add_argument("--pane", dest="pane", required=True)
     rs.add_argument("--frame", dest="frame", default=None)
     rs.set_defaults(func=commands_frame.cmd_respawn)
+
+    # Internal, and a top-level sibling for the same `_split_frame_argv` reason as the
+    # ones above. Fired by the frame window's own `window-resized` hook (#488,
+    # `commands_frame._resize_hook_argv`) — never typed by an operator. The hook used to
+    # carry the sizes as literal text; `bottom` is content-sized now, so the sizes have
+    # to be RECOMPUTED against the window that just changed, and only charter can do
+    # that. `--frame` travels on the argv for `frame-respawn`'s reason: on the operator's
+    # own server there is no `$CHARTER_SESSION_ID` for a `run-shell` child to read.
+    # Optional, not required, so a hook installed by an older charter and still sitting
+    # in a running frame's window options resolves its frame from the environment.
+    rz = sub.add_parser("frame-resize")
+    rz.add_argument("--frame", dest="frame", default=None)
+    rz.set_defaults(func=commands_frame.cmd_resize)
 
     # Internal, and a top-level sibling for the same `_split_frame_argv` reason as the
     # three above. Fired by a hotkey-menu selection (`commands_frame._menu_entries`), whose
--- a/charter/commands.py
+++ b/charter/commands.py
@@ -422,8 +422,25 @@ def _clone_one(r: dict, wd) -> dict:
     which is the whole point), `registry.for_repo` builds a fresh backend per call, and
     `gitpolicy.apply` writes only inside this clone's own `.git/config`. Nothing here is
     shared between repos.
+
+    **The network half is recorded as in flight (#420).** #387 promised the frame a
+    spinner while "a dispatch, clone or `gl-refresh`" runs and only dispatches animated,
+    because `inflight.start` had one caller. This is one of the two that were missing. The
+    record is per REPO rather than per command, so eight parallel clones read as eight —
+    `inflight`'s own `mkstemp` naming exists for exactly that concurrency, and this
+    function is already the per-repo, thread-local unit.
+
+    `kind=inflight.CLONE`, which is what keeps the dispatch-overlap nudge out of it: that
+    nudge reads names back to an operator as a sentence, and `inflight.still_running`
+    answers dispatches unless asked otherwise (see that module's docstring).
+
+    Recorded only around the parts that take time — after the destination and the URL are
+    settled, so a refusal never leaves a record, and released in a `finally`, so a `git`
+    that raises does not either. A process KILLED mid-clone still leaves one, and that is
+    the case `PRESUMED_DEAD_SECONDS`/`PRUNE_SECONDS` already exist for: the frame draws
+    such a record statically (`⋯`), never spinning.
     """
-    from . import gitpolicy
+    from . import gitpolicy, inflight
     from .forge import registry
 
     # #325: the destination is a name out of `inventory/repos.json`, a TRACKED file, and
@@ -446,15 +463,19 @@ def _clone_one(r: dict, wd) -> dict:
                 "reason": "its inventory record carries no HTTPS clone URL, and the "
                           "`ssh_url` it does carry is not a form this forge recognises — "
                           "charter will not hand git a string it did not build"}
-    proc = _git([*_cred_flag(forge), "clone", "--branch", r["default_branch"],
-                 "--", url, str(dest)])
-    if proc.returncode != 0:
-        return {"repo": r, "dest": dest, "status": "failed", "forge": forge,
-                "stderr": (proc.stderr or "").strip()}
-    # Golden rule 0: every git op from this clone uses ITS FORGE's token over HTTPS —
-    # credential helper + signing off + SSH→HTTPS rewrites (see charter/gitpolicy.py).
-    gitpolicy.apply(dest)
-    return {"repo": r, "dest": dest, "status": "ok", "forge": forge}
+    inflight.start(dest.name, kind=inflight.CLONE)
+    try:
+        proc = _git([*_cred_flag(forge), "clone", "--branch", r["default_branch"],
+                     "--", url, str(dest)])
+        if proc.returncode != 0:
+            return {"repo": r, "dest": dest, "status": "failed", "forge": forge,
+                    "stderr": (proc.stderr or "").strip()}
+        # Golden rule 0: every git op from this clone uses ITS FORGE's token over HTTPS —
+        # credential helper + signing off + SSH→HTTPS rewrites (see charter/gitpolicy.py).
+        gitpolicy.apply(dest)
+        return {"repo": r, "dest": dest, "status": "ok", "forge": forge}
+    finally:
+        inflight.finish(dest.name, kind=inflight.CLONE)
 
 
 def cmd_save(args) -> int:
@@ -614,7 +635,23 @@ def cmd_gl_refresh(args) -> int:
     if not dirs:
         util.info(f"No repos in workspace '{ws}'.")
         return 0
-    cache = glstate.refresh(dirs)
+    # In flight while the fetch runs (#420) — the second of the two callers #387's
+    # spinner was promised and never got. Recorded HERE, in the process that does the
+    # work, never in the `--detach` branch above: that branch returns immediately, and a
+    # record taken there would clear before the child it started had begun. `glstate
+    # .maybe_spawn` starts that child on the status line's own render path, so an
+    # operator sees the frame move while their forge state is being fetched rather than
+    # wondering why a column changed by itself.
+    #
+    # The WORKSPACE is the agent name, not a repo: one refresh covers every tree in it,
+    # and naming one of them would be a claim about which. `kind=inflight.REFRESH` keeps
+    # it out of the dispatch nudge and the persona chips.
+    from . import inflight
+    inflight.start(ws, kind=inflight.REFRESH)
+    try:
+        cache = glstate.refresh(dirs)
+    finally:
+        inflight.finish(ws, kind=inflight.REFRESH)
     util.ok(f"Refreshed forge state for {len(dirs)} tree(s) in '{ws}'.")
     for d in dirs:
         ent = cache.get(str(d), {})
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -1067,9 +1067,9 @@ def _wait_for_harness(socket: str, harness_pane: str) -> int | None:
         time.sleep(_POLL_SECONDS)
 
 
-def _launch_sizes(fid: str, slots: list[str], rows: int) -> dict[str, int]:
-    """How big each of *slots* is split, in a *rows*-row window — the launch-time half of
-    `layout.slot_sizes`, shared by both launch paths so they cannot disagree.
+def _launch_sizes(fid: str, slots: list[str], cols: int, rows: int) -> dict[str, int]:
+    """How big each of *slots* is split, in a *cols* x *rows* window — the launch-time
+    half of `layout.slot_sizes`, shared by both launch paths so they cannot disagree.
 
     `bottom` is the only slot whose answer is not a constant (#488): it is sized to the
     repo table it is about to draw, floored at the one row it always was and capped so
@@ -1078,13 +1078,21 @@ def _launch_sizes(fid: str, slots: list[str], rows: int) -> dict[str, int]:
     coming up with a pane taller than its content or a table cut off with nothing saying
     so.
 
+    **`cols` is not decoration (#500).** The renderer draws NO table below
+    `statusline._LEFT_W` (95) and at most `slots._TERSE_ROWS` of one at a `terse`
+    density, and neither is a rare shape: `layout.visible_slots` keeps `bottom` down to
+    `min_cols // 2`, so an 80-column terminal is one, and `minimal` is a level the F2
+    menu offers. Sizing from the repo count alone gave both of them a pane sized for a
+    table the panel then refused to draw — up to fourteen blank rows taken off the
+    harness. `bottom` is split BEFORE `right`, so the window's width IS the pane's.
+
     Affordable at launch by construction, and that is not incidental: `cmd_launch` calls
     `gather.discard(fid)` before it draws anything, so this reaches `gather.row_count`
     with no cache and gets the directory-listing answer — an `iterdir`, never the git
     sweep `gather.scan` would run. See `gather.row_count`'s own docstring for both paths.
     """
     return layout.slot_sizes(slots, window_rows=rows,
-                             content_rows=frame_slots.bottom_rows_wanted(fid))
+                             content_rows=frame_slots.bottom_rows_wanted(fid, cols=cols))
 
 
 def _drawable_slots(cols: int, rows: int, configured: list[str] | None = None) -> list[str]:
@@ -1501,7 +1509,7 @@ def _launch_in_operator_tmux(socket: str, session: str, *, fid: str, argv: list[
     # the whole environment here (#446): see `_guest_harness_env`.
     panes = _draw_panels(socket, slots=slots, fid=fid, harness_pane=harness_pane,
                          env=None, v=v, pane_env=_pane_identity_env(env, v),
-                         sizes=_launch_sizes(fid, slots, rows))
+                         sizes=_launch_sizes(fid, slots, cols, rows))
     _arm_panel_respawn(socket, fid=fid, panes=panes, env=None)
     tmuxctl.run("focusing the harness pane",
                 tmuxctl.server_argv(socket, "select-pane", "-t", harness_pane))
@@ -2221,7 +2229,7 @@ def cmd_launch(args) -> int:
             panes = _draw_panels(
                 SOCKET, slots=slots, fid=fid, harness_pane=harness_pane, env=env, v=v,
                 pane_env=_pane_identity_env(env, v),
-                sizes=_launch_sizes(fid, slots, rows))
+                sizes=_launch_sizes(fid, slots, cols, rows))
             _arm_panel_respawn(SOCKET, fid=fid, panes=panes, env=env)
 
             # `split-window` makes the newly created pane the ACTIVE one by default, so
@@ -2407,7 +2415,7 @@ def _relayout_pane_env(fid: str, v: tuple[int, int]) -> dict[str, str] | None:
 
 def _relayout(socket: str, *, fid: str, harness_pane: str, panels: dict[str, str],
               want: list[str], v: tuple[int, int],
-              window_rows: int) -> dict[str, str]:
+              window_cols: int, window_rows: int) -> dict[str, str]:
     """Make the running frame's panes match *want*, and return the map that resulted.
 
     Kill what is no longer wanted, split what is newly wanted, re-arm the hooks, re-assert
@@ -2426,8 +2434,11 @@ def _relayout(socket: str, *, fid: str, harness_pane: str, panels: dict[str, str
       engine the `window-resized` hook exists to correct after a window resize. The panes
       that merely SURVIVED a density change are the ones that get stretched by it, so
       they are exactly the ones a `-l` on the new pane cannot fix. :func:`_reassert_sizes`
-      does it, and *window_rows* is why it takes a measurement rather than a constant:
-      `bottom`'s height depends on the window it is in (#488).
+      does it, and *window_cols*/*window_rows* are why it takes a measurement rather
+      than a constant: `bottom`'s height depends on the window it is in — on its rows
+      (#488) and, since #500, on its columns and this frame's density too, because a
+      panel narrower than `statusline._LEFT_W` draws no table and `terse` draws at most
+      `slots._TERSE_ROWS` of one.
 
     Slots whose `split-window` fails are simply absent from the returned map, as at launch:
     a decorative panel that could not be drawn must not take down a frame that is running.
@@ -2475,14 +2486,16 @@ def _relayout(socket: str, *, fid: str, harness_pane: str, panels: dict[str, str
         keep.update(_split_panels(
             socket, slots=missing, fid=fid, harness_pane=harness_pane, env=None,
             pane_env=pane_env,
-            sizes=layout.slot_sizes(want, window_rows=window_rows,
-                                    content_rows=frame_slots.bottom_rows_wanted(fid))))
+            sizes=layout.slot_sizes(
+                want, window_rows=window_rows,
+                content_rows=frame_slots.bottom_rows_wanted(fid, cols=window_cols))))
         _arm_panel_respawn(socket, fid=fid,
                            panes={s: keep[s] for s in missing if s in keep}, env=None)
 
     _install_resize_hook(socket, harness_pane=harness_pane, panes=keep, v=v,
                          env=None, fid=fid, replacing=True)
-    _reassert_sizes(socket, fid=fid, panes=keep, window_rows=window_rows)
+    _reassert_sizes(socket, fid=fid, panes=keep,
+                    window_cols=window_cols, window_rows=window_rows)
     # `split-window` makes each new pane the ACTIVE one, so without this the operator is
     # left typing into a panel. The same correction `cmd_launch` makes after its own
     # splits, for the same reason.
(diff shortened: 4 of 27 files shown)


diazoxide:
## Round 2 — the three blocking findings are closed

Rebased onto current `main` (through #499), then one commit: `7e4f395`.

**Finding 1 (A) — `bottom` sized for a table it refuses to draw between 51 and 94
columns.** Reproduced (6 repos at 80 columns: a 7-row pane holding 1 line; 26x80 with 14
repos: 11 rows drawn into as 1, harness pinned at `HARNESS_MIN_ROWS`). Fixed.

**Finding 2 (A) — `minimal` costs the harness the same rows as `normal` and draws fewer.**
Reproduced (10 repos, `minimal`, 200 columns: an 11-row pane with 5 lines in it). Fixed.

Both by the same seam rather than two patches. `slots._table_cap(fid, width)` now holds
every reason the renderer draws fewer rows than there is content — too narrow, `terse`,
`_MAX_REPO_LINES` — and **both sides call it**: `bottom_rows_wanted` with the window's
width to size the pane, `_bottom` with the pane's own measured width to bound what it
draws. `cols` is required and keyword-only, threaded through `_launch_sizes`, `_relayout`
(its `split-window -l` included) and `_reassert_sizes`; `cmd_resize` stopped throwing its
measurement away as `_cols`.

**Finding 3 (C) — the four claims that said this could not happen** are narrowed in the
same commit: `_table_lines`' "reachable but not ordinary" comment, the identical sentence
in the news entry, `bottom_rows_wanted`'s "if those two disagreed", and `FRAME_DENSITY`'s
"the two that keep `bottom` differ in how many of its rows the table is allowed".
`docs/frame.md` carried the same two claims and gets the same correction.

### The property, and the next spelling

The property is **"the pane the launcher asks for is the pane the panel fills"**, not "the
row count matches". #488 matched the spelling. The property test now loops over every
input either side consults — repo count across the cap, width across `_LEFT_W` and down to
the smallest `visible_slots` still draws `bottom` at, and every level in `FRAME_DENSITY` —
renders into a pane of exactly the height the sizer asked for, and counts the lines. The
next spelling is any fourth input the renderer learns to read: another density level, a
`bottom`-specific `min-cols`, an alert that wraps. Its docstring says so, so adding one
without adding it to the loop leaves that dimension unpinned rather than silently wrong.

### Mutation testing

Seven guards, each applied → RED → restored → GREEN, `__pycache__` cleared between:

| mutation | result |
|---|---|
| `_table_cap` drops the `width < _LEFT_W` guard | RED, 58 failures |
| `_table_cap` drops the `terse` cap | RED, 5 |
| `_bottom` stops asking `_table_cap` | RED, 3 |
| `_launch_sizes` stops passing the window width | RED, 3 |
| `_reassert_sizes` stops passing it | RED, 3 |
| `cmd_resize` discards it again (the shipped defect) | RED, 3 |
| `_relayout`'s `split-window -l` stops passing it | **SURVIVED** → test added → RED, 1 |

The last one is why the mutation pass earned its keep: `_relayout` calls `_reassert_sizes`
immediately afterwards, which would correct the split — but that correction is
`report=False` best-effort against a pane that may already be gone, so the `-l` has to be
right in the first place. `tests/test_frame_density.py::LiveOverride::
test_a_relayout_splits_bottom_for_one_row_on_a_narrow_window` pins it.

### Idle cost, verified rather than assumed

The coordinator asked for this specifically. `IdleCostsOneStat` bounds `panel._running`
(the gate) and the existing table budget bounds `_table_lines` in isolation with a
hand-built dict — neither answers "does a REPAINT scale with the pane's height", which is
the question #488 opened by making `bottom` both the tall slot and the only animated one.

New differential test over the whole `slots.render("bottom", …)` a panel actually calls:
**45 filesystem calls for a 2-line repaint, the same 45 in the same order for a 15-line
one.** Counted at `os.stat`, `os.lstat`, `os.scandir`, `os.listdir`, `io.open`,
`builtins.open`, `subprocess.run` and `subprocess.Popen` — a directory walk does not spell
itself `stat`, and `Path.read_text` is `io.open` rather than `builtins.open`, which the
sibling budget would have missed. Differential rather than absolute so it does not need
revising every time the attention row learns a field, and it asserts the tall render
actually drew more rows first, so it cannot pass vacuously.

### Judged as a user

Measured end to end, 14 repos, `[frame] min-cols` 100 / `min-rows` 20:

| window | slots drawn | `bottom` | lines it draws | harness |
|---|---|---|---|---|
| 200x50 | top, bottom, right | 15 | 15 | 32 |
| 100x30 | top, bottom, right | 15 | 15 | 12 |
| 80x26 | top, bottom | **1** | 1 | **22** (was 12) |
| 70x14 | bottom | 1 | 1 | 12 |
| 200x15 | bottom | 2 | 2 | 12 |
| 49x40 | — | — | — | 40 |

The alert line survives everywhere `bottom` is drawn — it is the floor, and the first line
of the render. The full table renders at 95 columns and up, at `statusline.py`'s own column
widths. `left` is decided, not half-done: retired from `FRAME_SLOTS`, `SLOT_SIZE`,
`slots.SLOTS` and the shipped `slots` list, with a committed `slots = [… "left" …]` dropped
the way any unknown slot is (pinned by `Unimplemented::
test_the_retired_sidebar_is_gone_from_the_registry_too`).

### Not weaker than `main`

Still true, and the round-2 changes only tighten: `bottom_rows_wanted` and
`_reassert_sizes` both gained a REQUIRED keyword argument (a missing width is a
`TypeError` at the call site, not a default of "wide" — pinned); `_table_cap` returns
strictly fewer rows than the old count; nothing new reaches a tmux argv (`window_cols`
comes from `_window_size`, which already `isdigit`-validates and falls back to
`_FALLBACK_SIZE`, and it never reaches action text — only arithmetic); no new file is
read or written.

### Filed, not fixed here

**#506** — `statusline.render` composes its repo rows at `_LEFT_W` and lets `_columns`
crop them, so at 70 columns the CI cell is gone entirely and a failing repo reads as merely
dirty. Same class of defect on the surface you see when you are *not* in a frame, with the
measurement in the issue. Referenced from the news entry.

Full suite: `Ran 5251 tests`, `OK`.

diazoxide:
## Round 3 disposition — all three blocking items closed in `58cdeaa`

Rebased onto `origin/main` (`eab4fad`, past #492/#495/#497/#499). Full suite 5312 tests OK locally; CI green on 3.11/3.12/3.13/3.14. Not weaker than `main` — see the bottom of this comment.

---

### 1 — (A) `bottom` still sized for a table it refuses to draw when `right` precedes it. **Fixed.**

Reproduced first, in an isolated plane (`config.ROOT` printed as a scratchpad dir, never the worktree): `_launch_sizes("f-1", ["right","top","bottom"], 110, 40)` answered `bottom: 7` where the pane is 87 wide. Then measured against real tmux 3.7c on a private socket, splitting exactly as `panel_argvs` does:

```
["top","bottom","right"]  ->  bottom 110x7   harness 87x30
["right","top","bottom"]  ->  bottom  87x7   harness 87x30
["top","right","bottom"]  ->  bottom  87x7   harness 87x30
["bottom","right"]        ->  bottom 110x7   harness 87x32
```

and, for the re-layout case, `kill-pane` on `right` widened that same `bottom` from 87 back to 110, while re-splitting `right` off the harness afterwards left it at 110.

**The property, named once:** *the width that decides how tall `bottom` is must be the width of `bottom`'s own **pane**.* `layout.bottom_cols(slots, window_cols=…)` is that arithmetic — walk the split order, charge every slot that takes columns until `bottom`, stop there.

All three sites pass its answer, and each passes the order **its** panes are actually in:

* `_launch_sizes` — the drawable slot list;
* `_relayout` — `list(keep) + missing`, **not** `want`. `want` at `full` is `["top","bottom","right"]`, which is a *fresh launch's* geometry; a frame launched `right`-first keeps its inset 87-column `bottom` because nothing was killed or re-split, and sizing from `want` would answer 110 and hand back the seven-row pane;
* `_reassert_sizes` — `list(panes)`, whose insertion order is the split order (`_split_panels` writes it that way, `state.panes` is JSON and round-trips it).

**The spelling that produced the bug is gone.** `bottom_rows_wanted` takes `pane_cols`, not `cols`; the old call is a `TypeError` at the call site rather than six blank rows nobody re-measures. `_launch_sizes` takes `window_cols`/`window_rows` keyword-only for the same reason. `panel_argvs` and `bottom_cols` now read one `_COLUMN_SLOTS` rather than two lists of names, so the next side slot cannot be `-h` in one place and free in the other. *The next spelling to refuse is `width`* — `_table_cap`'s own parameter name, correct there because the renderer measures a pane, and a caller who copies it up would be right by luck; the signature test says so.

**Judged as a user, on a real frame with real clones** — real tmux, real `charter panel` processes, isolated plane, 6 and 14 dirty clones. Full sweep, `right` first:

```
window 200x40 -> bottom 177x15  drew 15, 0 blank, harness 22
window 130x40 -> bottom 107x15  drew 15, 0 blank, harness 22
window 118x40 -> bottom  95x15  drew 15, 0 blank, harness 22
window 117x40 -> bottom  94x1   drew  1, 0 blank, harness 36
window 110x40 -> bottom  87x1   drew  1, 0 blank, harness 36
window  80x40 -> bottom  57x1   drew  1, 0 blank, harness 36
window 200x40 -> bottom 177x15  drew 15, 0 blank, harness 22
```

Zero blank rows at every step of a drag across the boundary and back, and the harness gets its 14 rows back the moment the table cannot be drawn — the reported case was `harness 22 where it should have 36`, which now reads 36. The shipped order is clean at the same sizes (boundary between 96 and 80). The **alert line survives everywhere**: every one-row pane above drew `0 todos · F2 menu`. **Small terminals degrade sanely**: 60x24 and 50x14 give the one-row strip; 40x10 is below half the floors and the harness takes the whole terminal, as documented. **`left` is decided, not half-done**: absent from `FRAME_SLOTS`, `SLOT_SIZE`, `_DROP_ORDER` and the renderer registry, dropped like a typo with a test on the exact four-slot list that shipped a release ago, and named as retired in `docs/frame.md`, the news entry and `charter.toml`.

**Pinned against tmux, not against my measurement.** `BottomsWidthIsWhatTmuxActuallyGivesIt` drives `panel_argvs`' own commands (only the program after `--` is swapped for a `sleep`) in four orders and asserts `#{pane_width}` equals `bottom_cols`' answer — plus an `assertNotEqual` so "always the window's width" cannot pass the loop. Unit tests in `test_frame_layout.py`; regression tests at the launcher (`test_a_sidebar_split_first_makes_the_pane_narrower_than_the_window`, with the shipped-order control beside it), at the resize hook, and at the re-layout.

### 2 — (C) Three docstrings asserting the invariant this breaks. **Narrowed, same commit.**

`slots._table_cap`, `commands_frame._launch_sizes` and `slots._table_lines` each said "`bottom` is split BEFORE `right`, so the window's width IS the pane's". `FRAME_FIELDS` supplies the *default* order; the operator's own order is the geometry, and the config test says so in as many words. All three rewritten to say the pane's width, and to name `layout.bottom_cols` as where the launcher's side of it comes from. The two operator-facing copies (`docs/frame.md`, the news entry) are narrowed the same way — "with the same width" is now "at the same density and at the width of the **pane**", with the 118-column threshold a `right`-first frame actually has spelled out in both, and in `charter.toml`'s own comment (which had claimed the inset table "starts losing its right-hand columns" — it refuses to draw instead).

### 3 — (C) Two overstatements about what `minimal` costs. **Narrowed, same commit.**

Re-measured independently, isolated plane, `bottom_rows_wanted(fid, pane_cols=200)`:

```
repos:  0  1  2  3  4  5  8  10  14  20
normal: 1  2  3  4  5  6  9  11  15  15
minimal:1  2  3  4  5  5  5   5   5   5
```

Confirms the report exactly. The docs said the `minimal` pane "is four rows tall" (it is at most **five** — the attention row plus `_TERSE_ROWS`) and "four rows shorter" (true at exactly one repo count; 0 rows below five clones, 10 at fourteen). Both now say what the code does, with the range named.

---

### The idle-cost property — verified, not assumed

A taller pane redrawing more is exactly where per-row filesystem work would appear, so this was measured independently rather than taken from its test. Isolated plane, counted at every spelling a walk arrives by (`os.stat`/`lstat`/`scandir`/`listdir`, `io.open`, `builtins.open`, `subprocess.run`/`Popen`):

* **idle tick: 1 filesystem call** at a 1-row pane and at a 15-row pane alike;
* **repaint: 13 calls at 2 lines, the same 13 at 15 lines.**

The one call is the frame's version file, and it is a `read_text` — so an `os.stat` budget would have missed it entirely, which is why the new test counts `io.open` too. Pinned by `IdleCostsTheSameWhateverTheBottomPaneIsTallEnoughToDraw`, differential (an absolute number is not a budget) with a `len(short) == 1` floor so "repaint on every tick" is red. The news entry's "exactly one `stat`" is corrected to "one filesystem touch — a read of the frame's own version file", with the measured repaint numbers.

### Mutation testing

Eleven mutations, each applied → RED, restored → GREEN, `__pycache__` cleared both ways:

| | mutation | red |
|---|---|---|
| M1 | `bottom_cols` answers the window's width (the pre-fix behaviour) | 12 |
| M2 | `_BORDER_COLS = 0` | 4 |
| M3 | the split order ignored (no `break` at `bottom`) | 7 |
| M4 | `_COLUMN_SLOTS` emptied | 11 |
| M5 | `max(0, …)` dropped | 1 |
| M6 | the launcher passes the window's width again | 2 |
| M7 | `_reassert_sizes` passes the window's width again | 1 |
| M8 | `_relayout` sizes from `want`'s order | 1 |
| M9 | `pane_cols` re-admits a `cols` alias | 2 |
| M10 | the table `stat`s once per row | 2 |
| M11 | the idle tick repaints unconditionally | 2 |

M1/M2/M3 are each caught by the real-tmux test on its own — tmux refuting the arithmetic, not two literals agreeing. M2 is caught by **only** the new tests, which is the point of adding them. M10 and M11 are what stop the two idle-cost tests from being vacuous.

### Filed, not widened

**#510** — the running frame *derives* `bottom`'s width from the recorded slot order where tmux could be asked for the pane's own (`display-message -p -t <pane> '#{pane_width}'`). Correct today; the record and the geometry can part if the `panes` file is corrupted or hand-edited, or if a future re-layout moves a surviving pane. One extra `display-message` on the resize hot path is a real trade and belongs to its own change. Referenced from the news entry.

### Not weaker than `main`

Everything this round adds is a narrowing. `bottom_cols` returns `min(window_cols, …)` at every input — never wider — and clamps at 0, so no site can be handed a larger width than before and none can be handed a negative one. `bottom_rows_wanted` and `_launch_sizes` each gained required keyword arguments, so an under-specified call is a `TypeError` at the call site rather than a silent default. Nothing new reaches a tmux argv: `bottom_cols` consumes `window_cols` (already `isdigit`-validated in `_window_size`, falling back to `_FALLBACK_SIZE`) and a slot list already filtered through `instance.FRAME_SLOTS`, and its output is arithmetic that only ever reaches a `<` against `_LEFT_W`. The only thing that still reaches `resize-pane`/`split-window` is `str(int)` out of `layout.slot_sizes`. Every pre-existing guard is untouched — `_PANE_ID_RE` on every pane id, the `slot not in _RESIZE_FLAG: continue` filter, `_relayout`'s shape check above the `want` branch, `_action_word_is_safe`/`_FRAME_ID_RE` on the resize hook. No file is newly read or written and no parsing is widened. `panel_argvs`' direction now comes from `_COLUMN_SLOTS`, so a slot not in it splits `-v` where the old expression split `-h` — reachable only for a name that is neither `top`/`bottom`/`right`, which `instance.FRAME_SLOTS` already rejects before `panel_argvs` is ever called, and a `-v` costs the harness rows rather than columns either way. `bottom`'s floor is still `SLOT_SIZE["bottom"]` = 1, so tmux is never asked for a 0-row split.

Issue #532, opened 2026-08-25
TestShippedSkillsStayInSync reads skills/ off the filesystem, so a locally drafted skill turns the suite red


diazoxide:
Found while fixing #529. Same defect shape one level down, in a different file, so it is filed
rather than folded into that PR.

## The site

`tests/test_doctor_shadowed.TestShippedSkillsStayInSync.test_the_constant_matches_the_shipped_directory`:

```python
on_disk = {d.name for d in (REPO_ROOT / "skills").iterdir()
           if d.is_dir() and (d / "SKILL.md").is_file()}
self.assertEqual(set(doctor.SHIPPED_SKILLS), on_disk)
```

`assertEqual` in both directions against a **filesystem** enumeration. `doctor.SHIPPED_SKILLS`
is a constant about what charter *ships*; `skills/*/SKILL.md` on disk is whatever the person
running the suite happens to have in their checkout.

## Why it is the #529 shape

Anyone drafting a skill does it where skills live. `skills/my-draft/SKILL.md`, untracked and
uncommitted, turns the suite red with a message about a constant being out of sync — and the
correct response is neither to add the draft to `SHIPPED_SKILLS` nor to delete the draft. The
same is true of a `git stash` that leaves a directory behind, or a half-finished branch's
leftovers, or a `skills/` copy someone made to diff against.

It is milder than #529 — it needs a directory containing a `SKILL.md`, not just any local
artefact — but it is the same mistake: the question is about the repository and the answer is
read off a machine.

## The fix, and the part worth being careful about

`git ls-files skills/*/SKILL.md` (or the top-level split that #529's fix uses) answers the
actual question, and keeps the property the test is for: adding a skill without updating
`SHIPPED_SKILLS` still fails on the PR that commits the skill, which is when the decision
needs making.

What must NOT change is the direction the assertion runs. It is an equality on purpose:
`SHIPPED_SKILLS` naming a skill that no longer exists is as much a defect as a skill missing
from it, and `doctor`'s shadow check reports on both. Swapping it for a subset assertion to
make the local-draft symptom go away would give up the removal half — the exact trade #529's
history warns about, where the first version of the classification test was a subset assertion
that was true by construction.

Issue #531, opened 2026-08-25
26 test modules define classes after their `if __name__ == "__main__"` trailer, so running one module directly silently skips them


diazoxide:
Found while fixing #529 (a test enumerating the developer's filesystem). This is the same
family — a test that cannot fail — but the mechanism is different and it is in 26 modules, so
it is filed rather than folded into that PR.

## What happens

`tests/test_doctor_shadowed.py` puts its `if __name__ == "__main__": unittest.main()` block at
line 104, and then defines three more test classes after it. The `if` block executes at module
exec time, *before* those classes exist, so running the module the way a developer runs one
module runs a strict subset:

```
$ python3 tests/test_doctor_shadowed.py
Ran 8 tests in 0.007s
OK

$ python3 -m unittest tests.test_doctor_shadowed
Ran 14 tests in 1.203s
OK
```

Six tests vanish, and the run says `OK`. The six that vanish are not the cheap ones —
`TestCharterItselfIsExemptHoweverItWasInstalled`, `TestTheCheckItselfIsWired` and
`TestNoCheckHidesAPythonErrorInAWarning` are the `PersonaIso` classes that exercise the real
seam; the eight that survive are the pure-function ones. The subset that runs is biased toward
the tests least likely to catch anything.

## How widespread

26 modules, mechanically:

```
test_asks_do_not_hang_bypass_permissions.py  main@156, 2 classes after
test_browser_install.py                      main@82,  3 classes after
test_cli_smoke.py                            main@126, 1 class after
test_docs_show.py                            main@148, 2 classes after
test_doctor_personas.py                      main@128, 1 class after
test_doctor_shadowed.py                      main@104, 3 classes after
test_guard_denials_record_their_shape.py     main@117, 1 class after
test_hooks.py                                main@444, 2 classes after
test_hooks_need_no_async.py                  main@80,  1 class after
test_inflight_dispatch.py                    main@305, 3 classes after
test_init.py                                 main@183, 1 class after
test_opencode_shim.py                        main@56,  2 classes after
test_persona_agent_forge_wording.py          main@87,  2 classes after
test_persona_memory_sync.py                  main@69,  1 class after
test_persona_skill_lint.py                   main@55,  1 class after
test_plane_repo.py                           main@219, 1 class after
test_plane_root_branch_guard.py              main@163, 1 class after
test_plugin.py                               main@284, 1 class after
test_routing_roster_block.py                 main@177, 2 classes after
[...]

Issue #519, opened 2026-08-25
The suite fails 16 tests inside a charter frame: nothing isolates a test from the operator's ambient session environment


diazoxide:
Found while verifying #513 — adjacent to that PR, not part of it.

16 tests fail on a machine where the operator's own shell exports a charter/Claude Code
session id or is inside tmux. That is not an exotic environment: it is **every charter
frame session**, which is exactly where the operator runs the suite.

## Reproduce

```
$ python3 -m unittest discover -s tests 2>&1 | grep -E "^(OK|FAILED|Ran )"
Ran 5420 tests in 198.640s
FAILED (failures=16)
```

Same tree, same commit, with four variables unset:

```
$ unset CHARTER_SESSION_ID CLAUDE_CODE_SESSION_ID TMUX TMUX_PANE
$ python3 -m unittest discover -s tests 2>&1 | grep -E "^(OK|FAILED|Ran )"
Ran 5420 tests in 205.429s
OK
```

Reproduced on `origin/main` content with the #513 branch stashed, so it predates that
work and is unrelated to it.

## The failures

```
FAIL: test_session_identity.AbsenceIsRepresentable.test_current_reads_the_environment
FAIL: test_session_identity.AbsenceIsRepresentable.test_both_readers_now_agree
FAIL: test_trace.TestTrace.test_record_and_read_in_order
FAIL: test_trace.TestTrace.test_remember_emits_memory_event
FAIL: test_trace.TestTrace.test_set_active_emits_persona_use
FAIL: test_piece_claim.TestClaimingRecordsWhoTookIt.test_the_claim_carries_the_session_and_host
FAIL: test_piece_claim.TestClaimingRecordsWhoTookIt.test_a_claim_with_no_session_is_still_recorded
FAIL: test_piece_claim.TestTheListingNamesTheClaimant.test_the_claimant_appears_in_the_listing
FAIL: test_piece_history.TestReadingHistory.test_history_names_the_claimant
FAIL: test_piece_session_announce.TestCollisionIsReportedNotRefused (x3)
FAIL: test_workspace_scope_honesty.TestScopeNamesWhatWasWritten.test_the_session_pointer_is_still_written_either_way
FAIL: test_workspace_lock.TestConfirmNudge.test_no_nudge_once_locked
FAIL: test_statusline_watch.WatchIsReachableFromTheCli.test_without_the_flag_it_still_reads_a_payload_and_prints_once
[...]

Issue #510, opened 2026-08-25
frame: cmd_resize derives bottom's width from the recorded slot order where tmux can be asked for the pane's own


diazoxide:
`cmd_resize` derives `bottom`'s width from the recorded slot ORDER; tmux can be asked for the pane's actual width.

## What it is

#500 (round 3) makes every site that sizes `bottom` ask `layout.bottom_cols(order, window_cols=…)` instead of handing the window's width straight through — because a `[frame] slots` naming `right` before `bottom` insets the pane by 23 columns, and a pane below `statusline._LEFT_W` (95) draws no table at all. Measured on tmux 3.7c: a 110-column window with `slots = ["right", "top", "bottom"]` gives an 87-column `bottom`.

The *order* those panes were split in is what `bottom_cols` reads. At launch that order is the drawable slot list, which is unambiguous. On the resize path it is `list(state.panes(fid))` — the insertion order of a JSON dict that `commands_frame._split_panels` wrote in split order and `state.panes` round-trips.

That is correct today, and it is a *derivation* where a *measurement* is available: `bottom`'s pane id is already in that map, and one

```
display-message -p -t <bottom pane> '#{pane_width}'
```

answers the question tmux itself is the authority on. `cmd_resize` already runs exactly one `display-message` (for the window size), so this is a second one on the same path.

## Why it is worth closing rather than leaving

The derivation is right only while the recorded order and the real geometry agree. Two ways they can part:

* the `panes` file is JSON on disk in the frame's own state directory — a truncated write, a hand edit, or a charter that wrote a different shape all reach `state.panes`, which validates the VALUES (`isinstance(v, str)`) and says nothing about the order;
* a future re-layout that moves a surviving pane rather than killing and re-splitting it would change the geometry without changing the map's order.

In both cases the failure is silent and is exactly the pre-fix symptom: an over-tall `bottom` re-asserted on every step of a terminal drag, with the rows taken off the harness (`layout.HARNESS_MIN_ROWS` keeps the agent session at 12 while the table's blank rows sit below it). Measured on tmux 3.7c with 14 real clones at 110x40 in the `right`-first order, before the fix: the pane got 15 rows, drew 1, and the harness kept 22 where it should have had 36.

## Cost

[...]

PR #533, opened 2026-08-25
The plugin classification asks the index, not the developer's filesystem (#529)

### LINKED ISSUE #529: test_plugin_freshness enumerates the working directory, so a real plane's .charter/ and .superpowers/ fail it
Found while enabling `[update] channel = "dev"` on a real plane. Unrelated to the channel — reproduced with the setting reverted.

## What happens

```
$ python3 -m unittest tests.test_plugin_freshness.TheHashCoversWhatThePluginLoads
AssertionError: Items in the first set but not the second:
'.superpowers'
'.charter' : new top-level directories ['.charter', '.superpowers']:
             does a Claude Code plugin LOAD it? Add it to plugincache.PLUGIN_SURFACE,
             or to _NOT_PLUGIN_SURFACE in this file.
```

`.charter/` is the plane's own **gitignored** state directory. `.superpowers/` is the **gitignored** SDD workspace directory. Neither is repo content, neither is shipped, and neither can be "classified" in a list of what a plugin loads — they are local artefacts of using charter, present on any working plane and absent everywhere else.

## Why it passes everywhere it was tested

CI checks out a fresh tree: no `.charter/`, no `.superpowers/`. A `git archive` extraction: same. A linked worktree: same. **The only place it fails is a real plane that has actually been used** — which is the operator's machine, and nobody else's.

That is the same shape as #519/#521/#528 (tests reading the developer's environment) and as #402/#492 (tests reading the developer's plane), one layer over: this one reads the developer's **filesystem**.

## The defect

The test enumerates the **working directory**, when the question it is asking is about the **repository**. A directory that is gitignored is by definition not part of what the plugin ships, so it cannot be an answer to "does a Claude Code plugin load it".

Enumerate tracked content instead — `git ls-files` already appears elsewhere in the suite for exactly this reason (`tests/test_workflows.py` reads the index). Then a new top-level directory that is actually *committed* still fails the test on the PR that adds it, which is the property the test was written for (see #510's `M34`, where this test was strengthened precisely so an addition could not slip through).

Alternatively, skip dot-directories — but that is weaker: it would also stop catching a committed `.github`-style directory that genuinely does ship.

## Note for the fixer

Do not fix this by adding `.charter` and `.superpowers` to `_NOT_PLUGIN_SURFACE`. That makes the symptom go away on this machine and leaves the test still reading the filesystem — the next developer with a different local artefact (`.venv`, `.idea`, `.pytest_cache`, a scratch directory) hits the same failure, and the list grows one machine at a time.

diazoxide:
Closes #529

## The defect

`TheHashCoversWhatThePluginLoads.test_every_top_level_directory_is_classified_one_way_or_the_other`
enumerated `REPO.iterdir()`. The question it asks is about the **repository** — does a Claude
Code plugin load this directory — and a gitignored directory cannot be an answer to it, because
it is not shipped at all.

It passed in CI, in a `git archive` extraction, and in a linked worktree — every tree nobody has
worked in. The only place it failed was a plane charter is actually used from, where `.charter/`
(state) and `.superpowers/` (SDD workspaces) exist and are gitignored. Third layer of the same
mistake: the developer's environment (#519/#521/#528), their plane (#402/#492), now their
filesystem.

## The fix

`git ls-files` — already the seam `tests/test_workflows.tracked` uses, for the same stated
reason: a denylist of local artefacts is a guess at where untracked content lives, and the index
is the answer.

`.charter` and `.superpowers` are **not** added to `_NOT_PLUGIN_SURFACE`. Two entries that were
already there and are not repo content — `.git` and `dist` — are removed with the filesystem
walk, and a second test now holds every remaining name to being tracked. The cheap fix now fails
by name rather than going green on one machine while the list grows one developer at a time.

`PLUGIN_SURFACE` is deliberately exempt from that guard, and the docstring says so: it names what
a plugin loads *if present*, and `commands/` and `agents/` are two Claude Code surfaces charter
does not use yet.

## Verified the way the defect appears

A tree carrying `.charter/`, `.superpowers/`, `.venv/` and `.pytest_cache/` — the pre-fix
enumeration and the post-fix one, side by side against that same tree:

```
$ find . -maxdepth 1 -type d -not -name . | sort
./.charter
./.claude
./.claude-plugin
./.github
./.pytest_cache
./.superpowers
./.venv
./charter
./docs
./hooks
./personas
./skills
./tests
./workspaces

OLD (REPO.iterdir) unclassified: ['.charter', '.pytest_cache', '.superpowers', '.venv']
NEW (git ls-files)  unclassified: []
```

Green in that tree:

```
$ python3 -m unittest tests.test_plugin_freshness.TheHashCoversWhatThePluginLoads.test_every_top_level_directory_is_classified_one_way_or_the_other
Ran 1 test in 0.308s
OK
```

And separately, a **genuinely tracked** new top-level directory still turns it red — the property
#510's `M34` work strengthened this test for:

```
$ mkdir mcp && echo '{"servers": {}}' > mcp/servers.json && git add mcp/servers.json
$ python3 -m unittest tests.test_plugin_freshness.TheHashCoversWhatThePluginLoads.test_every_top_level_directory_is_classified_one_way_or_the_other
AssertionError: Items in the first set but not the second:
'mcp' : new top-level directory ['mcp']: does a Claude Code plugin LOAD it? Add it to
        plugincache.PLUGIN_SURFACE, or to _NOT_PLUGIN_SURFACE in this file.
FAILED (failures=1)
```

Nothing here renders, so there is no frame output to paste; the only user-facing surface is the
assertion message above.

## Mutation tests

Each applied, run RED, restored, run GREEN, with `__pycache__` cleared between.

**M1 — park the untracked names in the list** (the workaround #529 forbids):
`_NOT_PLUGIN_SURFACE` += `".charter", ".superpowers"` →

```
FAIL: test_nothing_untracked_can_be_parked_in_the_not_plugin_surface_list
AssertionError: Items in the first set but not the second:
'.charter'
'.superpowers' : _NOT_PLUGIN_SURFACE names ['.charter', '.superpowers'], which this
                 repository does not track. This list classifies directories charter SHIPS;
                 a local artefact belongs in .gitignore, not here (see #529).
```

**M2 — the helper reads the filesystem again** (the pre-fix behaviour, restored inside
`tracked_top_level_directories`) → **two** failures, the new property test and the
classification itself:

```
FAIL: test_the_classification_reads_the_index_and_not_the_working_directory
AssertionError: Items in the first set but not the second:
'.superpowers'
'.charter'
'.venv'
'.git' : an untracked directory reached the classification
FAILED (failures=2)
```

**M3 — drop the directory filter** (`if p` instead of `if p and "/" in p`, so tracked top-level
*files* count as directories) →

```
FAIL: test_every_top_level_directory_is_classified_one_way_or_the_other
FAILED (failures=1)
```

Restored: `Ran 53 tests in 2.251s / OK`.

One mutation is honestly **not** caught: flipping `check=True` to `check=False` on the
`git ls-files` call changes nothing while git succeeds, so no test distinguishes it. It is a
robustness choice rather than a property, and the docstring says why it is there — a helper that
returns an empty set when it cannot read the index classifies vacuously, which is a guard that
reports green forever.

## Suite

Ambient variables cleared, outside a live frame, in a tree that had `.charter/`,
`.superpowers/`, `.venv/` and `.pytest_cache/` present:

```
$ env -u CHARTER_SESSION_ID -u CLAUDE_CODE_SESSION_ID -u TMUX -u TMUX_PANE \
    python3 -m unittest discover -s tests
Ran 5528 tests in 227.864s
OK
```

`config.ROOT` resolved to `/Users/aharon/IdeaProjects/charter` — the worktree's own
`charter.toml` redirects to the main tree via `root._plane_of`, as the brief warns. It does not
affect this change: the module's `REPO` comes from `Path(__file__).resolve().parents[1]`, so the
index read is the worktree's own.

## Filed, not widened

Two adjacent surfaces found while here, filed rather than folded in:

* **#532** — `TestShippedSkillsStayInSync` reads `skills/` off the filesystem, so a locally
  drafted `skills/<name>/SKILL.md` turns the suite red. Same shape one level down.
* **#531** — 26 test modules define classes *after* their `if __name__ == "__main__"` trailer, so
  `python3 tests/test_doctor_shadowed.py` runs 8 tests and says `OK` while
  `python3 -m unittest tests.test_doctor_shadowed` runs 14. The six that disappear are the
  `PersonaIso` ones that exercise the real seam.

### DIFF
--- a/docs/news/unreleased-the-plugin-classification-reads-the-index.md
+++ b/docs/news/unreleased-the-plugin-classification-reads-the-index.md
@@ -1,0 +1,50 @@
+---
+version: unreleased
+headline: charter's suite no longer fails because you have actually used charter — the plugin classification reads the index, not your filesystem
+---
+
+`tests.test_plugin_freshness.TheHashCoversWhatThePluginLoads` holds a classification: every
+top-level directory in this repository is either something a Claude Code plugin **loads**
+(`plugincache.PLUGIN_SURFACE`) or something it does not (`_NOT_PLUGIN_SURFACE`), and adding a
+third kind fails the test until a person decides which it is. That decision is the only thing
+the check exists to force, and it is a good one to force: the surface constant is a whitelist,
+so a plugin directory missing from it is a whole category of staleness `doctor` would report
+as clean.
+
+It enumerated `REPO.iterdir()` — the working directory. So on a plane that has actually been
+used it failed, naming the plane's own gitignored `.charter/` (state) and `.superpowers/` (SDD
+workspaces) as unclassified directories:
+
+```
+AssertionError: Items in the first set but not the second:
+'.superpowers'
+'.charter' : new top-level directories ['.charter', '.superpowers']:
+             does a Claude Code plugin LOAD it? Add it to plugincache.PLUGIN_SURFACE,
+             or to _NOT_PLUGIN_SURFACE in this file.
+```
+
+Neither is repo content, neither ships, and neither is an answer to "does a plugin load it".
+It passed in CI, in a `git archive` extraction and in a linked worktree — every tree that has
+never been worked in. **The only place it failed was a machine charter is used from.**
+
+That is the third layer of the same mistake this suite keeps making: a test reading the
+developer's environment (#519/#521/#528), the developer's plane (#402/#492), and now the
+developer's filesystem.
+
+**Fixed by asking the index, not the disk.** `git ls-files` is already the seam
+`tests/test_workflows.py` uses to decide which files move only with a commit, for exactly this
+reason: a denylist of local artefacts is a guess at where untracked content lives, and the
+index is the answer. A directory that is gitignored cannot reach the classification; a
+directory that is genuinely committed still fails the test on the PR that commits it, which is
+the property the check was strengthened for.
+
+**The cheap fix is now refused mechanically.** Writing `.charter` and `.superpowers` into
+`_NOT_PLUGIN_SURFACE` would have gone green on one machine and left the test reading the
+filesystem — and the list would then grow by one name per developer (`.venv`, `.idea`,
+`.pytest_cache`, a scratch directory) while the check quietly stopped classifying anything. A
+second test now holds every name in that constant to being a directory this repository
+actually tracks, so a local artefact parked there fails by name. `.git` and `dist` were listed
+and are gone with it; they were never repo content either.
+
+This only ever reached you if you run charter's own test suite. Nothing to adopt: upgrading is
+the whole of it.
--- a/tests/test_plugin_freshness.py
+++ b/tests/test_plugin_freshness.py
@@ -26,6 +26,8 @@ own plugin installation.
 
 from __future__ import annotations
 
+import os
+import subprocess
 import unittest
 from pathlib import Path
 from unittest import mock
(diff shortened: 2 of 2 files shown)


diazoxide:
Amended one docstring-only addition after opening: the helper now states what it cannot see. `git ls-files` prints a top-level **submodule** as a bare path with no slash, indistinguishable from a tracked top-level file, so a gitlink at the top level would be classified as neither. charter has no submodules and `dependencies = []`; adding one would need `ls-files -s` and a mode-160000 check. Recorded rather than left implicit — a guard that overclaims is the defect twice over.

`Ran 53 tests in 2.324s / OK` after the amend.

Issue #528, opened 2026-08-25
tests: 108 patch.dict(os.environ) calls omit clear=True, so the suite's answers depend on the developer's shell


diazoxide:
Found while closing a single instance of this on PR #525. The instance is fixed; the class is not.

## The measurement

**108 `mock.patch.dict(os.environ, {...})` calls without `clear=True`, across 38 files** — out of 168 total `patch.dict(os.environ` calls in the suite.

Every one is a test that sets some environment variables and lets the rest of the developer's shell through. When one of those survivors is a charter variable the code under test reads, the test's answer depends on the shell it was run in.

## Why this is not theoretical

Already established by #519 and #521: run the suite from inside a live charter frame and it reports **16–17 failures** that do not exist. Confirmed again on #525: the developer's own shell (itself inside a frame) gives `Ran 5526 … FAILED (failures=17)`, while the same tree with `CHARTER_SESSION_ID`, `CLAUDE_CODE_SESSION_ID`, `TMUX` and `TMUX_PANE` cleared gives `Ran 5526 … OK`.

And the failure runs both directions. On #525 the leak produced a **false green**: `test_a_frame_with_no_recorded_workspace_refreshes_what_it_always_did` had both sides of its assertion collapse to the ambient pinned name, so mutation M19 died with a clean environment and **survived under the pin**. A test that quietly stops testing when a variable happens to be set is worse than one that fails.

## Why it is being filed rather than fixed

108 call sites across 38 files is not a drive-by. The areas are unrelated to each other — secrets and vault, session identity, roster, toolgate, XDG/Codex home — so a single sweeping change would be a large diff with no single reviewer who understands all of it.

What was verified before drawing that line:

- PR #525's own new tests (four classes across `test_frame_slots.py`, `test_frame_state.py`, `test_frame_tmux_integration.py`) already use `clear=True`; the two in `test_frame_liveness.py` were its only blind spot and are now closed.
- The flagged-adjacent instances in `test_frame_slots.py` and `test_frame_state.py` are nested inside a `setUp` that clears first — safe.
- `test_frame_launcher.py::Launch` holds the largest single concentration (16 instances) with no class-level clearing, but the ones inspected assert on variable *names* or on "nothing carried", not on values that ambient leakage would flip.

## What a fix should look like

[...]

Issue #487, opened 2026-08-24
test_frame_tmux_integration is module-wide flaky on CI: same commit, one run red and its sibling green


diazoxide:
Hit while cutting 0.52.0, and worth a tracking issue because it now costs a re-run on the release path.

## The observation

Commit `717ede0`'s parent content ran twice on CI at the same time — once push-triggered, once pull_request-triggered, **identical tree**:

| run | 3.11 | 3.12 | 3.13 | 3.14 |
| --- | --- | --- | --- | --- |
| [32769902396](https://github.com/diazoxide/charter/actions/runs/32769902396) (push) | pass | pass | pass | pass |
| [32769943858](https://github.com/diazoxide/charter/actions/runs/32769943858) (pull_request) | pass | cancelled | **fail** | cancelled |

The failure:

```
FAIL: test_the_harnesss_real_exit_code_survives_in_someone_elses_server
      (test_frame_tmux_integration.WindowInsideAnOperatorsTmux)
  self.assertEqual(commands_frame._pane_state(OP_SOCKET_PATH, pane_id),
                   (commands_frame._DEAD, 33))
AssertionError: Tuples differ: ('dead', 1) != ('dead', 33)
```

`gh run rerun --failed` went green on all four with no change to the tree.

The pane died with **1** rather than the harness's real **33** — i.e. the spawned command failed before it reached the exit the test is about, which is what load on a shared runner looks like rather than what a wrong exit-code path looks like. Five different tests in this module were reported failing-then-passing on re-run on the same day, so this is the module and not this case.

## Why it matters more than an ordinary flake

`release.yml` runs the same suite on all four Pythons **as a gate on the irreversible PyPI upload**. A flake there fails a release that is otherwise correct, and the recovery — re-run the failed job, or `workflow_dispatch` — is safe but is a judgement call made under time pressure, at the one moment when "just re-run it" and "do not merge over a red check" are hardest to tell apart. The release engineer's charter has to carry a standing exception for one module, which is exactly the kind of exception that later swallows a real failure.

## Not asking for the assertion to be widened

`('dead', 33)` is the property the module exists for — the harness's real exit code surviving a tmux respawn — and an assertion that also accepts 1 tests nothing. The fix wants to be in what the test waits for: an exit of 1 means the command under test never got as far as exiting 33, so there is a readiness condition being assumed rather than waited on.

Evidence trail: the sibling-run table above is the useful artifact — same bytes, opposite results, no re-run needed to see it.

Issue #518, opened 2026-08-25
frame: pick or create a workspace when the harness opens, instead of resolving one silently


diazoxide:
Requested by the operator: *"when opening charter — we need to show first some workspace selector, or create new workspace, that will automatically create, switch and then open harness."*

Today `charter claude` resolves a workspace silently (`workspace.resolve`'s precedence chain) and goes straight into the frame. Choosing where to work means quitting, running `charter ws use <name>` or `charter workspace create <name>`, and launching again.

## The flow being asked for

1. `charter <harness>` opens with a **workspace picker** — the existing workspaces, plus a "create new" entry.
2. Picking an existing one switches to it.
3. Picking "create new" prompts for a name, **creates it, switches to it**, and continues.
4. Then the harness starts, in the frame, in the chosen workspace.

## What has to be settled

**Where the picker lives.** The frame's launch path already has a gap it could occupy: `layout.PLACEHOLDER` (`["cat"]`) holds the harness pane open before the harness replaces it, precisely because the pane must exist before `remain-on-exit` can be set. That is a real moment where the frame exists and the harness does not. The alternative is a plain terminal prompt before tmux takes over — simpler, but it means the picker looks nothing like the rest of the TUI.

**How it is skipped.** An operator who always works in one workspace must not answer a prompt every launch. Options: honour an explicit `$CHARTER_WORKSPACE` or `--workspace` and skip the picker entirely (that is already the top of `workspace.resolve`'s precedence, so it falls out naturally); a `[frame]` config key; or remembering the last choice. **Whatever is chosen, a non-interactive launch must never block** — `charter claude` is run from scripts and from other agents, and a picker that waits for input in a pipe is a hang.

**The session lock.** `workspace.set_active` locks the session to the chosen workspace and refuses a later switch (returns `"locked"`). A picker that selects, then hits its own lock, would be a confusing first experience — decide whether picking *is* the confirmation that locks, and say so on screen.

**Containment.** Workspace names come from the plane and are **committed values**. If the picker is a tmux `display-menu`, every name reaches tmux config text — the exact class that made `[frame] hotkey` a code-execution bug and `mcp.json` keys a YAML injection (#453). `menu._ACTION_ID_RE` and `contain.one_line` exist for this and must be applied before a name reaches a menu item. A name typed by the operator for "create new" is different — that is their own input, but it still has to pass `workspace` name validation before anything is created.

[...]

PR #541, opened 2026-08-25
A green tmux run means something: the suite waits for the thing itself, and stops reading a void death as an exit code (#494, #507, #531)

### LINKED ISSUE #494: MenuClientIntegration assumes the menu is open after a fixed 0.8s sleep, and under load it is not
`MenuClientIntegration` writes a hotkey to a pty and then sleeps 0.8 s standing in for "the
menu is now open on that client". Under cpu contention it is not, the next keystroke lands
before the menu exists, and the test fails on an assertion about charter.

This is the one case #487's fix did not close. Everything else in
`tests/test_frame_tmux_integration.py` that waited by sleeping now polls for the condition it
depends on; this one cannot, because tmux appears to expose no way to ask whether a client has
a menu open.

## Reproduction

Ubuntu 24.04 container, tmux 3.4, one cpu, twelve spin loops — the same harness that
reproduced #487:

```
docker run --rm --cpus=1 -v <checkout>:/src:ro ubuntu-with-tmux34 sh -c '
  cp -r /src /work && cd /work
  for i in $(seq 12); do (while :; do :; done) & done; sleep 2
  python3 -m unittest tests.test_frame_tmux_integration.MenuClientIntegration'
```

Fails roughly half of runs on:

```
FAIL: test_each_presser_gets_their_own_menu_not_the_others
AssertionError: False is not true : A's own hotkey press must open A's own, selectable menu
```

Unloaded, and at four spin loops, it passes.

## Why it is a readiness bug and not a slow machine

The sequence is:

```python
os.write(fdA, b"\x1bOQ")   # A presses the hotkey; A's menu should open
time.sleep(0.8)
os.write(fdB, b"1")        # B types 1 — must NOT select in A's menu
time.sleep(0.8)
self.assertFalse(os.path.exists(canary_a), ...)
os.write(fdA, b"1")        # A types 1 — must select
```

If the menu opens after A's `1` is written, that `1` leaks into A's pane as literal text, the
menu then sits open with nothing selecting it, and the canary never appears. Raising the
deadline on the poll that follows does not help — the keystroke has already been spent. The
same fixed sleep also carries the NEGATIVE half: `assertFalse(exists(canary_a))` only means
something while A's menu is genuinely open when B types, so a late menu makes that assertion
pass for no reason at all.

`_press_hotkey` has the identical sleep for the single-client class.

## Why it needs design rather than the helper the rest of the module now uses

A menu is a client OVERLAY, not pane content, so `capture-pane` does not see it, and none of
the client/pane formats seem to report it. Probed against tmux 3.7c with a real attached
client, before and after `display-menu`, every one of these was byte-identical except
`client_activity` (a timestamp that also moves for unrelated reasons):

`client_name`, `client_flags`, `client_control_mode`, `client_readonly`, `pane_in_mode`,
`pane_mode`, `window_active`, `session_alerts`, `client_prefix`, `client_last_session`,
`client_session`

So this wants one of:

* a format or command that does answer it (worth re-checking on newer tmux, and worth
  checking `list-clients -F '#{client_...}'` exhaustively rather than the twelve above);
* a menu whose ITEMS are observable — e.g. an item whose key, when the menu is genuinely
  open, writes a marker the test can poll for before it sends the real selection;
* or a restructure that does not need to know, if one exists that keeps the negative half
  meaningful.

Widening the sleep is not it: the failure is a keystroke landing on the wrong recipient, and a
bigger guess is still a guess. Nothing here should accept two outcomes — the assertions are
right, it is the readiness that is assumed.

### LINKED ISSUE #507: test_frame_tmux_integration still flakes outside the void-death class #487 fixed
#487 is fixed for the class it named — a death tmux never saw is now detected and retried rather than asserted against. But the module still flakes on at least one test outside that class.

## Evidence, from #495's own CI

`test_a_lone_argument_reaches_a_shell_not_execvp` (`EarlyDeathIntegration`) failed on **Python 3.11 and 3.13** in run `32786379317`, then **passed on re-run with identical bytes**:

```
FAIL: test_a_lone_argument_reaches_a_shell_not_execvp
Ran 5161 tests / FAILED (failures=1, skipped=2)
   → rerun → 8 pass
```

That makes six distinct tests in `tests/test_frame_tmux_integration.py` observed failing-then-passing on CI in one day:

- `test_a_signal_death_writes_the_unknown_death_sentinel_not_an_empty_line` (the original #409/#487 case)
- `test_a_dead_panels_hook_runs_charter_with_the_slot_and_pane_it_must_revive`
- `test_the_exit_status_path_round_trips_a_hostile_plane_path`
- `test_an_awkward_interpreter_path_arrives_byte_for_byte`
- `test_nothing_of_the_operators_tmux_is_written_by_a_whole_launch`
- `test_a_lone_argument_reaches_a_shell_not_execvp` ← this one

## Why this is a separate issue, not a reopening

#495's fix targets the measured void-death mechanism: on tmux 3.4 under load, `pane_dead=1` with an empty `pane_dead_status` and the hook never firing, permanently. Its probe detects the void and retries. That mechanism is real and the fix for it is sound — the module alone runs 42 tests OK locally, and #495's attacker approved it.

`EarlyDeathIntegration` is a different shape: it asserts on how an argument reaches the child (shell vs `execvp`), not on how a death is reported. Whatever makes it intermittent on a loaded 3.11/3.13 runner is not the void.

## What a fix needs

Diagnose *this* test's timing rather than generalising #495's remedy to it. The standing rules still apply and are what makes this hard:

- **Do not widen an assertion to accept the failure mode** — that makes it unfailable, which is the trap this suite has shipped in six distinct shapes.
- **Do not skip the module.** The behaviour under test is real; the timing of the observation is what is wrong.
- A bounded poll is necessary but demonstrably not sufficient — a sibling test already polled 100 ms × 5 s and still timed out.

## Why it matters more than a re-run

The module is on the path of every PR *and* of the release workflow. It has already cost a re-run on the irreversible publish path once. A standing "just re-run it" exception is most dangerous exactly there, because it is the one place a wrong call cannot be undone.

## Adjacent, from #495's attacker — worth folding into whatever fixes this

The probe's evidence is a **file**, so "tmux did not run the array" and "the marker could not be written" are the same reading: anything making the marker unwritable is indistinguishable from a void and gets retried. Sabotaging the probe's action (`touch` → `true`) turns a real failure into a **skip**. The fix shape it proposed: put the evidence in tmux's own state instead — `set-option -p @charter-probe ran`, read back via `show-options -p -v @charter-probe` — so "the array ran" and "a file got written" stop being the same question. Equal to main today (main's `_a_death_the_hook_saw` has used a file probe since #409), so it is harness hardening rather than a regression.

### LINKED ISSUE #531: 26 test modules define classes after their `if __name__ == "__main__"` trailer, so running one module directly silently skips them

diazoxide:
Closes #494
Closes #507
Closes #531

Phase 0, group **timing-and-hygiene**. Three ways a green run meant less than it looked
like. No `charter/` file is touched.

## #494 — a menu that was assumed open

`MenuClientIntegration` pressed the hotkey on one pty and slept 0.8 s before typing on the
other. Under load the menu was not open yet, the keystroke was spent on the pane, and no
later deadline can un-spend it. The negative half was worse: *"the other client's key must
not select in this menu"* is satisfied for free by a menu that never opened.

The issue asked for an observable. There is none in a format — all **168** names
`display-message -a` reports were probed against tmux 3.7c with two real ptys on one
session, before and after a `display-menu`, and only `client_written` (a byte counter)
moved, for the client with **no** menu as well.

But the menu is *drawn*, and these tests own the master fd of the terminal it is drawn on.
`_Screen` gives every attached client a reader thread accumulating everything tmux paints;
the tests wait for the menu's own row to arrive on the presser's terminal, and assert it
arrived on **nobody else's** — which states the class's subject positively for the first
time, instead of only through "a canary did not appear". The other client's keystroke is
waited on too, by watching the pane's own `cat` echo it, so the negative half is about a
key that was really delivered and really spent. Every 0.8 s sleep is gone from
`_drive_menu` and both two-client tests; the wait returns in ~50 ms unloaded and holds on a
loaded machine.

Two `MenuFormatIntegration` tests whose only assertion was "no canary" gained a
precondition they never had: the hostile label has to be observed on screen first. A menu
that never rendered used to satisfy them.

`test_a_row_past_the_ninth…`'s `-` keystroke also stopped being slept at: it is followed by
the first of its own nine Downs and the repaint is waited for, and a pty is a FIFO, so tmux
cannot paint a response to the Down without having read past the `-`.

## #507 — a death tmux never had the status of, read as an exit code

`EarlyDeathIntegration` is a plain `TestCase` that builds its own sessions from
`layout.session_argv`, so it inherited none of #487's void detection — and three of its
tests read a dead pane's **status**, which is exactly what a void destroys.
`_query_pane_dead_status` answers `_UNKNOWN_DEATH_CODE` (`1`) for a pane tmux holds no
status for, and `assertEqual(code, 7)` cannot tell that from a shell that exited 1. That is
the reported failure: red on 3.11 and 3.13 in one run, green on a re-run of identical
bytes.

The issue read this as "a different shape from the void". It is the same mechanism; what
was different is that this class had no defence against it. `_VoidDeaths` is extracted from
`_TmuxServerFixture` so both share one detector, and `_a_measured_death` spends a fresh
pane per void with the void **asserted** (`_the_array_never_ran`) rather than assumed — a
pane tmux destroyed, or one it holds a status for but ran no hook for, is a loud failure
naming which. The probe is armed in the server's own config file here, because there is no
gate to hold this pane open: the whole subject is a command that dies before the frame is
drawn, so it is often dead before `new-session` has printed its id. `tmux -f <conf>` is
read at server start, before any pane exists.

Nothing is widened — `7` is still the only value that passes.

### folded in: the probe's evidence is no longer a file

#495's attacker's point. `_arm_array_probe` used `run-shell "touch <marker>"`, so *"tmux
never ran the array"* and *"the marker could not be written"* were one reading, and
sabotaging the probe turned a real failure into a **skip**. It is now
`set-option -p @charter-probe ran`, read back with `show-options -p -v` — set by tmux, in
the same act as running the array, with no shell, no fork and no filesystem in between.
Measured on 3.7c: `-p` resolves to the dying pane, the value survives on a dead-but-kept
pane, and an unarmed pane answers rc 1 / `invalid option` — three distinguishable answers
where a file had two.

The option is written and read back **at arm time**, before anything depends on it. A tmux
that cannot carry a pane user option would otherwise turn every death in the module into a
void and every death-dependent test into a skip, on a machine where nothing was wrong.

## #531 — 26 modules hid classes below their `if __name__` trailer

`unittest.main()` raises `SystemExit`, so everything after the trailer is dead in the one
invocation the trailer exists for. `python3 tests/test_doctor_shadowed.py` reported
`Ran 8 tests / OK` where `python3 -m unittest tests.test_doctor_shadowed` ran 14 — and the
six it dropped were the `PersonaIso` classes that exercise the real seam. **Across the 26,
154 tests were invisible to a direct run**, measured module by module.

**The coordinator asked whether any had silently never run. None had.** Discovery imports a
module rather than executing it as `__main__`, so CI collected every class; all 154 ran and
passed. The gap belonged to whoever was debugging one file, which is when a green run is
trusted hardest. (Direct-run counts now equal discovery counts for all 26 — spot-checked
after the fix.)

The trailers moved to the ends of their files, and `tests/test_no_test_hides_below_the_main_trailer.py`
walks `tests/` with `unittest discover`'s own pattern and fails on the 27th, naming the
module, the line, what got hidden, and both ways out — the `RealPlaneRead` standard. The
rule is *nothing* below the trailer, not *no class*: `SystemExit` hides a helper or a
constant just as completely.

## Not fixed, named rather than papered over

`WindowInsideAnOperatorsTmux.test_nothing_of_the_operators_tmux_is_written_by_a_whole_launch`
(one of the six flaky tests #507 lists) reads the code a whole real `cmd_launch` returns,
and that pane carries charter's production `pane-died` **pair** — `[0]` the write hook,
`[1]` `kill-session`. There is no free index: `[1]` is where a probe goes and is charter's
teardown, and measured on 3.7c a `[2]` armed beside them never runs at all, because
`kill-session` at `[1]` takes the server down first. A void death there is still read as an
exit code of `1`. Recorded in the module docstring.

## Mutations actually run

Each applied, run, then restored, with `__pycache__` cleared. Results as observed, not as
expected:

| mutation | result |
| --- | --- |
| `cmd_menu` sleeps 2.0 s before drawing — **this branch** | **GREEN** (2 tests, 10.9 s) |
| the same mutation against `origin/main`'s own test file | **RED** — `AssertionError: False is not true : A's own hotkey press must open A's own, selectable menu`, the issue's exact words |
| `cmd_menu` draws on the OTHER attached client | **RED** — `no menu carrying 'Detach chci-…-a' was ever drawn on A's own terminal` |
| `_Screen.saw` always returns True, with the 2 s delay | **RED** — the "not on the other client" half catches a permissive readiness check |
| `menu_argv` drops `-c client` | **GREEN**, and correctly so: `cmd_menu`'s own docstring already records that tmux's default guess was never reproduced failing for one session with several clients — the two-*frame* case is what `-c` fixes. Reported because it did not go red, not hidden. |
| the probe hook's action records nothing (`touch` → a no-op, the attacker's sabotage) | **RED**, 4 failures — `tmux never ran this pane's pane-died array, yet tmux is holding a status ('1') for it`. A failure, **not a skip**. |
| `_PROBE_OPTION` set to a name tmux refuses | **RED** at arm time — `this tmux will not set the pane option … and skip every death-dependent test on a healthy machine` |
| `session_argv` word-splits the harness argv | **RED** — `1 != 7 : a single argument must reach a shell` |
| `test_doctor_shadowed.py`'s trailer put back above its last three classes | **RED** — names all three, both line numbers, and both ways out |
| `hidden_below_the_trailer` stops looking | **RED** — 3 of its own controls |

## Four environments

`config.ROOT` printed before each. Whole suite, `unittest discover -s tests`.

| environment | `ROOT` | result |
| --- | --- | --- |
| 1. fresh checkout, ambient cleared | the fresh tree | `Ran 5621` — **OK** |
| 2. a used plane (`.charter/`, `.superpowers/`), ambient cleared | the operator's | `Ran 5621` — **OK** |
| 3. inside a live charter frame | the operator's | `Ran 5621` — 17 failures |
| 4. `CHARTER_WORKSPACE=anything` | the operator's | `Ran 5621` — 8 failures |

3 and 4 are the ambient-variable leaks this phase's *other* groups own, and both were
attributed rather than assumed:

* **Environment 3**: `origin/main`, stashed into the identical cwd and plane, gives
  **18** — the same 17, **plus**
  `MenuClientIntegration.test_a_submenu_opens_on_the_presser_and_its_rows_are_selectable`.
  That is #494 flaking on this machine, in a live frame, and this branch no longer has it.
  The 17 are #519/#521 — `test_frame_doctor` reading `harness-wrapper-17391` out of the
  developer's terminal, `statusline.main` printing nothing because a frame owns the
  surface, `session.current()` answering from the ambient variable.
* **Environment 4**: `origin/main` gives the same **8**, byte for byte — #528's class,
  `CHARTER_WORKSPACE` leaking into workspace resolution.

So the four answers still differ, and this group is not what makes them differ: with
#519/#521/#528 outstanding, environments 3 and 4 cannot yet agree with 1 and 2. Flagged for
the phase rather than claimed as done here.

Not merged. No version bump, no stamping, no tag.

### DIFF
--- a/docs/news/unreleased-a-green-tmux-run-now-means-something.md
+++ b/docs/news/unreleased-a-green-tmux-run-now-means-something.md
@@ -29,6 +29,21 @@ Two tests whose only assertion was that a hostile `#(…)` label creates no cana
 precondition they never had: the label has to be observed on screen first. A menu that never
 rendered used to satisfy them.
 
+One test resisted the whole approach, and it is worth naming because the first fix for it was
+wrong. `-` is not tmux's spelling for "no shortcut" — it is an ordinary key, which is why rows
+past the ninth are keyed with the empty string — and the test that measures this presses `-` on
+a rendered eleven-row menu and requires that nothing ran. Waiting for tmux to repaint proves
+the `-` was *read*; it does not bound the `run-shell` → `charter frame-action` → `subprocess`
+chain whose absence is the whole assertion, and a repaint arrives two orders of magnitude
+sooner than that chain finishes. Swapping the sleep for the repaint made the test pass whether
+or not `-` fired the row — the same defect as the sleep, wearing a fix's clothes. An assertion
+about an absence needs a *bound*, not a readiness signal, so it now gets one from a pacing
+control: the eleventh row's own action, dispatched to the same tmux server only after the
+repaint proves the `-` was consumed. Identical work, started strictly later — so when the
+pacer's canary lands, anything `-` could have started has had longer, and the absence is
+measured. Restoring the constant `-` that this row exists to refuse fails the test again;
+without the pacer it did not.
+
 **A death tmux never had the status of, read as an exit code.** On a loaded tmux 3.4 about one
 pane death in seventeen closes the pane's fd before tmux has the child's status, and never
 gets it — `pane_dead=1`, an empty status, the `pane-died` hook array never run, permanently.
--- a/docs/news/unreleased-a-green-tmux-run-now-means-something.md
+++ b/docs/news/unreleased-a-green-tmux-run-now-means-something.md
@@ -1,0 +1,68 @@
+---
+version: unreleased
+headline: The tmux tests wait for the thing itself instead of sleeping, a dead pane tmux never had a status for stops being read as an exit code, and no test module can hide classes below its `if __name__` trailer any more
+---
+
+Three ways charter's own suite could be green about something it had not measured. None of
+them reaches a plane you run; all three reach anyone who trusts a green run — which, on the
+release workflow, is the one place a wrong call cannot be undone.
+
+**A menu the test only assumed was open.** `MenuClientIntegration` pressed the hotkey on one
+attached terminal, slept 0.8 seconds, and then typed into the other. Under cpu contention the
+menu was not open yet, so the keystroke landed in the pane instead, and the test failed on an
+assertion about charter — about half of runs on a one-cpu tmux 3.4. The negative half was
+worse than flaky: "the other client's key must not select in this menu" is satisfied for free
+by a menu that has not opened, so a late menu made that assertion pass for no reason at all.
+
+There is no format that reports an open menu. All 168 that `display-message -a` names were
+probed against tmux 3.7c with two real terminals attached to one session, before and after a
+`display-menu`: only a byte counter moved, and it moved for the terminal with no menu too.
+But a menu is *drawn*, and these tests own the master fd of the terminal it is drawn on. Each
+attached client now has a reader thread accumulating everything tmux paints on it, and the
+tests wait for the menu's own row to appear there — on the presser's terminal and, asserted
+separately, on nobody else's. The 0.8 s guess is gone from every one of them; the wait
+returns in about 50 ms on an unloaded machine and holds on a loaded one. The other client's
+keystroke is waited on too, by watching the pane's own `cat` echo it — so "it did not select"
+is now about a key that was really delivered and really spent.
+
+Two tests whose only assertion was that a hostile `#(…)` label creates no canary gained a
+precondition they never had: the label has to be observed on screen first. A menu that never
+rendered used to satisfy them.
+
+**A death tmux never had the status of, read as an exit code.** On a loaded tmux 3.4 about one
+pane death in seventeen closes the pane's fd before tmux has the child's status, and never
+gets it — `pane_dead=1`, an empty status, the `pane-died` hook array never run, permanently.
+charter reports an unknown death as `1`. `EarlyDeathIntegration` compares that number against
+`7` and says "a single argument must reach a shell", about a machine where the argument
+reached a shell perfectly well. That is what CI showed: red on two Python versions in one run,
+green on a re-run of identical bytes, and it has already cost a re-run on the publish path.
+
+Those four tests now spend a fresh pane when a trial measured nothing, exactly as the rest of
+the module has since #487 — with the void *asserted* rather than assumed, so a pane tmux
+destroyed, or one it holds a status for but ran no hook for, is a loud failure naming which.
+The probe has to be armed in the server's own config file here, because there is no gate to
+hold this pane open: the whole subject is a command that dies before the frame is drawn.
+Nothing is widened; `7` is still the only value that passes.
+
+**The probe's evidence moved into tmux's own state.** It used to `touch` a file, so "tmux never
+ran the array" and "the marker could not be written" were the same reading — sabotage the
+probe and a real failure became a *skip*. It is now a tmux pane option, set by tmux in the same
+act as running the array, with no shell and no filesystem in between. The mechanism is proved
+usable — written and read back — before anything depends on it, so a tmux that cannot carry the
+evidence says so by name instead of turning every death into a skip.
+
+**And 26 test modules hid classes below their `if __name__` trailer.** `unittest.main()` raises
+`SystemExit`, so everything written after it is dead in the one invocation the trailer exists
+for: `python3 tests/test_doctor_shadowed.py` reported `Ran 8 tests / OK` where
+`python3 -m unittest tests.test_doctor_shadowed` ran 14 — and the six it dropped were the ones
+that exercise the real seam, the eight that survived the pure-function ones. Across the 26,
+154 tests were invisible to a direct run. CI never saw it, because discovery imports a module
+rather than executing it; all 154 do run and do pass there. The gap belonged to whoever was
+debugging one file, which is when a green run is trusted hardest.
+
+The 26 trailers moved to the ends of their files, and a guard now walks the suite and fails on
+the 27th — naming the module, the line, what got hidden and both ways out. All 26 got there the
+same way, by someone appending a class to a finished file and not noticing the trailer above
+the insertion point, so the guard is the point and the sweep is the smaller half.
+
+Nothing to adopt: none of this changes what charter does on your plane.
--- a/tests/test_asks_do_not_hang_bypass_permissions.py
+++ b/tests/test_asks_do_not_hang_bypass_permissions.py
@@ -153,10 +153,6 @@ class TestTheInjectedProseStopsTellingTheAgentToQuiz(InAControlPlane):
             self.assertIn("Scout first", ctx)
 
 
-if __name__ == "__main__":
-    unittest.main()
-
-
 class TestAnUnattendedRunWithNoWorkspaceFailsFast(InAControlPlane):
     """The one block that does NOT get an assume-and-continue rewrite. Every other nudge
     names a preference; this one names a missing input, and guessing it silently claims
@@ -197,3 +193,7 @@ class TestASuppressedAskIsCountedExactlyOnce(ModeCase):
     def test_attended_records_only_the_ask(self):
         self.nudge("default", sid="d")
         self.assertEqual(["dispatch-ask"], self.events("d"))
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/tests/test_browser_install.py
+++ b/tests/test_browser_install.py
@@ -79,10 +79,6 @@ class TestBridgeGuidanceSurvives(unittest.TestCase):
         self.assertIn("never read a filled secret back", body)
 
 
-if __name__ == "__main__":
-    unittest.main()
-
-
 class TestTheGeneratorCannotHangOrPromptInvisibly(unittest.TestCase):
     """Two ways the same command stalls with nothing on screen.
 
@@ -210,3 +206,7 @@ class TestTheInstallStatesBothPostures(unittest.TestCase):
         self.assertIn(str(browser.SKILL_DIR), said)
         self.assertIn(str(browser.CONFIG_DIR), said)
         self.assertIn("docs/browser.md", said)
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/tests/test_cli_smoke.py
+++ b/tests/test_cli_smoke.py
@@ -123,10 +123,6 @@ class CliSmokeTest(unittest.TestCase):
         self._assert_clean(self._run("persona", "secret", "--help"))
 
 
-if __name__ == "__main__":
-    unittest.main()
-
-
 class TestInitDoesNotPrintOneEnormousLine(unittest.TestCase):
     """The headline is a count; the paths go underneath it.
 
@@ -170,3 +166,7 @@ class TestInitDoesNotPrintOneEnormousLine(unittest.TestCase):
         self.assertEqual(p.returncode, 0, out)
         for expected in ("charter.toml", "personas/", "inventory/", "workspaces/"):
             self.assertIn(expected, out)
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/tests/test_docs_show.py
+++ b/tests/test_docs_show.py
@@ -145,10 +145,6 @@ class TestDocsCli(unittest.TestCase):
         self.assertIs(args.func, commands.cmd_docs)
 
 
-if __name__ == "__main__":
-    unittest.main()
-
-
 class TestTheContainmentCheckIsReal(unittest.TestCase):
     """`page.parent` is `root` by construction, so checking it asked a question whose
     answer was always yes. A symlink inside the docs directory pointed anywhere on disk,
@@ -205,3 +201,7 @@ class TestTheCheckoutFallbackIsNotSitePackages(unittest.TestCase):
         docsrc._CHECKOUT = repo / "docs"
         self.assertEqual(docsrc.source(), repo / "docs")
         self.assertEqual(docsrc.read("install"), "# ours\n")
+
+
+if __name__ == "__main__":
+    unittest.main()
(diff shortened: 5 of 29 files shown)


diazoxide:
Round two. Rebased onto `c1f9b95` (post-#540); one commit added.

### The weakened test is un-weakened

`test_a_row_past_the_ninth_is_fired_by_no_keystroke_but_is_still_reachable` had its
`time.sleep(0.8)` replaced with `await_more_drawn`, which returns the instant tmux repaints
(0–50 ms). That proves the `-` was **read**. It does not bound the
`run-shell` → `sh` → `"$CHARTER_PY" -m charter frame-action a9` → `subprocess.run` chain whose
absence is the whole assertion (`charter frame-action` alone measures ~0.08 s here, before the
argv it spawns). The assertion fired long before the effect it checks for could have happened,
and passed whether or not `-` fires the row — #494's own failure mode, one line lower.

The `-` half now gets a bound from a **pacing control**: the eleventh row's own action string,
lifted verbatim out of the argv `menu_argv` built (asserted to differ from the tenth row's by
the action id alone, so the two cost the identical interpreter start, `charter` import and
`subprocess.run`), dispatched to the same tmux server *after* the repaint proves the `-` was
consumed. tmux is single-threaded: had `-` been bound, tmux forked that item's job inside the
key event, which completes before it returns to the command socket for this one. Same work,
started strictly later — so the pacer's canary landing means anything `-` could have started
has had at least as long. Asserted positively first, so a pacer that never ran fails loudly
instead of licensing the negative for free.

**Mutation** — `menu._key` → `return str(i + 1) if i < 9 else "-"`, with
`assertEqual(cmd[10:][1::3][9:], ["", ""], cmd)` neutralised so it reaches the rendered-menu
half rather than the constant pin:

| | result |
|---|---|
| with the pacing control | **RED 5/5**, always on the `-` assertion itself; RED again under 16 cpu burners |
| pacer block deleted | GREEN 3/3 — the weakened form reproduced |
| restored, unmutated | GREEN; module GREEN, 53 tests |

### The five environments now agree

```
1 cleared / fresh checkout       Ran 5652 … OK
2 cleared / used plane           Ran 5652 … OK
3 live frame env vars            Ran 5652 … OK
4 CHARTER_WORKSPACE=anything     Ran 5652 … OK
5 EDM_WORKSPACE=anything         Ran 5652 … OK
```

(5652 vs main's 5645 = the new `test_no_test_hides_below_the_main_trailer` guard.)

### Two pre-existing defects found while running environment 3, filed not fixed

Running the suite from **inside a real tmux pane** — rather than only exporting a live frame's
variables — hangs, twice, for reasons that are on `main` and in no file this PR touches:

* **#545** — `test_frame_launcher`'s `_launch`/`_launch_inside` pin `sys.stdout.isatty` and
  inherit `sys.stdin.isatty`. `_picker_wanted` reads both, so from any terminal the #518
  workspace picker opens and `input()` never returns. Sampled: `builtin_input` →
  `PyFile_GetLine`, fd 0 on the pane's tty. CI never sees it; a developer running the suite in
  their own shell always would.
* **#546** — `test_vault_registration` runs the operator's **real** `op` binary against their
  real 1Password vaults (four invocations, two tests, measured with a logging stand-in on
  `$PATH`). Green only because an unauthenticated `op` fails fast; in a fresh pane it blocks on
  authorization and the suite parks in `subprocess.communicate` — 15 minutes and counting.

Both are the same family as #544 and are left for their own change rather than widened into
this one.

Issue #577, opened 2026-08-27
`valid_name` and `workspace_name_ok` decide on `$`, which is a SPELLING of "the end", not the end


diazoxide:
`valid_name` and `workspace_name_ok` decide on `$`, which is a SPELLING of "the end", not the end

`persona._NAME_RE` is `^[a-z0-9][a-z0-9._-]*$` and `instance.WORKSPACE_NAME_RE` is
`^[A-Za-z0-9][A-Za-z0-9._-]*$`. In Python, `$` matches at the end of the string **or just
before a trailing newline**, and both are used with `.match()`, never `.fullmatch()`. So a
name with a trailing `\n` is admitted by the rule whose whole job is to say it is not.

```
persona.valid_name("evil\n")          -> True
persona.reference_ok("evil\n")        -> True
persona.reference_refusal("evil\n")   -> None
instance.workspace_name_ok("evil\n")  -> True
```

**Reproduced end to end, against the current tree.** A directory `personas/evil<LF>/`
holding an ordinary `persona.md` resolves, loads, and `sync-agents` renders it:

```
---
name: evil

description: "The Victim persona. Delegate victim tasks to it. Holds no credentials of its own."
---
<!-- GENERATED by `charter persona sync-agents` from personas/evil
/persona.md — edit the persona, not …
```

The name writes a blank line into the generated agent's YAML frontmatter, and the
`GENERATED by` comment is split across two physical lines by a path that is one name.

**Why this is worth a rule change rather than a `strip()`.** `mcpseen.label`'s docstring
states the invariant this breaks in as many words: *"the `persona` half is already
contained: it is a directory under `personas/`, and `persona.reference_ok` refuses any
reference outside `[a-z0-9][a-z0-9._-]*` (#328), so a persona whose name is not printable
ASCII resolves to nothing and never reaches this line."* That sentence is false for exactly
one codepoint, and `\n` is the one codepoint the surrounding fix (#453) is about.

**Nine regexes in the tree share the shape.** Every anchored name/identifier rule is
`^…$` + `.match()`, and every one of them admits its own alphabet plus a trailing newline:

| regex | admits `<name>\n` | protected upstream? |
| --- | --- | --- |
| `persona._NAME_RE` | yes | **no** — confirmed live above |
| `instance.WORKSPACE_NAME_RE` | yes | **no** — `workspace_name_ok` does not strip |
| `docsrc._TOPIC` | yes | no, but the join then names no file |
[...]

Issue #569, opened 2026-08-27
Sweep harness: the trace map's own guards are unpinned, and stages B and C are not built


diazoxide:
Follow-ups from stage A (#565). Recorded now so the gaps are visible rather than assumed closed.

## The one to fix first

The harness swept itself: **132 of 196 mutations survived.** That is not a good number and the breakdown is what matters:

| area | pinned |
|---|---|
| operator table | 31 / 60 |
| verdict logic | 17 / 20 |
| report renderer + CLI | ~0 / 49 |
| **trace map** | **0 / 9** |

The renderer and CLI are cosmetic and can wait. **The trace map cannot.** It decides which tests a mutation runs against, so a silent narrowing there means mutations run against fewer tests than they should, survive nothing, and score **pinned**. That is a false-pin generator sitting under everything else the tool says — the exact failure the tool exists to prevent, in the tool.

Fix the trace map's 9 first, then the operator table's 29.

## Stage B — the baseline, not built

Run `--all` against `main` and write the number down. First honest count of the repo's unpinned guards. It also sets what stage C has to hold flat.

## Stage C — the CI gate, not built

Scoped to added lines, fails on a survivor. **Only after B**, because a gate whose baseline nobody has seen gets disabled the first time it is inconvenient. See also #542: the suite's per-run cost is multiplied by the mutation count, so that is worth fixing before this becomes per-pull-request.

## Two operator families missing

Every miss in the acceptance run was an operator-shape gap — **not** a selection bug and **not** a masked mutant. Eleven rows had no operator at all:

- **string / regex constants** (×5)
- **semantic near-synonym swaps** (×3)
- statement reordering (×2), multi-statement branch body, starred list element

The first two are the families worth adding next. Neither has a principled general form, which is a fair defence of the spec's shape table — and the reason they were reported rather than quietly fitted to make the hit rate look better.

## Second-order mutation, cross-function

`--second-order N` pairs survivors sharing an enclosing function. That does **not** reach the `layout.py:291` case (#553), because there the masking partner — a lookup *ordering* inside `_edge_of`/`_size_of` — **is not a survivor at all**, so no pairing of survivors can find it. It needs a reordering operator or a behavioural oracle.

So the single-order number is not a complete answer. Stage A compensates by refusing suppression, naming co-located survivors, and printing what the covering tests assert. Worth deciding whether that is enough before stage C makes it a gate.

## One live hazard the audit found

[...]

Issue #498, opened 2026-08-24
`contain.one_line` decides on a CATEGORY LIST, so a name that renders as nothing survives it


diazoxide:
`contain.one_line` decides on a CATEGORY LIST, so a name that renders as nothing survives it

`charter/contain.py:190` escapes a character when `unicodedata.category(ch)` is in
`_INVISIBLE = {"Cc", "Cf", "Cs", "Zl", "Zp"}`, or when it is whitespace other than
`" "`. That is a list of spellings, and the blank-rendering class is not inside it:

| codepoint | category | `one_line` output |
| --- | --- | --- |
| U+3164 HANGUL FILLER | `Lo` | unchanged |
| U+2800 BRAILLE PATTERN BLANK | `So` | unchanged |
| U+115F HANGUL CHOSEONG FILLER | `Lo` | unchanged |
| U+1160 HANGUL JUNGSEONG FILLER | `Lo` | unchanged |

All four render as nothing on every terminal, none is `isspace()`, and all four survive
`strip`.

`one_line`'s own docstring is honest about scope — it guarantees a value cannot forge a
second ROW, not that a value is readable — so this is not a defect in that function's
stated contract. It is a defect in the callers that need the second property and ask this
function for it.

**Reproduced, against the current tree:**

```
three HANGUL FILLERs (Lo)  -> lint row prefix: '✗ ㅤㅤㅤ: ...'   (rendered width 3)
three BRAILLE BLANKs (So)  -> lint row prefix: '✗ ⠀⠀⠀: ...'    (rendered width 3)
an ordinary name           -> lint row prefix: '✗ devops: ...'  (rendered width 6)
```

Three call sites want readability and get line-safety:

- `charter/commands_persona.py:1113` — the `lint` row prefix. `n` comes from
  `list_personas()`, which globs `personas/*/` and only excludes a leading underscore, so
  it is a directory name a commit chose. The row reads `✗ : <issue>` and names no persona.
- `charter/persona.py:1245` — `persona '<name>' does not load`, same input, same result:
  the sentence telling somebody which persona to go fix does not contain its name.
- `charter/commands_persona.py:720` — script paths in the brief the sub-agent reads. A
  `bin/` filename is committed; a blank-rendering one produces a bullet whose backticked
  path is empty, in the one document aimed at the model.

**Not a duplicate of #453.** That one is about a name FORGING a second line, which
`one_line` does stop. This is about a name that forges nothing and says nothing.

[...]

Issue #561, opened 2026-08-26
A check that never ran reports CLEAN, and nothing on main requires it to have run


diazoxide:
Found while merging #559/#560, and it is the same shape as #558 one level up.

## What happened

I pushed a commit to `fix/558-dispatch-version-guard` at 16:47Z. Seven minutes later:

```
$ gh api repos/diazoxide/charter/git/ref/heads/fix/558-dispatch-version-guard -q .object.sha
4916caaeeb0ba3760cbff8226751df213cd40dc3      # the push landed

$ gh api repos/diazoxide/charter/commits/4916caae/check-runs -q .total_count
0                                              # no run was ever created

$ gh pr checks 560
no checks reported on the 'fix/558-dispatch-version-guard' branch

$ gh pr view 560 --json mergeable,mergeStateStatus
mergeable=MERGEABLE  mergeStateStatus=CLEAN     # <-- merge is offered
```

`test.yml` triggers on `push:` with no branch filter, and there is no `concurrency:` block. The run simply was not created. A `workflow_dispatch` on the same sha then ran and passed on 3.11–3.14.

**This is not a one-off.** The agent on #559 hit it independently and unprompted on a different branch: `total_count: 0` for both branch and head sha on first push, requiring a dispatch and then a force-push of an identical tree to get runs. There is also main's own last push run (`32234bed`) which came back `conclusion: failure` with **every job `cancelled`** after 16 minutes, and an earlier one recorded as `startup_failure`.

So: GitHub is intermittently swallowing Actions triggers on this repository right now.

## Why that is charter's problem and not only GitHub's

```
$ gh api repos/diazoxide/charter/branches/main/protection
{"message": "Branch not protected"}
```

`main` has **no branch protection and no required status checks**. So when a run is swallowed, the pull request does not say "checks are missing" — it says **CLEAN**, and offers the merge button. `gh pr checks` distinguishes "no checks reported" from "checks passed", but `mergeStateStatus` does not, and neither does anything that gates the merge.

That is precisely the defect #558 just fixed in `release.yml`: **a step that was skipped reports success.** Here it is an entire check run that was never created reporting CLEAN. Same property, different layer — an absent verification is indistinguishable from a passed one.

It nearly bit: #560 was in exactly this state and would have merged unverified. It was caught only because the merge check was written as "is CI green **for this head sha**" rather than "does the PR look clean", and that difference should not depend on who is doing the merging.

## Proposed

[...]

PR #583, opened 2026-08-27
A persona frontmatter key is honoured or reported, and `Borrows:` no longer fails open (#575, #509)

### LINKED ISSUE #509: persona.parse keeps the LAST of two lines with the same key, so a duplicated frontmatter field is resolved by line order and reported by nothing
`persona.parse` builds a flat dict with `meta[key] = value` in line order, so two lines
carrying the same key silently keep the **last** one. Nothing anywhere reports that a key
was written twice.

For a news ordering field (#486) that is the value half's failure reached one row over
from #503's key half:

```
---
version: 0.60.0
headline: important
security: true
security: false
---
```

```
>>> persona.parse(text)[0]
{'version': '0.60.0', 'headline': 'important', 'security': 'false'}
>>> e = news._read(p); e.security, e.bad, news.ordering_errors([e])
(False, (), [])
```

The entry declares a security fix on one line, is read as ordinary, sorts below the
ordinary entries, and `charter news --for` exits 0 with an empty stderr. Same shape as the
empty-value collapse fixed on #486's branch — the field was declared, charter had the
information, and the sort acted as though it had not — except that this one is invisible
to `news` because `parse` has already thrown the first value away by the time `_read` sees
the dict.

It is not confined to news. `persona.parse` is the reader for every persona definition,
every skill and every news entry, so the same silence covers `vault:` declared twice,
`extends:` declared twice, `tools:` declared twice — a definition whose first line is the
one the author meant and whose last line is the one charter obeys.

**Property:** a frontmatter key written twice is a contradiction in the file, not a
value to be picked. It should be reported by name, the way #486's two `lead: true`
entries are, rather than resolved by line order.

**Why it is not folded into #486:** the fix has to live in `persona.parse`, whose return
type is a plain `dict` read by every caller in the tree. Detecting the duplicate means
either changing what `parse` returns or giving it a second output, and both land on
`persona.load`, `structural_errors`, `resolve`, `tools_of`, `vault_of`, `docsrc`,
`news._read` and the skill lint at once. #503 is blocked on the same decision — the two
should be fixed together, since both are "the dict lost information the file carried".

**Next spelling after this one:** a key that differs only by surrounding whitespace inside
the key (`security :`) — `parse` strips the key, so that IS the same key and does collide;
and a key differing by case, which does not (#503). Whatever fix lands has to say which of
those it treats as the same key, because "the same key" is exactly the property in
dispute.

Found while closing the empty-value bypass on #486's branch
(https://github.com/diazoxide/charter/pull/490).

### LINKED ISSUE #575: A persona frontmatter key spelled in any other case is read by nothing — and `Borrows:` fails OPEN
#503 is this defect in `news`, and it says in as many words that the fix does not live there:

> `persona.parse` is the shared frontmatter parser and its dict is read by key at every
> caller — `persona.load`, `persona.structural_errors`, `docsrc`, `news._read` — with keys
> like `role`, `vault`, `delegate-when`, `extends`, `tools`, `check`, `adopt`, `headline`.

PR #573 closes the `news` half by declaring a closed key set and reporting anything outside it. That is deliberately scoped to `news`, because a news entry's key set is closed and a persona definition's is not obviously so. **The persona half is untouched and is the half with teeth.**

## The same silence, one subsystem over

```python
(d / "demo" / "persona.md").write_text(
    "---\nname: demo\nrole: A demo\nVault: devops\nExtends: base\n---\n\ncharter body\n")
```

```
load meta:          {'name': 'demo', 'role': 'A demo', 'Vault': 'devops', 'Extends': 'base'}
vault_of:           None
structural_errors:  []
```

The persona declares a vault and has none. It declares a parent and inherits nothing. `persona lint` — the signal an operator actively checks — says the definition is fine.

## `Borrows:` is the one that fails open

The rest fail closed: a miscased `Vault:` means no credentials, a miscased `Tools:` means no auto-approvals. `borrows:` is the exception, and by design — `borrows_of` returns `None` for an *absent* key on purpose, and `None` means "keep the legacy `uses:` grant", which is the **wider** one:

> ``None`` and ``[]`` are different answers and the difference is the whole feature:
> absent means "I have not opted in, keep the legacy `uses:` grant", while
> ``borrows: none`` means "I borrow nothing".

An author who writes `Borrows: none` to opt *out* of the wide grant gets the wide grant:

```python
front:    uses: forge, release   Borrows: none
forge:    tools: Bash(gh:*)
release:  tools: Bash(git push:*)
```

```
borrows_of(front):      None
effective_tools(front): ['Bash(gh:*)', 'Bash(git push:*)']
structural_errors:      []
```

Both borrowed personas' tools auto-approved, from a definition whose author wrote the word that grants nothing. This is #257's defect reached through the key: "a front door declaring `uses: forge, release` could do both personas' work with both personas' tools and never pay a prompt."

## The property, and why case is not it

Same as #503's. **A key the author declared is honoured or reported, never neither.** Case-folding the lookup answers `Vault:` and leaves `vualt:`, `borrow:`, `delegate_when:` and `extend:` exactly as silent — every one of them parses cleanly, is looked up by nothing, and takes the default.

So the shape worth having is probably the one #573 used: a closed set per consumer, reported by `persona.structural_errors` — which is already the function whose job is "say what is wrong with this definition", already reached by `persona lint`, and already the signal #329/#337 are about being honest. It needs a decision `news` did not: whether a persona definition may carry keys charter does not read (a harness's own field, a comment-ish key), and if so which.

Folding `persona.parse` itself is the option to weigh against it, and #509 is the reason it needs a rule either way: two keys differing only in case have to resolve to one, or be refused.

## Where

* `charter/persona.py` — `parse`, `load`, `resolve`, `vault_of`, `tools_of`, `borrows_of`, `uses_of`, `structural_errors`
* `charter/news.py` `_KNOWN_FIELDS` / `entry_errors` is the worked example, added in #573

Found while fixing #502/#503, filed rather than folded in for the blast-radius reason #503 gives.

diazoxide:
Two defects in `persona.parse`'s handling of frontmatter keys, fixed as one property:
**a frontmatter key is honoured or reported — never silently resolved.** One is "the key
was not recognised" (#575), the other "the key was recognised twice" (#509), and both ended
in a value chosen by accident with no diagnostic. One of them is a permission grant.

## The fail-open, which is the point

`borrows_of` answers from the key's **absence**: `None` means "I have not opted in — keep
the legacy `uses:` grant", which is the *wider* one. Every other miscased key fails toward
less (a miscased `Vault:` means no credentials); this one failed toward more.

```
uses: forge, release
Borrows: none            ← read by nothing
```
```
borrows_of(front):      None
effective_tools(front): ['Bash(gh:*)', 'Bash(git push:*)']
structural_errors:      []
```

Both borrowed personas' tools auto-approved at `toolgate.decide`, from a definition whose
author wrote the word that grants nothing.

**A `borrows:` charter cannot read now answers `[]` and never `None`.** `None` means the
author never mentioned borrowing, and nothing else. Pinned by
`test_the_fallback_cannot_come_back`, which asserts on the sentinel as well as the tool set
— the two are one line apart in `effective_tools`, and a later edit could restore the
fallback while some other narrowing kept the fixture's tools looking right.

## I checked the other keys, and found two more fail-opens

The brief asked for `vault:`, `extends:`, `uses:`, `role:`. Those four fail *closed*
(no credentials, no inheritance, no reuse, no role) — but the sweep of the whole grant path
turned up three the issues did not name:

| key | miscased consequence |
|---|---|
| `Agent-tools:` | **no `tools:` line is emitted, and no `tools:` line means the sub-agent inherits EVERY tool.** The allowlist does not narrow — it vanishes. |
| `Disallowed-tools:` | the denylist vanishes the same way |
| `Draft:` | `is_draft` is False, so the unfinished charter `sync-agents` exists to withhold becomes a sub-agent's system prompt |
| `Extends:` | a child inherits nothing, so a parent that opted out with `borrows: none` hands the child the legacy grant it renounced |

All four were reached through a run printing `✓ Synced 1 persona sub-agent(s)`. As in #573,
the sibling was wider than the named bug. `sync-agents` now generates no agent from a
definition charter could not read, removes the stale one, and names the key;
`_GRANT_DECIDING_KEYS` covers the `extends:` path.

## What happens to somebody with a miscased key committed today

Stated explicitly, because it is a behaviour change:

* A miscased or duplicated key goes from a lint **warning** to an **error**, and appears on
  the status line rather than only in `charter persona lint`.
* The tool grant changes in exactly one shape: a `Borrows:`/`Extends:` charter could not
  read, which today silently grants the borrowed personas' tools and after this grants none.
* `sync-agents` writes no agent for that persona until the key is fixed.
* **Nothing changes** for a definition whose keys are spelled once and correctly —
  including the legacy `uses:` grant for a persona declaring no `borrows:` at all
  (`test_an_absent_borrows_still_means_the_legacy_grant`).

charter's own five personas and all 109 shipped news entries are clean.

## Refused, not case-folded — #573's call, kept

Folding the *lookup* answers `Vault:` and leaves `vualt:`, `borrow:` and `delegate_when:`
exactly as silent. The **closed vocabulary** is what catches all of them, and it already
existed — `persona lint` has long warned about a key charter neither reads nor emits.
(#575's claim that lint "says the definition is fine" is half right: `structural_errors` was
empty, but `lint` did warn. The warning's wording was the problem — "it does nothing
(typo?)" is exactly wrong for `Borrows:`, whose *absence* does a great deal.)

What changed is which finding carries weight:

* a key charter **does not read** stays a **warning** — a harness's own field is a
  legitimate thing to carry in a committed file, and an error there would break planes that
  are correct;
* a key charter **does read**, miscased or written twice, is an **error** — charter can name
  the word the author was reaching for, which is what makes it actionable and what lets the
  grant act on it.

Nothing is ever read out of the miscased key. `misspelled_key` folds the *sentence and the
severity*, never the lookup.

## #509 without the change it feared

The issue read the fix as "either change what `parse` returns or give it a second output",
both landing on `persona.load`, `structural_errors`, `resolve`, `tools_of`, `vault_of`,
`docsrc`, `news._read` and the skill lint at once. Neither happened: `_frontmatter` is the
one line-walk, **`parse` still returns `(dict, body)` and every caller of it is untouched**,
and the second answer is carried by `load` — where the persona-shaped questions already
are. `duplicate_keys(text)` is exposed for `news` and the skill lint to adopt when their
owners want it; I stayed out of `charter/news.py` per the brief, so #509's news half is
one call away rather than taken.

The issue also asks which spellings collide. Answered by test: `vault :` **is** `vault`
(`parse` strips), `Vault:` is **not** — it is reported as miscased rather than duplicated.

## Verification

* **Full suite, two local environments.** Python 3.14.4 → 6535 tests OK. Python 3.12 with
  `CHARTER_*`/`CLAUDE_*`/`ANTHROPIC_*`/`TMUX*` unset → see the comment below.
* **CI** at the head sha via `gh api .../check-runs` (3.11/3.12/3.13/3.14).
* **Baseline-diffed test verification.** All 45 new tests were run against unfixed
  `bf58b46`: **37 fail** (23 failures + 14 errors). The 8 that pass are exactly the
  no-change pins — an absent `borrows:` still meaning the legacy grant, a correctly spelled
  `borrows:` still granting what it names, an unknown key still a warning and still
  generating an agent, `parse` still returning `(dict, body)`, and the value never being
  read out of a miscased key.
* **`tools/sweep.py`** on the branch.
* **Hand-check, reported separately** (below), because the tool has no operator for string
  constants or semantic swaps (#569) — which is the shape of a key name and a case
  comparison.

### Hand-check: 30 mutations, 29 caught, 1 equivalent

Run against an **unmutated baseline** (`rc=0, failures=0`) and diffed against it, per #579:
a mutation only counts as caught when it adds a failure the baseline did not already have.

Three survived the first round, and none was equivalent:

* **`contain.readable` dropped from the miscased row** — the only unprintable-key fixture
  was an *unknown* key, which lands on a different row. A case variant can be non-ASCII:
  `Borrowſ` casefolds to `borrows` (U+017F). Pinned.
* **the gate moved after the draft check** — no fixture was both a draft and miscased. The
  agent is withheld either way, so this is about which sentence the operator gets. Pinned
  in that order.
* **`_FOLDED_KEYS` built unfolded** — genuinely equivalent, and only *because* every
  vocabulary key is lowercase. That invariant is what makes "miscased" mean anything, so it
  is now asserted rather than relied on.

`casefold` vs `lower` is **not** equivalent here and is pinned: `Borrowſ` folds to
`borrows` under one and not the other — one codepoint from the key that decides a grant.

## Collision with #580

#580 (`$`-anchored names, #577) also touches `charter/persona.py`. **No overlap:** it edits
`valid_name` (line ~64) and `reference_refusal` (line ~117); every line here is at 221+
(`_frontmatter` onward) plus `structural_errors`/`lint`. Semantically #580 only tightens
`reference_ok` → `load`, which is upstream of `key_issues` — it helps, it does not conflict.
Either order merges cleanly.

## Also

* The key vocabulary moved from `commands_persona` to `persona`
  (`KNOWN_KEYS`/`CHARTER_OWN_KEYS`/`AGENT_PASSTHROUGH_KEYS`, re-exported under the old
  names). That is what let `structural_errors` ask about keys at all — its docstring named
  that import as the reason it could not, and it runs on every turn for the status line.
* `contain.readable` on the key in all three report rows. A key cannot hold a line break
  (`splitlines` splits on `\r`, `\x85`, U+2028, U+2029), but U+3164 HANGUL FILLER survives
  `strip` and prints as nothing, so the row whose whole job is to say which key to edit
  said `frontmatter key ''` — #498's finding, on rows that had not been given the escape.

Closes #575
Closes #509

### DIFF
--- a/charter/commands_persona.py
+++ b/charter/commands_persona.py
@@ -599,19 +599,18 @@ def _credential_rule() -> str:
     )
 
 
-#: Frontmatter keys `_render_agent` copies straight into the generated sub-agent.
-#: Anything a charter sets that isn't here (and isn't consumed by charter itself)
-#: reaches `.claude/agents/<name>.md` never — a typo is silently inert today, which
-#: `persona lint` now reports.
-_AGENT_PASSTHROUGH_KEYS = ("model", "color", "memory")
-
-#: Keys charter reads itself rather than emitting. Together with the passthrough set
-#: this is the full vocabulary of a persona charter's frontmatter.
-_CHARTER_OWN_KEYS = (
-    "name", "role", "vault", "extends", "uses", "delegate-when", "description",
-    "agent-description", "agent-tools", "tools", "activity", "dispatch-isolation",
-    "draft", "skills", "disallowed-tools", "routing", "routes-to", "borrows",
-)
+#: Frontmatter keys `_render_agent` copies straight into the generated sub-agent, and the
+#: keys charter reads itself. Anything a charter sets that is in neither reaches
+#: `.claude/agents/<name>.md` never — a typo is silently inert, which `persona lint`
+#: reports.
+#:
+#: Both now DEFINED in `persona.py` and re-exported here under the names this module has
+#: always used. The vocabulary of a persona definition is a fact about the parser, not
+#: about the command that renders one, and `persona.structural_errors` — which runs on
+#: every turn for the status line — says in its own docstring that it cannot afford to
+#: import this module to fetch it.
+_AGENT_PASSTHROUGH_KEYS = persona.AGENT_PASSTHROUGH_KEYS
+_CHARTER_OWN_KEYS = persona.CHARTER_OWN_KEYS
 
 
 def _rel(path) -> str:
@@ -877,11 +876,35 @@ isolated context. Adopt the charter below as your role.
 
 
 def _write_agent(name: str) -> str | None:
-    """Generate/refresh one persona's sub-agent. Returns 'written'|'skipped'|'draft'|None.
-    Uses the RESOLVED persona (inheritance applied: merged charter + unioned tools)."""
+    """Generate/refresh one persona's sub-agent. Returns
+    'written'|'skipped'|'draft'|'unreadable'|None. Uses the RESOLVED persona (inheritance
+    applied: merged charter + unioned tools)."""
     d = persona.resolve(name)
     if not d:
         return None
+    if persona.key_issues(name):
+        # The sibling of #575's named bug, and the one with the wider blast radius. A key
+        # charter cannot read is not inert HERE — three of the fields this function renders
+        # are enforced by their PRESENCE, so misspelling one deletes the enforcement:
+        #
+        #   * `Agent-tools:` → no `tools:` line is emitted, and no `tools:` line means the
+        #     sub-agent inherits EVERY tool. The allowlist does not narrow; it vanishes.
+        #   * `Disallowed-tools:` → no `disallowedTools:` line. The denylist vanishes.
+        #   * `Draft:` → `is_draft` is False, so the unfinished charter this function
+        #     refuses to ship becomes a sub-agent's system prompt after all.
+        #
+        # All three were reached by `charter persona sync-agents` printing `✓ Synced 1
+        # persona sub-agent(s)`. Same answer as a draft, for the same reason the draft
+        # branch gives: the generated file IS the sub-agent's system prompt, and one built
+        # from a definition charter could not read is not the committed charter. The stale
+        # agent goes too — leaving it keeps the persona dispatchable under whatever it said
+        # before, which is exactly the grant the author was editing the file to remove.
+        #
+        # Deliberately NOT gated on lint's other findings, or on an unknown key generally:
+        # `key_issues` is the narrow set charter can prove was meant to be read (a key it
+        # reads, miscased or written twice), so `modell:` still renders an agent.
+        _remove_agent(name)
+        return "unreadable"
     if persona.is_draft(name):
         # The generated file IS the sub-agent's system prompt, so an unfinished charter
         # must not become one. Any agent generated BEFORE the persona was marked draft is
@@ -1646,6 +1669,7 @@ def cmd_persona_sync_agents(args) -> int:
     outcomes = {n: _write_agent(n) for n in names}
     written = [n for n, o in outcomes.items() if o == "written"]
     drafts = [n for n, o in outcomes.items() if o == "draft"]
+    unreadable = [n for n, o in outcomes.items() if o == "unreadable"]
     withheld = {n: persona.mcp_withheld(n) for n in written}
     withheld = {n: v for n, v in withheld.items() if v}
     # A server name the committed sidecar chose and `mcp_name_ok` refused. Said on the run
@@ -1669,6 +1693,21 @@ def cmd_persona_sync_agents(args) -> int:
         util.warn(f"Skipped {len(drafts)} draft persona(s): {', '.join(drafts)} — "
                   "an unfinished charter must not become a sub-agent's system prompt. "
                   "Finish it, drop the `draft: true` line, then re-run.")
+    if unreadable:
+        # Said on the run that would have written the agent, not left to `lint`. A
+        # miscased `Agent-tools:` does not narrow a sub-agent's tools, it removes the
+        # allowlist entirely — so a green tick over this is the shape #453 keeps arriving
+        # in, a bound that degrades in silence.
+        util.warn(f"Skipped {len(unreadable)} persona(s) whose frontmatter charter cannot "
+                  f"read: {', '.join(unreadable)}. A key spelled in another case, or "
+                  f"declared twice, is read by nothing — and `agent-tools:`, "
+                  f"`disallowed-tools:` and `draft:` are enforced by being PRESENT, so a "
+                  f"misspelling drops the allowlist, the denylist or the draft guard "
+                  f"rather than narrowing anything. No agent is generated until the key is "
+                  f"fixed:")
+        for n in unreadable:
+            for _lvl, msg in persona.key_issues(n):
+                util.info(f"  {contain.readable(n)}: {msg}")
     if removed:
         util.info("Removed stale generated agents: " + ", ".join(removed))
     if refused:
--- a/charter/persona.py
+++ b/charter/persona.py
@@ -258,9 +258,9 @@ def parse(text: str) -> tuple[dict, str]:
 
     **Lossy, and now provably so.** Two lines carrying one key collapse to the last of
     them, which is a fact about a `dict` and not a decision anyone made: the file said two
-    things and this says one. Callers that need to know a key was written twice ask
-    :func:`duplicate_keys` — :func:`load` does, so every persona-shaped consumer gets the
-    answer without a second read.
+    things and this says one. A caller holding only text asks :func:`duplicate_keys` for
+    what the dict dropped; :func:`load` reads both off one walk of the same file, so every
+    persona-shaped consumer gets the answer without a second parse.
     """
     pairs, body = _frontmatter(text)
     return dict(pairs), body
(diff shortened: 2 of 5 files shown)


diazoxide:
## Verification record — head sha `e2d4e1d`

### Full suite, two local environments

| env | python | result |
|---|---|---|
| 1 | 3.14.4 | `Ran 6543 tests` — **OK**, 0 failures |
| 2 | 3.12.13, with `CHARTER_*` / `CLAUDE_*` / `ANTHROPIC_*` / `TMUX*` unset | `Ran 6543 tests` — **OK**, 0 failures |

Env 2 scrubbed 15 names: `CHARTER_HARNESS`, `CHARTER_PERSONA`, `CHARTER_ROOT`,
`CHARTER_SESSION_ID`, `CHARTER_WORKSPACE`, `TMUX`, `TMUX_PANE`, `CLAUDE_CODE_ENTRYPOINT`,
`CLAUDE_CODE_MESSAGING_SOCKET`, `CLAUDE_CODE_MESSAGING_TOKEN`, `CLAUDE_CODE_EXECPATH`,
`CLAUDECODE`, `CLAUDE_CODE_SESSION_ID`, `CLAUDE_CODE_CHILD_SESSION`, `CLAUDE_PID`.

### CI at the head sha

`gh api repos/diazoxide/charter/commits/e2d4e1d26c0e0a7b94259d020e1e68df7f3b887d/check-runs`
(not `gh pr checks`, per #561):

```
test (3.11): success
test (3.12): success
test (3.13): success
test (3.14): success
A dev-channel install from main actually installs: skipped
```

### Baseline-diffed test verification

All 45 new tests, run against unfixed `bf58b46`: **37 fail** (23 failures + 14 errors).
The 8 that pass are exactly the no-change pins, and are supposed to pass on both sides:

```
test_an_absent_borrows_still_means_the_legacy_grant
test_a_correctly_spelled_borrows_none_still_borrows_nothing
test_a_correctly_spelled_borrows_still_grants_what_it_names
test_a_miscased_key_that_is_not_borrows_leaves_the_grant_alone
test_an_unknown_key_stays_a_warning
test_an_unknown_key_still_generates_an_agent
test_parse_still_returns_the_dict_and_the_body
test_the_value_is_never_read_out_of_the_miscased_key
```

### Hand-check — 30 mutations, 29 caught, 1 equivalent

Reported separately from the sweep because `tools/sweep.py` has no operator for string
constants or semantic swaps (#569) — which is exactly the shape of a key name and a case
comparison. Every round ran against an **unmutated baseline** (`rc=0, failures=0`) and a
mutation counted as caught only when it added a failure the baseline did not already have
(#579).

Round 1 — 26 mutations, 23 caught. Round 2 re-ran the 3 survivors after pinning, plus 4 new
mutations of the `extends:` fix — 6/7 caught, the 7th equivalent.

Survivor analysis (none was dismissed as equivalent without checking):

| survivor | verdict |
|---|---|
| `contain.readable` dropped from the **miscased** row | **real** — the only unprintable-key fixture was an *unknown* key, which lands on a different row. A case variant can be non-ASCII: `Borrowſ` casefolds to `borrows` (U+017F). Pinned; now caught. |
| gate moved after the draft check | **real** — no fixture was both a draft and miscased. The agent is withheld either way, so this decides which sentence the operator gets. Pinned in that order; now caught. |
| `_FOLDED_KEYS` built unfolded | **equivalent**, and only because every vocabulary key is lowercase. That invariant is what makes "miscased" mean anything, so `test_the_vocabulary_is_lowercase` now asserts it rather than relying on it. |

`casefold` vs `lower` is **not** equivalent and is pinned: `Borrowſ` folds to `borrows`
under one and not the other — one codepoint away from the key that decides a tool grant.

### `tools/sweep.py`

Running on this branch (`--ref HEAD --base bf58b46`, workdir under `/private/tmp/...`).
Selection map built (10664 source files, 92s); it is in its own unmutated baseline now.
The host is suspending repeatedly, so wall-clock is far ahead of CPU time — I will post the
charge/survivor table as a follow-up comment rather than hold the PR on it.

diazoxide:
### `tools/sweep.py` — charge, and where it got to

```
sweeping e2d4e1d26c0e (paths: charter)
  diff against bf58b4640c77: 3 file(s), 344 added line(s)
  selection map: cached (selection-e64bb79eb9cdbc6b.json)
  23 mutations across 3 file(s)
    [1/23] pinned    charter/persona.py:282 [drop-comprehension-if] n > 1   (1 module(s))
```

**It is still running.** The host this ran on is suspending almost continuously — the sweep
process has accumulated 0.62s of CPU across more than an hour of wall-clock, and two earlier
attempts were killed mid-flight by the same thing. This is host availability, not the
branch.

The run was restarted with `--no-baseline`, and only because the condition that flag
protects is independently established here: the **full suite is green at this exact sha in
two environments** (3.14.4 and 3.12.13-scrubbed, 6543 tests, 0 failures each), plus four
green CI legs. A red baseline makes every mutation look pinned; there is no red baseline to
hide behind. The first attempt did run the sweep's own unmutated baseline and was killed
inside it, having reported nothing red.

I will post the full survivor table when the run lands. **The part the sweep could never
have covered is already done and reported above**: it has no operator for string constants
or semantic swaps (#569), which is exactly the shape of a key name and a case comparison —
so the 30-mutation hand-check, baseline-diffed, is the one that speaks to this change's
actual risk surface, and it is complete at 29 caught / 1 equivalent.

diazoxide:
### `tools/sweep.py` — closing the loop on the earlier note

Final state: **23 mutations charged across 3 files, 2 pinned, run never completed.**

```
sweeping e2d4e1d26c0e (paths: charter)
  diff against bf58b4640c77: 3 file(s), 344 added line(s)
  23 mutations across 3 file(s)
    [1/23] pinned  charter/persona.py:282 [drop-comprehension-if] n > 1              (1 module)
    [2/23] pinned  charter/persona.py:332 [drop-if] if key in KNOWN_KEYS: return None (46 modules)
```

Five attempts, all killed or starved by the same thing: the host suspended almost
continuously — its clock advanced **16 seconds across a ten-minute wall-clock wait**, and
the sweep accumulated under a second of CPU per hour. Nothing about the branch; I could not
get the machine to stay awake long enough. Not re-run against `main` post-merge for the same
reason.

**What that leaves uncovered, stated plainly.** The sweep's deletion operators would charge
the added lines in `persona.py`/`commands_persona.py`. Those lines are covered by the
45 new tests (37 of which fail on the unfixed base) and by the 30-mutation hand-check at
29 caught / 1 equivalent — and the hand-check is the one that matters most here, because the
sweep has **no operator for string constants or semantic swaps** (#569), which is exactly the
shape of a key name and a case comparison. What the sweep would have added over that is
deletion coverage of the reporting/glue lines, and that is the gap.

Worth someone re-running on a machine that stays awake; the workdir cache
(`selection-e64bb79eb9cdbc6b.json`) makes a restart cheap.

Issue #582, opened 2026-08-27
Frame state files are written at the umask, and the scan cannot see them because `frame_dir` returns a local


diazoxide:
Two findings, and the second is why the first was never reported.

## 1. Every frame state file is written at `0o777 & ~umask`

`charter/frame/state.py` writes ten files — `version`, `exit`, `harness`, `session`,
`server`, `workspace`, `density`, `hidden`, `identity`, `panes` — through the same shape:

```python
    d = frame_dir(fid, create=True)
    tmp = d / "version.tmp"
    tmp.write_text(f"{time.time_ns()}\n")
    os.replace(tmp, d / "version")
```

`os.replace` carries the **source's** mode onto the destination, and `tmp.write_text`
creates at the umask. `charter/frame/gather.py` has the same shape for its cache file.

Measured, `umask 022`:

```
d 700  .charter/frame/<fid>              <- private_mkdir, correct
f 644  .charter/frame/<fid>/version
```

Charter-created directories make this defence-in-depth today: the file is 0644 inside a
0700 directory. But `frame_dir` calls `config.private_mkdir`, which by design **leaves a
directory it did not create exactly as it is** (#331), so a `.charter/frame/` or
`.charter/frame/<fid>/` restored from a tarball, made by hand, or carried over from a
charter older than 0.53 is 0755 — and then the file is readable by every account on the
machine. That is exactly the case #505 fixed for the rest of the state directory:

```
d 755  .charter/frame/<fid>              <- pre-existing, charter will not chmod it
f 644  .charter/frame/<fid>/version      <- readable by anyone
```

Frame state carries the workspace name, the harness, the tmux server socket, the session
id and the pane layout.

## 2. The coverage scan cannot see any of it, and the naive fix makes it useless

`tests/_statedirscan.py` follows a module-level function whose **return expression**
mentions a state name. `frame.state.frame_dir` does not:

```python
def frame_dir(fid, create=False):
    d = contain.child(_root(), fid)
    ...
    return d                 # a local — `_returns` reads the expression, not the binding
```

so `frame_dir` is not a state-path function, and neither the `mkdir` half (#470) nor the
file half (#505) sees anything in `frame/`. The `mkdir` half happens to be clean anyway —
`frame_dir` calls `config.private_mkdir` — so nothing has ever failed because of this.

The obvious fix is not available: substituting local assignments into return expressions
was measured on this tree and over-taints catastrophically. Every function returning a local
[...]

Issue #505, opened 2026-08-24
Files under .charter/ are written at the umask, so a state directory charter did not create leaks every one of them


diazoxide:
`config.private_mkdir` / `config.mkdir_for` make every directory charter creates under
`.charter/` come out 0700, and #470 closed the last writer that did not go through them
(the ephemeral persona store, whose path reaches `memstore.write` as a parameter — PR #493).

The **files** inside `.charter/` are a different question and are still written at
`0o777 & ~umask`. That is harmless as long as the directory above them is 0700 — which is
exactly the case charter deliberately does **not** guarantee. `private_mkdir` leaves a
directory it did not create exactly as it is (#331: `$CHARTER_HOME` can point the state
directory anywhere, so "chmod whatever we land in" is how charter comes to tighten a home
or a shared team directory unprompted), and `vault list` / `doctor` report the loose one
rather than fixing it. So on a plane whose `.charter/` predates charter — or was made by
`mkdir -p` at the umask default before 0.52.x — every account on the machine can read
charter's state.

## Reproduction

Measured on the #470 branch, so this is what ships after that fix, not before it:

```sh
cd /tmp && mkdir plane && cd plane
python3 -m charter init --forge github --owner acme --no-front-door
mkdir -p personas/steward && printf -- '---\nname: steward\n---\n\n# Steward\n' \
  > personas/steward/persona.md

rm -rf .charter
mkdir .charter && chmod 755 .charter      # the pre-existing loose one charter reports

umask 022
python3 -m charter vault add devops --provider plain-file
python3 -m charter persona remember steward "a scratch note" --ephemeral
echo '{"session_id":"s1","cwd":".","tool_name":"Bash","tool_input":{"command":"ls"}}' \
  | python3 -m charter hook pretooluse

find .charter -type f -exec stat -f "%Sp %N" {} \;
```

```
-rw-------  .charter/vaults.json
-rw-r--r--  .charter/guard-seen.json
-rw-r--r--  .charter/persona-state/trace/<session>.jsonl
-rw-r--r--  .charter/persona-state/ephemeral/<session>/steward/a-scratch-note.md
```

`vaults.json` is 0600 because the registry chmods it outright. Nothing else does.
`doctor` correctly reports `listed by other accounts: .charter 755 (want 700 — chmod 700)`
the whole time — it is telling the operator about a directory whose contents charter went
on writing world-readable.

## What is exposed

- **the trace log** — which persona did what, when, with what title. `charter persona
[...]

PR #603, opened 2026-08-28
A state file's mode comes from charter, and the scan can see a path a function returns (#581, #582)

### LINKED ISSUE #581: `persona.set_active` writes three state pointers at the umask — the last writers #505 did not route
#505 routed every file writer under `.charter/` through `config.write_for` / `open_for` /
`touch_for`, which settle the mode on the descriptor before content lands. Three sites in
`charter/persona.py` were left out of that PR — not because they are different, but because
two other branches were live in that file at the time and a three-line edit in the middle of
them buys a conflict for no schedule. They are listed in
`EveryStateWriterGoesThroughTheWalk.NOT_ROUTED_YET`, so the coverage scan reports them
rather than passing over them.

`persona.set_active`:

```python
    for f in (sf, tf):
        if f is not None:
            config.private_mkdir(f.parent)
            f.write_text((name or "") + "\n")        # <- the two pointer files
    if sf is None and tf is None:
        config.ACTIVE_PERSONA_FILE.write_text((name or "") + "\n")   # <- the plane-wide one
```

## Measured

Plane with `.charter/` pre-existing at 0755 (which charter will not chmod — #331), `umask 022`:

```
d 700  .charter/persona-state
d 700  .charter/persona-state/trace
f 600  .charter/persona-state/trace/<session>.jsonl      <- routed by #505
d 700  .charter/sessions
f 644  .charter/sessions/<session>.persona               <- this issue
d 700  .charter/terminals
f 644  .charter/terminals/<tid>.persona                  <- this issue
```

Under `umask 000` they are 0666, so another account can *rewrite* which persona a session
is on.

## What is exposed

A persona **name**, which is also the name of a committed directory under `personas/` — so
the read side leaks nothing new. The write side is the sharp half: `.charter/sessions/` and
`.charter/terminals/` decide which charter a session runs as, and a 0666 pointer is one
another account gets to set.

## The fix

Three one-line swaps to `config.write_for(f, …)`. No new mechanism; #505 shipped the
dispatch and the scan that proves the routing.

## The ratchet

`tests/test_the_state_directory_is_charters_to_choose.py`'s
`test_no_write_in_the_package_can_leave_a_loose_state_file` asserts **equality** against
`NOT_ROUTED_YET`, so routing these fails the test as a stale entry — deleting the entry is
part of the fix, and the list cannot rot into an allowance.

diazoxide:
Closes #581, #582.

One property, two issues: **a state file's mode comes from charter, not from the umask** — and the scan that proves it can now see a path a function *returns*.

## The three writers that were listed (#581)

`persona.set_active` writes `.charter/sessions/<sid>.persona`, `.charter/terminals/<tid>.persona` and — with neither id to key on — the plane-wide `active-persona`. They were out of #505 only because two other branches were live in `charter/persona.py` at the time, so they sat in a `NOT_ROUTED_YET` ratchet asserting **equality**: routing them fails as a stale entry, which is what happened, and deleting the entry is the other half of the fix.

Measured, `.charter/` pre-existing at 0755 (which charter will not chmod — #331):

```
f 644  .charter/sessions/<session>.persona      umask 022
f 666  .charter/sessions/<session>.persona      umask 000
```

The read side leaks little — a persona name is also a committed directory name. The write side is the sharp half: those directories decide which charter a session runs as.

**The ratchet is now empty and gone.** With nothing left to list, the file half reads exactly like the directory half, which is what it should always have looked like — one property asked about two kinds of inode. A list that can only ever be empty, plus a test that iterated it to check its entries still named real files, is two things with no remaining consequence.

## The thirteen nobody could see (#582)

Ten atomic writers in `frame/state.py`, the respawn counter, `gather`'s cache, and **two more the issue did not enumerate** — `commands_frame`'s `tmux.conf`, which carries the session id, the hotkey and the toggles, and which a 0666 file lets another account choose.

A temp file plus `os.replace` is not a way out of this; it is why it stayed quiet. `os.replace` carries the *source's* mode onto the destination.

## The scan half, measured three times

The issue said the naive fix is worse. It is, and here are this tree's numbers rather than the older ones:

| rule | state-path functions | false positives |
|---|---|---|
| today | 41 | none |
| substitute every local | 189 | 3 mkdir + 5 write, on **committed** paths |
| substitute state-mentioning locals | 138 | the same 8 |
| substitute call-bound locals into path returns | 76 | the same 8 |
| **shipped: decided by the return's shape** | **43** | **none** |

The over-taint is not a matter of degree — it follows the *data*: a workspace name read out of a state pointer flows into `workspace.ensure`, and `workspaces/<name>/` gets reported as charter's own state. Clearing that requires routing committed directories through the private walk, i.e. #331.

So the question is no longer *does this return mention a state path* but **is this return a path at all**:

* a path being built (`a / b`, or `contain.child` / `Path` / `joinpath` / `with_name` / `with_suffix`) may substitute any call-bound local — `/` is path arithmetic, and a value read out of a file is never its left operand;
* a bare local handed back may substitute only a local a constructor built — `return d` says nothing about what `d` is, so the call that made it stands in for a type;
* `read_text()`, `json.loads`, an f-string, a comparison: not followed.

**+2 functions (`frame.state.frame_dir`, `frame.gather._cache_file`), none lost, zero false positives.** With the writers put back as they were, the scan names all thirteen.

The named half also asks the **cross-module** question now, which is what `gather.save` needed: its path comes from `gather._cache_file`, whose state-ness comes from `frame.state.frame_dir` one import away. Both halves get the same package-wide answer, and `write_violations` is now literally `violations` with different sites rather than a second copy of the same eight lines.

## Baselines

Every new case run against the unmutated tree first. The persona pointers came back 0666/0644 exactly as #581 measured; the frame files the same; the scan-accuracy cases failed 7 + 1 error.

Suite: 7033 tests, OK.


diazoxide:
### The red `test (3.11)` was a void death, not this branch — filed as #609

`test_nothing_of_the_operators_tmux_is_written_by_a_whole_launch` came back `[1] != [21]`. `1` is the exact value `tests/test_frame_tmux_integration.py`'s own header says a void death produces here:

> **One test here still cannot carry a probe, and says so rather than pretending** … There is no free index: `[1]` is where a probe would go and is charter's teardown … **A void death there is still read as an exit code of `1`.**

Every other death-dependent test in that module detects the #487 window (a loaded tmux 3.4 closing a pane's fd before it has the child's status) and **skips**, naming what it could not measure. This one has no free `pane-died` index, so it asserts instead.

**Evidence it is not the diff:**

| | |
|---|---|
| Same sha, two concurrent workflow runs | `33148747699` 3.11 **red**, `33148750693` 3.11 **green** — started 06:40:53 and 06:40:55, same image, same code. The test **ran** in both, not skipped. |
| The previous head `eaf68bc` | 3.11 **green**, with every production change already in it: all eleven `frame/state.py` writers, `gather`'s cache, both `commands_frame` `tmux.conf` writes, `persona.set_active`. |
| `eaf68bc` → `6b80895` delta | `tests/_statedirscan.py` and one test module. Nothing a tmux integration test executes. |
| The code path | This test drives `_launch_in_operator_tmux` (1707–1930), which contains **no** `write_text` and no `write_for`. The changed writes are at 2760/2792 inside `cmd_launch`'s own-server branch, which the operator path returns before reaching (line 2670). |

The failed job has been re-run. **All four minors are now green at `6b808951`, in both workflow runs** — three independent 3.11 results at this sha, two green and the one void death.

The residual — that this test can report a red about the runner where its siblings skip — is #609, with two shapes worth measuring written down there.

PR #604, opened 2026-08-28
The surface is one keystroke, the floor was measured, and a component can match without charter drawing over it (spec Phases 3 and 4)

diazoxide:
Phases 3 and 4 of `docs/superpowers/specs/2026-08-28-frame-visual-design.md`. Phases 1–2 shipped in #599, which also landed most of Phase 3's config surface; what was left was the measurement the spec could not do, the palette rows, and the whole of Phase 4.

## Phase 3.5 — the one the spec could not finish

§9 recorded that every `window-style` measurement behind the pane surface was run on tmux 3.7c and on nothing else, while `tmuxctl.FLOOR` is 3.2. **"Believed safe" is not measured.** So tmux 3.2 was built from source and the whole of §1 and §4 was re-run against it — one style per server, with the SGR lifted out of an attached client's wire rather than read out of surrounding bytes.

| at tmux 3.2 | answer | 3.7c |
|---|---|---|
| `set -p window-style` pane-scoped? | **yes** — sibling panel, harness pane and `show -w` all read back `''` | same |
| honours colour only? | **yes** — `reverse`, `dim`, `bold` each put *no SGR at all* on the wire; `bg=colour236,dim` put `ESC[48;5;236m` and nothing else | same |
| do the 16 ANSI names resolve? | **all sixteen** — `bg=black`→`ESC[40m` … `bg=white`→`ESC[47m`, `bg=brightblack`→`ESC[100m` … `bg=brightwhite`→`ESC[107m` | same |
| active/inactive split | active `ESC[100m`, inactive `ESC[40m`, harness `ESC(B ESC[m` | same |
| style value format-expanded | stored verbatim | same |
| `#(...)` in a style | refused by the parser | same |
| refused `set-option` | rc 1, `invalid style:`, previous value intact — reported, not fatal | same |
| survives `respawn-pane` / `resize-window` | yes | same |
| per-client downsample | 256 → `ESC[48;5;236m`; 8-colour `xterm` → `ESC[40m`; `vt100` → `ESC[7m` | same |
| `set -p -u` removes it | rc 0, reads back `''`; unsetting one never set is rc 0 too | same |
| live `set -p` on an attached pane | repaints by itself, no `refresh-client` | same |

**66 of 68 answers byte-identical.** The two that were not were the measuring harness, not tmux: it batched four styled panes into one session, and tmux emits an SGR only when the style *changes* — so four panes that downsample to the same colour produce one escape and three silences, which reads as "three styles tmux ignored" and is nothing of the sort. Re-run one pane per server they matched too. That correction is in the harness's own docstring so nobody repeats it.

**So `chrome` carries no version gate**, and the absence is asserted behaviourally (`tmuxctl.version` raises; both builders must still answer) rather than left to be noticed. The spec's contingency — gate at the version that works, the way `display-popup` is gated at 3.3 — did not come due.

## Phase 3.6 — `F2` carries the surface

Three rows, `chrome: off` / `dark` / `light`, the one in effect marked. `charter frame-chrome <level>` records the word in the frame's own state directory and sets two pane options; **nothing splits**, because a live `set -p window-style` repaints the pane by itself on both versions.

`off` is a **removal**, not a third style. On a launch it is free (nothing is set); on a running frame the same word has to unset, or the row that says `off` reports success and leaves the surface where it was. *Which* options to unset comes from `instance.chrome_option_names`, derived from `FRAME_CHROME` rather than spelled beside it.

`_split_panels` now reads `_current_chrome` rather than `config.FRAME`, closing a reachable defect: change the surface, then change the density, and the pane the re-layout added used to come up bare beside three surfaced ones.

## Phase 4 — the provider seam

`ctx.chrome` — a `MappingProxyType` of `heading`, `muted`, `selected`, `ok`, `warn`, `bad`, `reset`, `inset`, built from `statusline`'s own constants so a provider's heading and charter's cannot be two different weights.

Served with the geometry, not through `SERVES`: every `SERVES` value is a function *of the snapshot*, and an entry that took the snapshot and ignored it is the exact shape `Ctx`'s docstring records as the defect that let an action reach the plane's whole vault inventory. `ctx.GEOMETRY` is therefore now `ctx.ALWAYS`.

Two deliberate absences, each with its reason in the code:

- **No `surface`/`focus` recipe.** Those are tmux pane options; no renderer can write one, and an empty string for them would claim to hand over something charter does not have.
- **No `chrome` argument on `recipes()`.** §7 asks for the roles "resolved for this frame's `chrome` setting" — and the measurement says there is nothing to resolve: `window-style` ignores every attribute, so nothing a renderer can write changes with the word. An argument that cannot change an answer is a line this repo deletes. What *is* resolved is `colour_ok`, which is per pane and changes every value.

Under `NO_COLOR` or a non-tty every escape is the empty string and **the keys stay**, so a component written against them does not raise inside its own draw.

The exact-attribute-set test on `Ctx` is updated deliberately and visibly per §4.2, spelled as a list rather than read off the constant. `builtin_actions.build` takes `current_chrome` as a **required** keyword for the same reason — a default would let a caller that forgot it mark `off` on a surfaced frame and be right about it in every test that did not set one. That cost six lines across five existing test files.

## Verification

- **Hand-check, reported separately from the sweep** (#569 has no operator for a string/regex constant or a semantic swap — a colour name, an escape sequence and a tmux option word are all that shape): unmutated baseline GREEN first, every mutation asserted on disk before a suite was spent on it, `__pycache__` cleared between runs. **16 applied, 16 pinned, 0 survived** — including §4.5's two by name (hand a provider a mutable recipes dict; let a foreign row skip `_fit`).
- Three of this branch's *own* tests were then mutation-checked and found to assert less than they read like — one asserted a mock had raised rather than that charter degraded. Fixed in a second commit, each with the mutation that now goes RED against it.
- Deletion sweep, full suite in two environments, and CI at the head sha: results posted below.

Not merged.

### DIFF
--- a/charter/cli.py
+++ b/charter/cli.py
@@ -722,7 +722,7 @@ def _add_frame_parsers(sub) -> None:
     _core_commands = set(sub.choices) | {"frame", "panel", "frame-palette",
                                          "frame-probe", "frame-respawn", "frame-density",
                                          "frame-resize", "frame-gather", "frame-switch",
-                                         "frame-toggle"}
+                                         "frame-toggle", "frame-chrome"}
 
     # Which harness (by `.name`, never `.cli_name` — that's the dict key below) has
     # already claimed each word, so a SECOND harness wanting it is told who got there
@@ -878,6 +878,23 @@ def _add_frame_parsers(sub) -> None:
     dn = sub.add_parser("frame-density")
     dn.add_argument("level")
     dn.set_defaults(func=commands_frame.cmd_density)
+
+    # Internal, and a top-level sibling for the same `_split_frame_argv` reason as the
+    # ones above. Started DETACHED by a palette row, whose argv is exactly
+    # `util.self_relaunch_argv("frame-chrome", <level>)`. It repaints the RUNNING frame's
+    # panel panes — never charter.toml, for `frame-density`'s reason — and resolves which
+    # frame from `$CHARTER_SESSION_ID` when it fires rather than from a shared bind.
+    #
+    # Deliberately NOT `choices=` on the level, for `frame-density`'s argument and one
+    # more that is specific to this command: `instance.chrome_level` is the one gate on
+    # the closed set, and it is also the boundary that keeps an operator's string away
+    # from a tmux style value. A second copy of the set in argparse would be a second
+    # answer to that question, and the weaker one — argparse exits 2 from inside a
+    # detached child where nothing prints the reason, while `cmd_chrome` is a quiet no-op
+    # that leaves the frame exactly as it was.
+    cr = sub.add_parser("frame-chrome")
+    cr.add_argument("level")
+    cr.set_defaults(func=commands_frame.cmd_chrome)
 
     # Internal, and a top-level sibling for the same `_split_frame_argv` reason as the
     # ones above. Fired by ONE COMPONENT's own `bind -n` — the `key` its
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -1671,6 +1671,47 @@ def _surface_argvs(*, socket: str, pane_id: str, chrome) -> list[list[str]]:
             for name, value in instance.chrome_options(chrome)]
 
 
+def _resurface_argvs(*, socket: str, pane_id: str, chrome) -> list[list[str]]:
+    """:func:`_surface_argvs` for a pane that is ALREADY DRAWN — with the unsets.
+
+    **The difference is `off`, and it is not a detail.** On a launch, `off` is free:
+    nothing is set, so `_surface_argvs` correctly issues no command. On a running frame
+    the same word has to mean *remove what is there*, or the palette row that says
+    ``chrome: off`` leaves the operator looking at the surface they just turned off — a
+    keypress that reports success and changes nothing, which is the shape this spec's own
+    §4 refuses an ``auto`` value for.
+
+    So: every option the new word does NOT set is unset, and every option it does set is
+    set. Which options exist comes from `instance.chrome_option_names`, derived from the
+    table rather than spelled here, so a third option added to `FRAME_CHROME` is unset by
+    this function on the day it is added instead of surviving an `off` nobody re-read.
+
+    `set-option -p -u` is the removal, and it is measured rather than assumed on both
+    tmux 3.7c and tmux 3.2 (`tmuxctl.FLOOR`): rc 0, `show -p` reads back `''` afterwards,
+    and unsetting an option that was never set is rc 0 as well — so this needs no "was it
+    set" round trip, which would be a subprocess per pane per keypress to answer a
+    question the unset already answers.
+
+    **Never the harness pane**, exactly as `_surface_argvs` is never handed one: the
+    caller (`cmd_chrome`) iterates `state.panes`, which is the PANEL map, and the harness
+    pane lives in a different record entirely (`state.harness_pane`). ADR 0018's boundary
+    holds here by the same construction.
+
+    **`NO_COLOR` wants NOTHING SET, which on this path means the unsets and not silence.**
+    `_surface_argvs` answers `[]` there because a launch under `NO_COLOR` has nothing to
+    remove. A frame surfaced before `NO_COLOR` was exported has, and answering `[]` here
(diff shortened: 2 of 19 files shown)


diazoxide:
## Verification, at head `761c403`

**Deletion sweep** — `python3 tools/sweep.py`, scoped to this branch's diff against its merge-base:

```
16 mutation(s) offered by this diff
16 pinned · 0 survived        wall clock 8.9 min
Every mutation this diff offered goes red. Nothing added here is a line the
suite would not miss.
```

The first run was **not** clean: it found **4 unpinned guards**, all in the two new state accessors — `record_chrome`'s `if d is None` and `except OSError`, and `chrome`'s `if d is None` and its `or None` fallback. Deleting each left all 7037 tests green. None is decoration: `$CHARTER_SESSION_ID` reaches both functions without passing through `state.frame_id`'s minting (`cmd_chrome` reads the variable directly), so a hostile or oversized id arrives as an ordinary argument and `frame_dir` answers `None` — without the guards that is a `TypeError` raised out of a command running detached from a palette row, where nothing prints it. All four are now pinned, each test naming *which* refusal fired and carrying a live control proving the fixture writes when the write works.

The fourth is worth stating plainly rather than quietly pinning: nothing downstream can currently tell `None` from `""`, because `chrome_level("")` is `None` either way. The line exists for the accessor's *documented contract* — "`None` for never set" — so it is pinned rather than deleted, and the test says so.

**Hand-check, separate from the sweep.** #569 records that the sweep has no operator for a string/regex constant or a semantic swap, and a colour name, an escape sequence and a tmux option word are exactly that shape. So those were run by hand: unmutated baseline GREEN first (#579), every mutation asserted present on disk before a suite was spent on it (#585/#586), `__pycache__` cleared between runs.

**16 applied, 16 pinned, 0 survived**, including §4.5's two by name:

| mutation | verdict |
|---|---|
| `off` stops removing the surface and merely sets nothing | RED (4 failures) |
| the unset list is spelled beside the table instead of derived from it | RED |
| the recorded word reaches `chrome_options` ungated | RED |
| a pane id off disk goes straight into a tmux argv | RED |
| a frame with no recorded panes records the word anyway | RED |
| a split reads the configured word instead of the live one | RED |
| `NO_COLOR` stops reaching the live surface path | RED |
| **the recipes are handed over as a mutable dict** | RED |
| `NO_COLOR` stops reaching the recipes | RED |
| a recipe names a colour charter picked out of the 256 | RED |
| the heading recipe drifts from charter's own heading weight | RED |
| the inset recipe stops being the column the frame starts at | RED |
| the recipes stop reaching the ctx | RED |
| **a foreign row skips `_fit`'s containment** | RED |
| the clear-screen loses its reset | RED |
| the palette row asks the harness-pane question, not the pane-map one | RED |

A further pass mutated the *tests*, and found three of this branch's own asserting less than they read like — one patched `state.chrome` to raise and then asserted that it raised, which is an assertion about the mock and would have stayed green with `except (OSError, ValueError)` deleted. Fixed, each with the mutation that now goes RED against it (second commit).

**Full suite, two environments, both at `761c403`** — run in six module batches because this box was carrying three other agents' suites:

| | python | environment | result |
|---|---|---|---|
| env 1 | 3.14.4 | as-is | **7041 tests, OK** |
| env 2 | 3.12.13 | `CHARTER_*` / `CLAUDE_*` / `ANTHROPIC_*` / `TMUX*` unset (16 variables) | **7041 tests, OK** |

They agree.

**CI at the head sha under review**, read with `gh api repos/diazoxide/charter/commits/761c403.../check-runs` rather than `gh pr checks` (#561), so "green" cannot be confused with "no run was ever created":

```
total_count=5
test (3.11): completed/success
test (3.12): completed/success
test (3.13): completed/success
test (3.14): completed/success
A dev-channel install from main actually installs: completed/skipped
```

**tmux servers leaked: none.** Every server the measurements started was killed *and* its socket unlinked in one cleanup — `kill-server` returns 0 with the socket still bound, so the unlink is not optional, and it goes after the kill or the kill is pointed at a path with no server on it. The only `charter-*` sockets left on the machine belong to live processes that are not mine. #590 took this box from 14 leaked servers to 0 and it is still 0.

No version bump, no stamp, no tag. News entry is `version: unreleased`. Not merged.

PR #605, opened 2026-08-28
A widget framework fits in a pane charter owns, and does not fit in the component contract (experiment)

diazoxide:
**An experiment, not a feature. Do not merge to ship it — merge it, or close it, for what it
found.**

One charter frame component — the repo table — drawn with
[Textual](https://github.com/textualize/textual) and shipped as a third-party provider
through the existing `charter.components` entry point. The question was whether the provider
model already permits a rich widget framework **without charter taking the dependency**.

`charter/` is untouched on this branch and `pyproject.toml` still says `dependencies = []`;
`tests/test_packaging.py::test_runtime_has_zero_dependencies` still passes. The one line
changed under `tests/` is the classification a new top-level directory is *required* to make
(§ below) — no production line changed, so there is no news entry.

Full writeup, with every measurement and the script that produced it:
`docs/superpowers/specs/2026-08-28-textual-frame-component-experiment.md`.

---

## Show it — real `capture-pane -p -e`

tmux 3.7c, 150x44 window, session `mouse off` (charter's default), panes split in charter's
own order. `alternate_on` is the whole architectural difference between the two components,
sitting side by side in one frame. Produced by `measure/m7_shot.sh`.

```
┌─ %0  rows 0-3  150x4  alternate_on=0  the harness (stand-in)
│ The default interactive shell is now zsh.
│ To update your account to use zsh, please run `chsh -s /bin/zsh`.
│ For more details, please visit https://support.apple.com/kb/HT208050.
│ bash-3.2$
┌─ %5  rows 5-16  150x12  alternate_on=1  panel textual.live
│  repos 5  ·  harness-wrapper
│  repo                 branch                      ci    change
│  ▸ charter ⑂3         main               * ↑2     fail  !554
│    easydmarc-app      feat/dmarc-rollup  ↓4       ok
│    infra              main                        run   #88
│    charter-docs       docs/frame         ? ↑1 ↓1  ok
│    statusline-lab ⑂2  main
│    ╰ pr-554                              ? ↑1     …
│    ╰ pr-560                                       ok    !560
│ 
│ 
│  snapshot 5s old  ·  clicks 0  ·  scroll 0  ·  keys 0  ·  live · frozen
┌─ %4  rows 18-29  150x12  alternate_on=0  panel textual.repos
│  repos 5  ·  harness-wrapper
│  repo                 branch                      ci    change
│  ▸ charter ⑂3         main               * ↑2     fail  !554
│    easydmarc-app      feat/dmarc-rollup  ↓4       ok
│    infra              main                        run   #88
│    charter-docs       docs/frame         ? ↑1 ↓1  ok
│    statusline-lab ⑂2  main
│    ╰ pr-554                              ? ↑1     …
│    ╰ pr-560                                       ok    !560
│ 
│ 
│  snapshot 0s old  ·  clicks 0  ·  scroll 0  ·  keys 0  ·  headless
┌─ %3  rows 31-39  150x9  alternate_on=0  panel repos
│ ▪ repos 5
│   ├─ charter ⑂3                        main*↑2                             ✗ failed      !554
│   ├─ easydmarc-app                     feat/dmarc-rollup↓4                 ? passed
│   ├─ infra                             main                                ● running     #88
│   ├─ charter-docs                      docs/frame*↑1↓1                     ? passed
│   └─ statusline-lab ⑂2                 main
│ 
│ 
│ 
┌─ %2  rows 41-41  150x1  alternate_on=0  panel attention
│ 0 todos · ⚠ reinit 9 ws · charter ws reinit --all · F2 palette
┌─ %1  rows 43-43  150x1  alternate_on=0  panel identity
│  ⬢ default  ◆ steward · ◇ personas forge · reddit · release · statusline                                                           charter 0.53.0 dev
└─ tmux 3.7c · window 150x44 · session mouse off · python 3.14.2
```

The same thing with `-e`, ESC shown as `\e` (the two Textual panes only — the rest is in the
writeup):

```
┌─ %5  rows 5-16  150x12  alternate_on=1  panel textual.live
│ \e[48;2;36;47;56m \e[38;2;226;227;229mrepos 5  ·  harness-wrapper\e[39m
│ \e[1m\e[38;2;224;224;224m\e[48;2;45;55;64m repo                 branch                      ci    change
│ \e[38;2;33;21;5m\e[48;2;254;166;43m ▸ charter ⑂3         main               * ↑2     fail  !554
│ \e[0m\e[38;2;224;224;224m\e[48;2;39;39;39m   easydmarc-app      feat/dmarc-rollup  ↓4       ok
│    infra              main                        run   #88
│    charter-docs       docs/frame         ? ↑1 ↓1  ok
│    statusline-lab ⑂2  main
│    ╰ pr-554                              ? ↑1     …
│    ╰ pr-560                                       ok    !560
│ 
│ 
│ \e[39m\e[48;2;36;47;56m \e[38;2;167;171;175msnapshot 4s old  ·  clicks 0  ·  scroll 0  ·  keys 0  ·  live · frozen\e[39m
┌─ %4  rows 18-29  150x12  alternate_on=0  panel textual.repos
│  repos 5  ·  harness-wrapper
│  repo                 branch                      ci    change
│  ▸ charter ⑂3         main               * ↑2     fail  !554
│    easydmarc-app      feat/dmarc-rollup  ↓4       ok
│    infra              main                        run   #88
│    charter-docs       docs/frame         ? ↑1 ↓1  ok
│    statusline-lab ⑂2  main
│    ╰ pr-554                              ? ↑1     …
│    ╰ pr-560                                       ok    !560
│ 
│ 
│  snapshot 0s old  ·  clicks 0  ·  scroll 0  ·  keys 0  ·  headless
```

**Three things to look at.** `%5` is `alternate_on=1` and carries Textual's real 24-bit SGR:
a widget framework running in a pane charter split. `%4` is the *same widget tree* through
charter's contract and has **not one escape in it** — `Registry.draw` escapes what a provider
returns, so the adapter can only ever hand back `Strip.text`. `%3` is charter's own `repos`,
in colour, off the same `gather` snapshot, and it still reads best.

---

## What was built

| id | shape | `render` |
|---|---|---|
| `textual.repos` | **adapter** — Textual headless on a background thread, composited screen copied out as lines | returns per repaint, like every other component |
| `textual.live` | **takeover** — Textual on the pane's own tty, alternate screen, own asyncio loop, `mouse=True` | never returns |

`providers/charter-textual-repos/`, a real distribution: `uv build` makes a wheel, and
installing that wheel into a **3.12** venv beside charter — no editable install, no
`PYTHONPATH` — gives the same two entry points and the same two panes. It is placeable from a
committed `charter.toml`, and on a machine without it the arrangement falls back whole:

```
components: [{'use': 'textual.live', 'slot': 'textual.live', 'edge': 'bottom',
              'size': Fixed(n=12), 'visible': True, 'key': 'F9'}]
slots: ['top', 'bottom', 'textual.live']

# same charter.toml, provider not installed:
components: []
slots: ['top', 'bottom', 'repos', 'right']
```

**The extension model is real.** §4b property 4 works exactly as designed.

---

### Did the contract fit? Where did it chafe?

The declaration half fits. The rendering half fits one of the two shapes. **The input half
does not exist.** Six named places, in the writeup; the three that matter:

* **`render(ctx) -> list[str]` is a renderer contract.** A framework that owns the terminal
  has no seam there. The adapter obeys it by running Textual where nobody can see it; the
  takeover obeys it by never returning from it. Nothing distinguishes "a component that
  draws" from "a component that owns its pane".
* **`Component.events` is validated and read by nothing.** `grep -rn '\.events\b' charter/`
  finds two hits: `commands_worktree.py:275` (unrelated) and `frame/component.py:320`, *the
  line that validates the field*. There is no dispatcher. No event kind fires for any
  component, ever. `textual.live` declares four and receives all four — by having taken the
  tty.
* **A provider's lines are escaped, so a provider component is monochrome.**
  `Registry.draw`'s `escape=cid in self._foreign` runs `contain.one_line`, which turns ESC
  into the four characters `\x1b`. Visible in the capture above.

Two smaller ones: a provider can never be `Content()` or `Fill()` where anyone has
configured it (`instance.component_tables` coerces `size` to `Fixed(n)`); and the contract is
a Python class checked with `isinstance`, so `charter-cp` is a real install-time dependency
of every provider.

**And one that CI found and a hand-run could not.** Committing `providers/` failed
`test_every_top_level_directory_is_classified_one_way_or_the_other` on 3.11 and 3.13 — the
check doing exactly what its docstring promises for a new top-level directory. The decision
it asks for is made in `_NOT_PLUGIN_SURFACE`: a provider is a separate distribution, nothing
under it is importable from charter, none of it ships in the wheel, and it must never become
plugin surface, because a provider declares dependencies charter does not have and that is
the property `test_runtime_has_zero_dependencies` protects. The classification reads the git
**index** (#529), so every local full-suite run before the commit passed with `providers/`
untracked; the suite was re-run on the committed tree afterwards. **So the cost of hosting a
provider inside charter's repo is exactly one line and a paragraph saying why.**

### The one that must be fixed regardless: `sys.stdout`

`textual/app.py:3491` wraps the message pump in `redirect_stdout(self._capture_stdout)`,
which assigns `sys.stdout` **process-wide**, headless or not. `_PrintCapture`
(`textual/app.py:260`) answers `isatty() -> True` and `fileno() -> -1`. charter reaches
`sys.stdout` in three places and that object subverts all three:

| charter | what it gets |
|---|---|
| `panel._out` — `sys.stdout.write(payload)` | the paint goes into Textual's print log |
| `slots._width`/`_height` — `os.get_terminal_size(sys.stdout.fileno())` | `OSError` on fd -1, caught, **silent 80x24 fallback** |
| `chrome.colour_ok` — `sys.stdout.isatty()` | `True`, whatever stdout actually is |

Measured in a real 150x10 pane (`measure/m8_stdout.sh`), with the provider's workaround
disabled:

```
  0.5s  restored=10  captured=10
  2.0s  restored=10  captured=10
  2.5s  restored=10  captured=1    <- version bumped at 2.0s
  5.0s  restored=10  captured=1        (non-blank rows in each pane)

== and nothing was reported: pane alive, no charter error, no stderr ==
  %2 dead=0 cmd=python3.14
```

charter measures 80x24, the component lays out 24 rows, `panel._write` writes 24 rows into a
10-row pane, the pane scrolls its own content away. **Nothing raises. `Registry.draw` has
nothing to catch, because nothing failed.** The first paint is correct, which is the shape
of bug that ships.

This is charter's, not Textual's. Any provider using any library that touches `sys.stdout` —
a progress bar, a logging handler, `rich.Console`, a debugger — lands somewhere on that
table. §4b's four properties all rest on a global a provider can take by accident.

### Does the alternate screen cost anything charter cares about?

**No, and less than expected** (`measure/m2_altscreen.sh`).

```
%2 python3.14 alternate_on=1   charter panel textual.live   history_size=0
%1 python3.14 alternate_on=0   charter panel repos          history_size=0
%0 bash       alternate_on=0   the harness                  history_size=191
```

`history_size=0` for the Textual pane **and for charter's own panel** — the alternate screen
costs nothing charter's panels were not already costing. `capture-pane -p` and `-p -e` work
normally. `resize-window 150x40 -> 100x30 -> 150x40` reflows correctly (pane 150x14 → 100x11
→ 150x14). `kill -9` with `remain-on-exit on` leaves the pane showing Textual's content,
minus tmux's usual one scrolled line. `respawn-pane -k`, the argv `cmd_respawn` emits, brings
it back clean.

### What happens when the Textual app crashes?

**Its own pane, every time** — but one of the four modes loses the reason
(`measure/m3_crash.sh`, `m3b_crash_message.sh`). charter's own `repos` pane and the harness
pane were untouched in every case.

| failure | the pane says |
|---|---|
| `render` raises before Textual starts | `textual.live failed to draw — RuntimeError: injected fault before the app started` |
| provider module raises on import | `charter-textual-repos 0.1.0 supplies textual.live, and importing charter_textual_repos raised ImportError: … Its pane says so; the rest of the frame is drawn` |
| id nothing supplies | `charter: unknown slot 'notinstalled.thing' (known: bottom, repos, right, top)` |
| **raises inside Textual's message pump** | `textual.live: app exited after 1s (clicks 0, scroll 0, keys 0)` |

Textual does not let that last one out of `run()`: it catches it, prints a 2,520-byte Rich
traceback with locals to `sys.__stderr__`, and returns **normally**. `Registry.draw` sees a successful
render and `panel._write` clears the pane over the traceback within one tick — sampled every
100 ms, it is never visible even once. **A crash inside a framework's own loop is
indistinguishable, from charter's side, from a clean quit.**

Also: **`NO_COLOR` is not honoured by a takeover component.**
`textual/drivers/linux_driver.py:58` sets `self._file = sys.__stderr__` and the app writes
straight there, past `panel._write` and `chrome.plain`. Measured SGR sequences in the pane:
charter's `repos` 54 → 0; `textual.live` 15 → 15.

### Startup: `split-window` to first paint

Polled at 10 ms against real `capture-pane`, five runs each (`measure/m1_startup.sh`):

```
repos            174ms 115ms 130ms 142ms 113ms
textual.repos    339ms 273ms 242ms 241ms 245ms
textual.live     231ms 212ms 228ms 272ms 262ms
```

~2x charter's own, ~110-130 ms extra, paid once per pane. Per repaint through
`Registry.draw`: charter `_table_lines` 0.38 ms median, `textual.repos` 4.79 ms — 12x, and
both two orders of magnitude under `panel.TICK` (200 ms). Import: `charter.frame.slots`
31 ms / 25 MB, `textual.app` 102 ms / 39 MB, this package 128 ms / 43 MB. **Cost is not the
objection.**

### Mouse — measured against real tmux, and the premise is half right

`measure/m4_mouse.py` attaches a real tmux client on a pty it owns and reads every byte tmux
writes to a terminal. tmux 3.7c, session `mouse off`.

```
== what tmux writes to the terminal as the active pane changes ==
  select-pane -> textual.live : 1006l 1000l 1002l 1003l 1006h 1000h 1002h 1003h
  select-pane -> harness      : 1006l 1000l 1002l 1003l
  select-pane -> textual.live : 1006l 1000l 1002l 1003l 1006h 1000h 1002h 1003h

== does an injected report reach the app? ==
  click, textual.live active  : clicks 1  scroll 0
  wheel, textual.live active  : clicks 1  scroll 2
  same clicks, harness active : clicks 2  scroll 3     <- still delivered
  active pane after that click: %0 (harness is %0)     <- and it did not focus the pane

== the harness pane's scrollback, throughout ==
  history_size = 181  ·  last line = 'bash-3.2$'
```

**The modal behaviour is real** — reporting follows the active pane, exactly as §4i predicts,
now confirmed with a program that actually wants the mouse. **And it does not dissolve the
trade.** tmux routes a report by POSITION, not by active pane, so the modality lives entirely
in the *terminal's* reporting mode; a click does not focus the pane, so the operator must
focus it with a key first; and while it is focused, native selection is gone for the **whole
window**. The harness pane's scrollback is untouched throughout, so nothing is lost — what is
unavailable is dragging to select while that pane has focus.

Net: `[frame] mouse` can stay `False` and a Textual panel can request the mouse for itself.
That is better than `mouse = true`, which costs selection unconditionally. It is not free, and
`EVENT_KINDS`' "give every pointer affordance a `key` as well" survives unchanged.

### Do the two loops fight?

**No — because in the takeover shape only one of them is running.** `render` blocks inside
`Registry.draw`, so `_watch` never ticks, never calls `state.version`, never calls `_write`.
No clear-screen underneath Textual, no flicker. The cost is that charter also never hands it
another `ctx`, so the table freezes at the snapshot the pane started with — visible on screen
as a climbing "snapshot Ns old" clock. The adapter is the opposite and entirely well-behaved:
a version bump updates it within one tick, idle cost unchanged at one `stat`.

The obvious compromise — start the app on a thread and return `[]` — is **refuted**, not
merely rejected (`measure/m10_thread_takeover.py`): `LinuxDriver.__init__` calls
`signal.signal(SIGTSTP, …)`, Python refuses that off the main thread, Textual catches it,
prints 16,747 bytes of traceback into the pane and `run()` returns with nothing raised.

---

## Would you ship it?

**No — not this, and not yet.**

1. **The adapter is the only shape charter can host, and it is strictly worse than what
   charter already has.** No colour, no input, 12x the repaint cost and 2x the startup, for a
   table charter draws better. Every advantage a widget framework has is on the far side of
   either `Registry.draw`'s escaping or a dispatcher that does not exist.
2. **The takeover works and is outside the contract in a way that is not a detail.** It draws
   well, survives resize/respawn/death, costs one pane when it fails — but it freezes on the
   first snapshot, silently ignores `NO_COLOR`, loses its crash reasons, and the only way to
   make it live is to import charter's internals from a stranger's package. Shipping it would
   ship "components may own their pane" as folklore rather than as a contract.
3. **The `sys.stdout` finding must land first, and it is charter's.**

**What it would take**, smallest first:

1. Make charter's pane immune to `sys.stdout` — capture the panel's real fd once at startup
   and write, measure and `isatty` through *that*. Three call sites. Worth doing whether or
   not anything else here happens.
2. Decide what an escaped provider line means: say plainly that a provider component is
   monochrome, or give providers a narrow validated colour channel.
3. Deliver `events`, or delete the field. A closed vocabulary that is validated and never
   dispatched is a promise the code does not keep.
4. Only then: a second component shape — `owns_pane = True`, or a `run(ctx, updates)` handed
   a channel the frame pushes fresh snapshots onto. That is the only thing that makes the
   widget's advantages reachable, and therefore the only thing that makes Textual worth its
   dependencies.

**What the experiment proves, and it is the important part:** the provider model is sound. A
package outside `charter/` declaring ten dependencies charter does not have was discovered,
placed from a committed file, split into a pane, drawn, crashed four ways, and never cost the
frame anything but its own pane. The gaps are in the *component* contract, not in the
extension model, and each is a few dozen lines from closed.

---

## Suite, and what CI does and does not validate

| | |
|---|---|
| Python 3.14.4 (Homebrew), inherited environment | **6975 tests, OK**, 362.6 s |
| Python 3.12.13 (Homebrew), every `CHARTER_*`/`CLAUDE_*`/`ANTHROPIC_*`/`TMUX*` unset (16 present → 0; the child saw none) | **6975 tests, OK**, 380.0 s |

Both on the **committed** tree — the classification check reads the git index, so a
pre-commit run cannot stand in for one. CI green at the branch head, asked of the sha rather
than of the PR (#561): `gh api repos/diazoxide/charter/commits/<HEAD>/check-runs` reports
`test (3.11)`, `test (3.12)`, `test (3.13)` and `test (3.14)` all **success**.

`test_runtime_has_zero_dependencies` passes in both, and both were of the **unmodified**
`charter/` and `tests/` — `git diff origin/main -- charter tests pyproject.toml` is empty.
The provider's own 14 pytest tests pass on 3.14 (editable) and 3.12 (from the built wheel).

**CI validates none of the Textual half, and it should not be read as if it did.**
`.github/workflows/test.yml` runs `python -m unittest discover -s tests -v` on 3.11-3.14 and
installs nothing: not this provider, not Textual, not pytest. Nothing in `providers/` is
built, imported or executed by any job, and no job starts a tmux server for these
measurements. What CI proves is the thing that matters most — charter is unchanged and its
suite still passes. Every number above was taken by hand on this machine and is reproducible
with the script named beside it.

No charter production line was changed, so there is no news entry. No bump, no stamp, no tag.

Every tmux claim is real tmux 3.7c. Each script owns one socket named
`charter-textualexp-<pid>`, killed before unlinking in a single trap, in charter's own
namespace so `tests/_tmuxreap.py` collects it if a script is interrupted (#590). Verified: no
leaked sockets and no leaked servers after the run.

### DIFF
--- a/docs/superpowers/specs/2026-08-28-textual-frame-component-experiment.md
+++ b/docs/superpowers/specs/2026-08-28-textual-frame-component-experiment.md
@@ -297,8 +297,9 @@ So `Registry.draw` sees a successful render, charter's "failed to draw" never ap
  1200ms | textual.live: app exited after 1s (clicks 0, scroll 0, keys 0)
 ```
 
-Confirmed out of tmux with stderr redirected: 19,358 bytes of Rich traceback ending
-`RuntimeError: injected fault inside Textual's message pump`, written to stderr, while
+Confirmed out of tmux with stderr redirected: the app writes 19,358 bytes to **stderr**
+(Textual's driver is `self._file = sys.__stderr__`), the last 2,520 of them a Rich traceback
+with locals ending `RuntimeError: injected fault inside Textual's message pump` — while
 stdout carries the two lines charter painted over it. **A crash inside a framework's own
 loop is indistinguishable, from charter's side, from a clean quit.**
 
--- a/docs/superpowers/specs/2026-08-28-textual-frame-component-experiment.md
+++ b/docs/superpowers/specs/2026-08-28-textual-frame-component-experiment.md
@@ -1,0 +1,626 @@
+# The Textual experiment — one frame component, drawn by a real widget framework
+
+**2026-08-28.** One charter frame component — the repo table — rendered with
+[Textual](https://github.com/textualize/textual) and shipped as a third-party provider
+through the existing `charter.components` entry point.
+
+The package is `providers/charter-textual-repos/`. It is **not** in `charter/`, it declares
+`textual` as its own dependency, and **charter's tree is unchanged**: `dependencies = []`
+still holds, `tests/test_packaging.py::test_runtime_has_zero_dependencies` still passes,
+and the full suite passes in two environments (counts at the end).
+
+Everything below was measured. The scripts are in `providers/charter-textual-repos/measure/`
+and each finding names the one that produced it. Every tmux claim is against real tmux
+3.7c, one socket per script, killed before unlinking.
+
+---
+
+## What was built
+
+Two components, because the experiment is a comparison and one of them could not exist.
+
+| id | shape | `render` |
+|---|---|---|
+| `textual.repos` | **adapter** — Textual on the headless driver, on a background thread; its composited screen copied out as lines | returns, per repaint, like every other component |
+| `textual.live` | **takeover** — Textual on the pane's own tty, alternate screen, own asyncio loop, `mouse=True` | never returns |
+
+Both draw the same table off `ctx.gather`: repo name, branch, dirty/ahead/behind markers,
+CI, open change, worktree badge, pieces indented underneath. Both are discovered by
+charter with no change to charter:
+
+```
+$ python -c "from importlib import metadata; [print(ep.name,'->',ep.value,'|',ep.dist.name)
+              for ep in metadata.entry_points(group='charter.components')]"
+textual.live  -> charter_textual_repos:live_component  | charter-textual-repos
+textual.repos -> charter_textual_repos:adapter_component | charter-textual-repos
+```
+
+and both are placeable from a committed `charter.toml`, resolved end to end into
+`layout.panel_command`'s argv:
+
+```toml
+[[frame.component]]
+use = "textual.live"
+edge = "bottom"
+size = 12
+key = "F9"
+```
+```
+components:
+  {'use': 'textual.live', 'slot': 'textual.live', 'edge': 'bottom',
+   'size': Fixed(n=12), 'visible': True, 'key': 'F9'}
+slots: ['top', 'bottom', 'textual.live']
+argv: [... 'panel', 'textual.live', '--session', 'f']
+
+# the same charter.toml on a machine WITHOUT the provider installed:
+components: []
+slots: ['top', 'bottom', 'repos', 'right']
+```
+
+That is §4b property 4 working exactly as designed. **The extension model is real.**
+
+It is a real distribution, not a `sys.path` trick: `uv build` produces
+`charter_textual_repos-0.1.0-py3-none-any.whl`, and installing that wheel into a **Python
+3.12** venv beside charter — no editable install, no `PYTHONPATH` — gives the same two
+entry points and the same two panes (`m7_shot.sh` under `venv312`, 14 provider tests pass).
+
+---
+
+## 1. Did the `Component` contract fit?
+
+**The declaration half fits. The rendering half fits one of the two shapes. The input half
+does not exist.** Six places it chafed, each named.
+
+### 1a. `render(ctx) -> list[str]` is a *renderer* contract, and a widget framework is not a renderer
+
+`component.Component.render` is `Callable[[Any], list[str]]`. `Registry.draw` calls it,
+type-checks what came back, escapes it, clips it to the rectangle and hands it to
+`panel._write`. A framework that owns the terminal has no seam there. The two ways out are
+the two components in this package, and the second one leaves the contract:
+
+* the **adapter** obeys the signature by running Textual where nobody can see it and
+  copying its screen out;
+* the **takeover** obeys the signature by never returning from it.
+
+Nothing in the contract distinguishes "a component that draws" from "a component that owns
+its pane", and the panel process has exactly one thread that a framework needs.
+
+### 1b. A provider's lines are escaped, so a provider component is monochrome
+
+`Registry.draw` ends in `_fit(lines, …, escape=cid in self._foreign)`, and every provider is
+foreign. `escape=True` runs `contain.one_line`, which replaces every `Cc` character with its
+own escape text — so `\x1b` arrives on screen as the four characters `\x1b`. The adapter
+therefore returns `Strip.text` (plain) rather than `Strip.render(console)` (SGR), and
+**every colour Textual computed is discarded at the boundary**:
+
+```
+m5_no_color.sh, SGR sequences visible in the pane:
+                  colour on   NO_COLOR=1
+  charter repos       54          0
+  textual.live        15         15
+  textual.repos        0          0
+```
+
+The zero in both columns is the finding. The comment in `Registry.draw` is right about
+*why* — a provider's row could otherwise move the cursor out of its own rectangle — but the
+consequence is not written down anywhere a provider author meets it: a third-party
+component can never be as legible as charter's own, and the widget framework's entire
+styling system is dead weight in the adapter shape.
+
+### 1c. `Component.events` is declared, validated, and read by nothing
+
+`grep -rn '\.events\b' charter/` returns two hits. One is
+`commands_worktree.py:275`'s `pieces.events(ws)`, an unrelated function. The other is
+`frame/component.py:320` — **the line that validates the field**. Nothing reads
+`Component.events`. There is no dispatcher. `frame/overlay.py` does decode SGR mouse
+reports and does know the modal rule, but it is charter's own palette surface and is not
+wired to components at all.
+
+So `EVENT_KINDS`' careful "declaring one of these is a declaration of what you HANDLE, never
+a promise that it FIRES" is currently the weaker statement of the two: **none of them
+fire, ever, for anybody.** `textual.live` declares `("key", "click", "scroll", "resize")`,
+receives all four, and receives them by having taken the tty — charter neither delivers
+them nor knows it is not delivering them. The adapter declares `()` and is the honest one:
+a headless app has no terminal to read from, so its `DataTable` cursor cannot be moved and
+its scrollbar cannot be reached. **A scrolling table that cannot scroll is the one thing
+the widget was worth having.**
+
+### 1d. `ctx` is one snapshot with no way to ask for another — which is right, and it is what freezes the takeover
+
+`frame/ctx.py` is explicit and correct: "A component never gets a way to fetch a fresher
+one: refreshing is the frame's decision, on the frame's clock." For the adapter that costs
+nothing, because charter calls `render` again. For the takeover it is fatal — `render` never
+returns, so nothing ever calls it again, and the table is frozen at the instant the pane
+started. Measured (`m6_repaint.sh`), across one `gather.save` + `state.bump` that flips CI
+from `failed` to `passed`:
+
+```
+                 before        after the bump
+  repos          ✗ failed      ? passed          charter's own renderer
+  textual.repos  fail          ok                the adapter — render returns
+  textual.live   fail          fail              frozen; snapshot 6s old and climbing
+  live+refresh   fail          ok                snapshot 2s old
+```
+
+That last row is `CHARTER_TEXTUAL_LIVE_REFRESH=1`, and it is the measurement of what
+un-freezing costs: the app imports `charter.frame.state` and `charter.frame.gather` and
+polls them itself. It works, and it reimplements inside a provider four things that are
+charter's — which file carries the version, that comparing it is how a panel learns
+anything happened, which workspace the *frame* resolved (#526), and that `gather.read` is
+the cache rather than a scan. A provider that gets any of those wrong gets it wrong quietly:
+a pane that stops updating, or a full `gather.scan` five times a second in every pane on the
+machine. **`ctx`'s one-snapshot rule is right; what is missing is a way for the frame to
+hand a long-lived component the next one.**
+
+### 1e. A provider can never be `Content()` or `Fill()` where anyone has configured it
+
+`instance.component_tables` requires a provider's `[[frame.component]]` table to carry both
+`edge` and `size`, and coerces `size` to `Fixed(n)`. The reasoning is sound and is
+`config.FRAME`-is-on-the-import-path — but the effect is that `Content(cap=…)` and `Fill()`
+are reachable only on a frame nobody has arranged, i.e. from `Registry.place` with no
+rectangle. charter's own `repos` is `Content()` and its height is what the plane's repos
+need; a provider's table of the same data is a fixed rectangle or nothing.
+
+### 1f. The contract is a Python class, so every provider hard-depends on charter
+
+`Providers._one` finishes with `isinstance(obj, self.kind)` where `kind` is
+`charter.frame.component.Component`. A provider cannot express a component without importing
+charter, so `charter-cp` is a real install-time dependency of this package. That costs
+nothing at runtime — charter is already in the process asking — but the entry-point group
+alone does not make the coupling loose, and it is worth being accurate about that.
+
+---
+
+## 2. `sys.stdout` — the one that would have shipped broken
+
+This is the sharpest finding and it is not about Textual specifically. It is about who owns
+`sys.stdout` in a panel process.
+
+`textual/app.py:3491` wraps the app's message pump in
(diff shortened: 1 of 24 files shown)


diazoxide:
Closing — and the reason is that **this PR won.** Both of its headline findings became issues on the day it opened, and both are closed and shipped.

## What it found, and where it went

| finding | filed | closed by |
|---|---|---|
| `Component.events` is validated and read by nothing | **#607, nineteen minutes after this PR opened** | #621 (focus/blur/resize), #636 (click/scroll), #658 (built-ins get a registry, `repos` the first consumer) |
| `sys.stdout` is a mutable global a provider can replace | **#606** | `charter/frame/pane.py` — the pane is claimed before any provider import |

`charter/frame/events.py` is now 709 lines, `DELIVERED` carries five of six kinds, `on_event` is invoked at `:633`, and `tests/test_frame_input_reaches_a_component.py` installs four provider distributions and asserts click/scroll/focus/resize arrive in the component's own cells under real tmux. The `sys.stdout` news entry names Textual's `redirect_stdout` as where it was found — this PR's evidence, already harvested.

So the branch has been a receipt for work that was done starting about nineteen minutes after it opened.

## Why it cannot land as it stands

**The central claim is no longer true.** The title says a widget framework *"does not fit in the component seam"*. That seam changed underneath it: components declare `events` and receive them, and built-ins get a registry.

**And the artifact it argues from is broken**, in a way worth stating precisely:

```
ComponentError component textual.live: declares events key, click, scroll, resize
and supplies no on_event to receive them — give it one, or drop the declaration.
```

`textual.live` — the takeover, the half this PR reports as *working* — no longer constructs, and it is refused by `charter/frame/component.py:390`, **a check that exists because of this PR's own finding 1c.** The branch rebases across 77 commits with zero conflicts and the suite is green at 8930 tests, but only because it touches nothing charter runs. The conflict is semantic, not textual.

**Landing it would also put a 674-line spec into `docs/superpowers/specs/` whose sections 1c, 2 and most of 6 are now false** — asserting there is no dispatcher, in a tree that has a 709-line one with its own real-tmux test file. That is the worst single consequence of merging, worse than the second way to draw a pane.

## The residue, filed rather than lost

- **#707** — a provider's `ctx.chrome` is escaped on the way out, so the colour channel `docs/frame.md:1169-1176` documents renders as literal text. Verified round trip. This one got *worse* after this PR was written: #604 landed the recipes the same day.
- **#708** — a foreign component can only ever be `Fixed(n)`; #660/#662/#702 made `size` real for built-ins and left the provider branch coercing.
- **#709** — `charter-cp` is a hard install-time dependency of every provider (`isinstance`, `registry.py:387`); the entry-point group makes discovery loose and construction tight.

## And one thing recorded as decided, not open

The to-do item for a second component shape — `owns_pane=True` with `run(ctx, updates)` — is **decided against**, and this PR is the best evidence for the decision. Its own measurements are the argument: a frozen snapshot, `NO_COLOR` ignored, crash reasons swallowed, and a thread takeover that did not survive contact. *"A provider cannot take your pane"* is now a shipped headline, and this experiment is why it can be said with confidence.

That is a good outcome for an experiment. Thank you for running it — closing it is the result, not the rejection.

PR #612, opened 2026-08-28
Pin the four guards the sweep found in the new state accessors (#604 follow-up)

diazoxide:
Recovers a commit lost to a merge race: I merged #604 at head `1e99d0bc` while its author was still working, and `761c403` landed on the branch afterwards. My error, not theirs.

## What this carries

`761c403` — *"The deletion sweep found four unpinned guards in the two new state accessors"* — 67 lines in `tests/test_frame_surface_live.py`.

The author's own report on it:

> **Sweep: 16/16 pinned, 0 survived.** The *first* run was not clean — it found 4 unpinned guards in the new `state` accessors, all live (`$CHARTER_SESSION_ID` reaches them unminted). Pinned, each test naming which refusal fired.

So these are guards on a path an environment variable reaches with an unminted value, and without this commit they are exactly the shape this project has spent a week closing: correct code with nothing behind it.

## Why it merged green without them

The guards are correct — the suite passes either way. That is the entire point of the deletion sweep, and it is why "CI green" was never the merge gate on this project. I checked CI at a head the author had already moved past.

## Verification

CI is read at this branch's head, not at `#604`'s. `charter/` production lines are untouched by this commit — it is tests only, so no news entry.

### DIFF
--- a/charter/cli.py
+++ b/charter/cli.py
@@ -722,7 +722,7 @@ def _add_frame_parsers(sub) -> None:
     _core_commands = set(sub.choices) | {"frame", "panel", "frame-palette",
                                          "frame-probe", "frame-respawn", "frame-density",
                                          "frame-resize", "frame-gather", "frame-switch",
-                                         "frame-toggle"}
+                                         "frame-toggle", "frame-chrome"}
 
     # Which harness (by `.name`, never `.cli_name` — that's the dict key below) has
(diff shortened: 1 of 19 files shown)


diazoxide:
Superseded — the branch carries the two commits already squashed into `main`, so it conflicts. Cherry-picking the single missing commit onto current main instead.

Issue #606, opened 2026-08-28
A panel's pane IS sys.stdout, so any provider library that touches it silently blanks the pane and drops charter to 80x24


diazoxide:
Found by the Textual experiment (#605) and reproduced here on `main`. **This is charter's bug, not the experiment's**, and it is the one finding from that work that matters regardless of whether anything Textual-shaped ever ships.

## Reproduction

```
$ python3 -c "... sys.stdout = a stream answering isatty()=True, fileno()=-1 ..."
  slots._width()  = 80
  slots._height() = 24
```

The experiment's pane was **150×10**.

## The mechanism

A panel process paints by writing to `sys.stdout`, and it measures its own rectangle from that same descriptor. So `sys.stdout` is not an output stream here — it **is the pane**.

Textual's `redirect_stdout` replaces `sys.stdout` process-wide with a `_PrintCapture` that answers `isatty() -> True` and `fileno() -> -1`. Three things follow at once:

- `panel._out` paints into the framework's log instead of the pane.
- `slots._width()` / `_height()` measure fd −1, fail, and fall back to **80×24 silently**.
- `chrome.colour_ok` is told the pane is a tty regardless of what the pane actually is.

**Measured symptom in a real 150×10 pane: a correct first paint, then blank on every repaint. Nothing raised. Nothing logged.**

## Why this is not about Textual

Any provider using **any** library that touches `sys.stdout` lands on this table — `rich`, `click`, `tqdm`, `colorama`, a logging handler installed at import, a progress bar. §4b promises a provider's component draws its own rectangle and that a broken one costs its own pane. Here a provider that never crashes silently blanks its pane *and* mis-measures the frame, which is worse than crashing: `Registry.draw`'s catch never fires because nothing goes wrong.

## The shape

This is **the ninth instance** this month of matching a spelling rather than a property, and one of the most consequential: `isatty()` and `fileno()` are asked of *whatever is currently bound to `sys.stdout`*, when the property wanted is *the pane this process was given*. Prior instances, all closed: #547, #558, #537, #498, #577, #594, plus two caught inside proposed fixes.

## Suggested direction

The pane is a fact of the process's environment, not a property of a mutable global. Capture the descriptor once, at start, before any third-party import can rebind it — the way `tests/_ttyguard.py` had to install **above** the import that pulls charter in, because `util._USE_COLOR` is `sys.stderr.isatty()` evaluated at that import (#545/#546).

[...]

PR #619, opened 2026-08-28
The palette claims its pane above the registry that imports every provider (#611)

### LINKED ISSUE #611: The palette process has #606's hole too: its registry imports every provider before it takes the tty
Found while fixing #606, verified by reading, and deliberately left out of that PR so the fix stayed scoped to the panel.

#606's property is **the pane this process was given** — not whatever `sys.stdout` is bound to at the instant of the ask. `charter/frame/pane.py` now answers that, and `panel.run` claims the pane above the point where a provider's module can be imported. The palette is the *other* pane-owning process, and it has the same ordering the other way up.

## The ordering

`commands_frame.cmd_palette`:

```python
reg = builtin_actions.build(fid, current_density=..., current_chrome=...)   # imports every provider
...
surface = palette.Palette(catalogue=... reg.offers(...) ...)
chosen = palette.own_the_tty(surface, then=...)                            # out defaults to sys.stdout
```

`actions.ActionRegistry.add` → `registry.Registry` → `importlib.import_module(ep.module)` runs a stranger's module first. Then `palette.own_the_tty` (charter/frame/palette.py:257) takes `out=None` and resolves it at line 306:

```python
out = sys.stdout if out is None else out
```

and measures the overlay pane from that same object:

```python
def size() -> tuple[int, int]:
    return tuple(os.get_terminal_size(out.fileno()))
```

So a provider whose module installs `textual.redirect_stdout`, a `rich` console, `colorama`, or a logging handler at import gets the palette painting into its log and measuring fd -1 — the same silent blanking #606 measured for a panel, in the surface `F2` opens.

## What it would take

The same two lines the panel took: claim the pane in `cmd_palette` **above** `builtin_actions.build`, release it in that function's existing `finally`, and let `own_the_tty`'s `out` default to `pane.stream()` rather than `sys.stdout`.

The one wrinkle, and why it was not folded into #606's PR: the claim has to sit above the build while `cmd_palette`'s existing `try` starts below it, so either the claim leaks on a raise between the two or four lines move inside the `try` — and moving them changes when `_close_palette` runs. That is a decision about `commands_frame`, not a detail of #606, and it wants its own change.

`sys.stdin` is a separate question and a quieter one: `own_the_tty` does `fd = sys.stdin.fileno()` and hands it to `termios.tcgetattr`, so a rebound stdin **raises** rather than silently mispainting. Worth deciding on, but it is not the silent failure this is about.

diazoxide:
Closes #611.

`charter frame-palette --pane` is the second charter process handed a pane, and it had #606's ordering the other way up.

## Reproduced first, and the symptom is not an exception

`_draw_palette` builds `builtin_actions.build` — `importlib.import_module` on every installed provider, the first instant a stranger's code runs in this process — and only then calls `palette.own_the_tty`, whose `out` resolved `sys.stdout` at that later instant and whose `size` measured `out.fileno()`.

Driving a whole real `charter frame-palette --pane` on `main`, in a 150x10 pane, with a provider that rebinds `sys.stdout` as its module is imported:

```
pane transcript : ''
library's log   : '\x1b[?1049h\x1b[?25l\x1b[?1006h\x1b[?1000h…'
raised          : OSError('[Errno 25] fd -1 is not a terminal')
```

Not one byte reached the pane — not even the alternate-screen enter, which is written before anything is measured. The `OSError` landed in a process whose `finally` had already run `_close_palette`, so the traceback went into a pane tmux was about to kill. What the operator sees is `F2` carving a pane off the harness, drawing nothing in it, and killing it again; `cmd_palette`'s documented "always 0" went with it. So every new assertion is **on the pane's content**, and the return code is asserted last, as a consequence.

## The structural decision #611 asked for

The claim goes in **`cmd_palette`**, not in `_draw_palette`. It has to sit above the registry build, and `_draw_palette`'s existing `try` starts below it — its `finally` is `_close_palette`, which kills the palette's pane. Folding the claim into that `try` would have moved four lines under that `finally` to satisfy an ordering.

Putting it in the enclosing function is `panel.run`/`panel._run`'s split, one surface over, and for the same stated reason: `--pane` is literally the flag that says this process was handed a rectangle, and `_draw_palette` is the work done inside it. `_draw_palette` is not touched at all.

The other branch claims nothing, deliberately: without `--pane` this is the hotkey bind's `run-shell` child, whose stdout is a pipe and which paints in no rectangle. That is the process `frame/pane.py`'s fallback is written for, and it is pinned by a test.

## Exact lines changed in `commands_frame.py` (contended)

| Lines | Change |
|---|---|
| 136–137 | `pane` added to the existing `from .frame import (...)` tuple — one word, no reflow beyond the wrap |
| 4062–4080 | `cmd_palette`'s docstring gains two paragraphs (the decision above, and why the other branch claims nothing) |
| 4081–4087 | the body's `if` branch becomes `held = pane.claim()` / `try: return _draw_palette(args)` / `finally: pane.release(held)` |

Nothing else in the file is touched. `_draw_palette`, `_open_palette` and every other function are byte-identical.

`charter/frame/palette.py`: `own_the_tty`'s `out` defaults to `pane.stream()` instead of `sys.stdout` (one line, plus the import and the docstring paragraph that carries the measurement). Every caller that passes its own stream — the real-pty tests — is unchanged, and a process that claimed no pane still gets `sys.stdout`, which is what that default already meant.

## Tests

Added to `tests/test_a_panes_descriptor_is_not_a_mutable_global.py`, which already owns this property for the panel — the module is about "the pane this process was given", and there are now two such processes. Four tests, driving a whole `cmd_palette` against a real pty:

* the palette's own rows reach the pane and the library's log stays empty;
* the paint has exactly the pane's own number of lines (`Surface.render` answers `height` lines, so the transcript carries the rectangle it was laid out for);
* the claim does not outlive the palette that took it;
* the half without `--pane` claims nothing.

**Baseline (#579).** With the production change reverted and the tests unchanged, the two content tests are red — `nothing was drawn in the pane: ''` and `OSError('[Errno 25] fd -1 is not a terminal') != 0`. Restored and re-verified afterwards.

The fixture types its Ctrl-C from inside the surface's **first write**, and both stand-ins carry it. Two earlier shapes were measured and rejected: a byte written before the call is discarded (`tty.setraw` uses `TCSAFLUSH`, which empties the input queue), and a main-thread write loop left the worker inside `tcsetattr` for the whole deadline. Both stand-ins carry the trigger so a palette painting into the *wrong* one still ends and the class reports a failed assertion rather than hanging; the worker is still joined on a deadline for the case where neither is written to.

## Suite

* 3.14.4: `python3 -m unittest discover -s tests` — **7323 tests, OK**
* 3.12.7, with every `CHARTER_*`/`CLAUDE_*`/`ANTHROPIC_*`/`TMUX*` unset — **7323 tests, OK**

## Residual, stated rather than left to be found

`own_the_tty` reads `sys.stdin.fileno()` for raw mode, and a library that rebinds *that* global makes `termios.tcgetattr` raise. #611 names it as a separate, quieter question and it stays open: it is a different failure from this one (loud in the code's own terms rather than silent), it is not what the palette paints or measures through, and covering it would mean `frame/pane.py` claiming a second global — a real extension of #606's module rather than a use of it.

Do not merge.

### DIFF
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -133,7 +133,7 @@ import sys
 import time
 
 from . import config, contain, harness, instance, tui, util, workspace
-from .frame import (builtin_actions, choose, component, gather, layout, overlay,
+from .frame import (builtin_actions, choose, component, gather, layout, overlay, pane,
                     palette, picker, state, switch, tmuxctl)
 # Aliased: `cmd_launch` already has a local variable named `slots` (the VISIBLE slot
 # list `layout.visible_slots` returns) — importing the renderer registry under its own
@@ -4059,9 +4059,31 @@ def cmd_palette(args) -> int:
     already open" is a stale-state problem of its own (the hatch deliberately leaves the
     window option naming a pane that is gone), and a guard that refuses the palette after
     an escape-hatch press would be worse than the state it is preventing.
+
+    **``--pane`` is this process being handed a rectangle, so it is where the pane is
+    claimed** (`frame/pane.py`, #606/#611) — and the claim is HERE rather than inside
+    :func:`_draw_palette` for the reason `panel.run` is split from `panel._run`, which is
+    the same split one surface over. The claim has to sit above `builtin_actions.build`,
+    which is the first instant a provider's module can execute in this process
+    (`registry.Registry.add` → `importlib.import_module`); `_draw_palette`'s own `try`
+    starts *below* that build and its `finally` is `_close_palette`. Folding the claim
+    into that `try` would have moved four lines under a `finally` that kills the palette's
+    pane, changing when `_close_palette` runs to satisfy an ordering — so the claim gets
+    the enclosing function instead, exactly as `panel.run` does, and `_draw_palette` keeps
+    the shape it was reviewed in.
+
+    **And the other branch claims nothing, deliberately.** Without ``--pane`` this is a
+    `run-shell` child whose stdout is a pipe: it was given no rectangle, it paints in
+    none, and `frame/pane.py`'s fallback is written for exactly that process. A claim
+    there would record a pipe as "the pane this process was given", which is the sentence
+    that module exists to keep true.
     """
     if getattr(args, "pane", False):
-        return _draw_palette(args)
+        held = pane.claim()
+        try:
+            return _draw_palette(args)
+        finally:
+            pane.release(held)
     return _open_palette(args)
 
 
--- a/charter/frame/palette.py
+++ b/charter/frame/palette.py
@@ -61,7 +61,7 @@ from dataclasses import dataclass, field
 from typing import Callable, Iterable
 
 from .. import contain
-from . import component, overlay
+from . import component, overlay, pane
 
 #: What the header says the palette is, before the operator types anything.
 HEADING = "charter"
@@ -282,6 +282,27 @@ def own_the_tty(surface: overlay.Surface, *, fd: int | None = None, out=None,
 
     ---
 
+    **`out` defaults to the pane this process was CLAIMED, never to `sys.stdout`** — #611,
+    which is #606's property reaching the other pane-owning process. This name is not an
+    output stream here, it is the rectangle: it is what the surface paints into AND the
+    descriptor `size` measures, so the two cannot be allowed to come from a mutable global
+    that any library the process imported may have replaced. `commands_frame.cmd_palette`
+    claims above `builtin_actions.build`, which is where a provider's module first gets to
+    run (Textual's `redirect_stdout`, a `rich` console, `colorama`, a logging handler
+    installed at import) — measured on `main` before the claim: every byte of one whole
+    palette, the alternate-screen enter included, went into a `_PrintCapture` and the pane
+    got nothing at all, while `size` asked `os.get_terminal_size(-1)` and raised into a
+    pane `_close_palette` was already killing. `frame/pane.py` carries the argument.
+
+    `pane.stream()` and not `pane.size()`: this function measures the descriptor it is
+    painting into, whichever stream that is, so an explicit *out* (every test with a real
+    pty passes one) is measured as itself rather than against a claim nobody took. The
+    fallback inside `stream()` is `sys.stdout` for a process that claimed no pane, which
+    is the same answer this parameter defaulted to before and is why no existing caller
+    moves.
+
+    ---
+
     **`then` is how one pane holds more than one surface**, and it is Task 6's whole
     mechanism: a chosen row is offered to it, and a `Surface` coming back is run next, in
     this same pane, without the tty leaving raw mode in between. `None` — the default, and
@@ -303,7 +324,7 @@ def own_the_tty(surface: overlay.Surface, *, fd: int | None = None, out=None,
     tmux's own root key table before any byte reaches this process.
     """
     fd = sys.stdin.fileno() if fd is None else fd
-    out = sys.stdout if out is None else out
+    out = pane.stream() if out is None else out
     before = termios.tcgetattr(fd)
     tty.setraw(fd)
     try:
--- a/docs/news/unreleased-the-palette-cannot-have-your-pane-taken-either.md
+++ b/docs/news/unreleased-the-palette-cannot-have-your-pane-taken-either.md
@@ -1,0 +1,60 @@
+---
+version: unreleased
+headline: A provider's library cannot take the palette's pane either, and F2 stops opening a pane it never draws in
+---
+
+The panel process stopped asking `sys.stdout` what its rectangle is. The palette process
+had not. `F2` is the other charter process that is handed a pane, and it had the same
+ordering the other way up.
+
+`charter frame-palette --pane` builds its action registry first — which is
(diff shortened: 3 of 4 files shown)


diazoxide:
## Hand-check: mutations `tools/sweep.py`'s operators do not make

Against a verified unmutated baseline (#579), each mutation asserting it applied (#585/#586) — old text present, new text absent, and the reverse on disk afterwards — and each restore verified byte-for-byte before the next ran, with the script refusing to continue if a restore ever failed.

Modules run per mutation: `test_a_panes_descriptor_is_not_a_mutable_global`, `test_frame_palette`, `test_frame_palette_names`, `test_frame_pickers`, `test_frame_panel`, `test_frame_slots`, `test_frame_chrome` — 339 tests.

```
BASELINE (unmutated): green  [Ran 339 tests / OK]
pinned    no claim at all — the palette resolves whatever `sys.stdout` is bound to      [failures=2]
pinned    release resets instead of restoring (`pane.release(None)`)                    [failures=1]
pinned    the claim is not released at all                                              [failures=81, errors=4]
pinned    the claim is hoisted above the `if`, so the run-shell child claims a pipe      [failures=1]
pinned    the claim sits BELOW the registry build — the ordering, and the whole of #611  [failures=2]
pinned    `out` goes back to defaulting to `sys.stdout`                                  [failures=2]
pinned    `out` ignores an explicitly passed stream and always takes the claim           [failures=1]
final restore verified: both files byte-identical to the baseline
BASELINE RE-RUN: green
```

**The last one was a survivor, and closing it added a test.** `out = pane.stream()` with the parameter dropped altogether kept 338 tests green. The reason is worth recording rather than suppressing: every caller that passes a stream today (`test_frame_palette.TheTtyIsOwnedAndHandedBack`, `test_frame_pickers._RealPty`) drives a real pty with `Surface.run` **mocked**, so nothing was ever written to the stream they passed and nothing could notice it being ignored. That is "the test asserts too little", not an equivalent mutant. `test_a_stream_handed_in_is_still_the_one_written_to` now pins it, and the mutation is red.

`tools/sweep.py` found the same line independently and reports it pinned:

```
[1/2] pinned  charter/frame/palette.py:327 [collapse-ifexp] pane.stream() if out is None else out  (3 module(s))
```

## Suite

* 3.14.4, at the head sha: `python3 -m unittest discover -s tests` — **7324 tests, OK**
* 3.12.7 with `CHARTER_*`/`CLAUDE_*`/`ANTHROPIC_*`/`TMUX*` unset — refreshed at the head sha in a follow-up

No tmux server was left behind by any of this; the palette tests use a pty, not tmux.

Do not merge.

diazoxide:
## Verification at the head sha (`8a4041b`)

**Full suite, two environments, both at this exact tree:**

| environment | result |
|---|---|
| 3.14.4 | **7324 tests, OK** (565s) |
| 3.12.7, with every `CHARTER_*`/`CLAUDE_*`/`ANTHROPIC_*`/`TMUX*` unset (`env -i`) | **7324 tests, OK** (434s) |

**CI at the head sha**, read from `repos/diazoxide/charter/commits/8a4041b1.../check-runs` rather than `gh pr checks` (#561):

```
test (3.11)=success   test (3.12)=success   test (3.13)=success   test (3.14)=success
Every guard this branch adds is a line a test goes red without=success
A dev-channel install from main actually installs=skipped     (push-to-main only, by design)
```

The sweep gate is the `Every guard…` job, and it is green at this sha.

**No tmux was involved and none was leaked** — the palette tests own a pty, not a server. Checked afterwards: of the socket files in `/tmp/tmux-<uid>/`, exactly one server is live, and it is the operator's own `charter`.

One late addition since the first push, listed for the diff reader: `_close_pty` now writes one byte through the master before closing, because closing an fd does not wake a thread already inside `os.read` on it and this class runs the surface on a worker — `test_frame_palette.TheTtyIsOwnedAndHandedBack._close`'s reason, which applies harder here. Both suite runs above are after that change.

Do not merge.

diazoxide:
## `tools/sweep.py` at the head sha

```
deletion sweep — 8a4041b10516 against 27da88eb3ffa
measured on       : darwin, CPython 3.14.4
baseline          : Ran 7324 tests — OK
mutations applied : 2
pinned            : 2
SURVIVED          : 0
UNRESOLVED        : 0
wall clock        : 6.9 min

Every mutation this diff offered goes red. Nothing added here is a line the
suite would not miss.
```

Both mutations are `collapse-ifexp` on `charter/frame/palette.py:327` — `pane.stream() if out is None else out`, collapsed each way. Both go red, and the second direction is the one the hand-check found surviving before `test_a_stream_handed_in_is_still_the_one_written_to` was written for it.

**One earlier attempt of this sweep refused rather than reporting**, and it is worth saying so rather than quietly re-running: its sandbox baseline came back `Ran 0 tests — no tests ran in 171s` on a box at load ~50 (four sibling agents running their own suites), and the sweep printed

```
! the tree is RED before any mutation. Every mutation below will look
! pinned for a reason that has nothing to do with the guard. Fix first.
```

That is #553's rule working — a run that measured nothing must not become evidence — not a result about this branch. The numbers above are from a re-run whose baseline really ran 7324 tests.

PR #620, opened 2026-08-28
The launch test detects a void death instead of asserting through it (#609)

### LINKED ISSUE #609: A void death in the one test that cannot probe for one is read as an exit code of 1, so it fails instead of skipping
Found while verifying #603: `test (3.11)` went red at the branch head, and the same commit's *other* workflow run — started two seconds later, same runner image, same code — went green at 3.11. The test ran (not skipped) in both.

```
FAIL: test_nothing_of_the_operators_tmux_is_written_by_a_whole_launch
      (test_frame_tmux_integration.WindowInsideAnOperatorsTmux)
AssertionError: Lists differ: [1] != [21]
      : the harness's own exit code did not come back
```

## This is the failure the module already documents

`tests/test_frame_tmux_integration.py`'s own header, verbatim:

> **One test here still cannot carry a probe, and says so rather than pretending**:
> `WindowInsideAnOperatorsTmux.test_nothing_of_the_operators_tmux_is_written_by_a_whole_launch`
> reads the code a whole real `cmd_launch` returns, and that pane carries charter's
> production `pane-died` PAIR — `[0]` the write hook, `[1]` `kill-session`. There is no free
> index: `[1]` is where a probe would go and is charter's teardown, and measured against tmux
> 3.7c a `[2]` armed beside them never runs at all, because `kill-session` at `[1]` takes the
> server down first. **A void death there is still read as an exit code of `1`.**

`1` is exactly what came back. Every other death-dependent test in the module detects a void death (#487: on a loaded tmux 3.4 a pane's fd closes before tmux has the child's status, and for some deaths it never gets it) with `_VoidDeaths`' constant-action probe hook and **skips**, naming what it could not measure. This one has no free hook index, so a void death lands as a plain assertion failure that describes the runner rather than charter.

## Why it is not the change it appeared on

* `test (3.11)` was **green** at `eaf68bc`, which already carried every production change on that branch — all eleven `frame/state.py` writers, `gather`'s cache, both `commands_frame` `tmux.conf` writes and `persona.set_active`. The delta from there to the red head is `tests/_statedirscan.py` and one test module; nothing a tmux integration test executes.
* The path this test drives is `_launch_in_operator_tmux`, which contains no file write at all. `cmd_launch`'s own-server branch is where the changed writes are, and the operator path returns before reaching it.
* Two concurrent runs at one sha disagreed. That is non-determinism in the environment, not in the diff.

## What would close it

The module's own answer everywhere else: detect the void death and skip rather than assert. The obstacle is written down — there is no free `pane-died` index on that pane, because `[0]` is charter's write hook and `[1]` is its `kill-session`.

Two shapes worth measuring:

1. **Ask tmux, not the hook.** The probe does not have to be a third hook: `_VoidDeaths` already learned (#507) to keep its evidence in tmux's own state. A void death is "the pane is dead, tmux is holding no status for it, and the array never ran" — `#{pane_dead}` / `#{pane_dead_status}` are readable without spending an index, and `_a_measured_death` already classifies on whether the array ran.
2. **Distinguish charter's `1` from the harness's.** The gate harness exits `21` precisely so the code is unmistakable; a launcher that returns `1` because it never got a status is a different fact from one that returns `1` because the harness did. If `state.exit_code(fid)` came back `None` and the launcher fell back, the test could say so and skip on that.

Until then the residual is: on a loaded `ubuntu-latest` (tmux 3.4), this one test can report a red that is about the runner. Everything around it already refuses to.

Related: #487 (the void-death window and `_HOOK_TRIALS`), #507 (the probe's evidence in tmux's own state), #494 (waiting for readiness rather than sleeping), #590 (an interrupted run leaving a live server behind).

diazoxide:
Closes #609.

## The obstacle written down for this test was not this test's

The module header said the harness pane in `test_nothing_of_the_operators_tmux_is_written_by_a_whole_launch` carries charter's production `pane-died` **PAIR** — `[0]` the write hook, `[1]` `kill-session` — so there was no free index for a probe and a void death had to be read as an exit code of `1`.

Measured, against a real `cmd_launch` in a real operator's tmux, with `show-hooks` called from inside `_wait_for_harness` while the harness was still alive:

```
frame window/pane : ['demo-33178 %1']
HOOKS while alive : ''
HOOKS -w          : ''
wait returned     : 21 raw: '1:21'
rc                : [21]
state.exit_code   : 21
```

**Empty.** That pair belongs to charter's *private-server* path (`conf_text` + `_pane_died_write_hook_argv`). This test drives `_launch_in_operator_tmux`, whose own docstring says so in as many words — *"The harness's exit code travels without hooks here… this launcher is awake for the whole life of the frame, so it reads the status and closes the window itself"* — and which writes no `source-file`, no `set -g` and no hook on the harness pane. `_PROBE_INDEX` was free the whole time. The header is corrected in this PR.

## What is true of that path is why the probe still needed something new

That launcher runs `kill-window` the instant it has the code, so a probe option written on the dying pane is destroyed microseconds later **by the path under test**. `_VoidDeaths` therefore gains a scope for where the evidence lives — `_ON_THE_PANE` (the default, unchanged for every existing caller) and `_ON_THE_SERVER`.

Everything #507 says about an option over a file holds unchanged at either scope: the action is a tmux command, so it is set *by tmux, in the same act as running the array*, with no shell, no fork and no filesystem in between — "the array ran" and "a marker got written" stay one question rather than two.

Measured against tmux 3.7c before writing any of it:

* a `pane-died` hook whose action is `set-option -s @charter-probe ran` fires on the dying pane, and the option **survives the `kill-window`** and the frame's whole session;
* `respawn-pane -k` does **not** fire `pane-died`, and the pane's hook array **survives** it — so the probe can be armed as soon as the window exists, before the harness starts;
* the server option appears in **none** of the five readings this test snapshots (`show-options -g`, the operator's session, `show-options -g -w`, their own window, `list-keys` — grep count 0 on each, after the hook had fired), so the test's own subject cannot be answered by its own instrument. It is armed *after* the `during` snapshot, too.

## A void is retried and then skipped; every other `1` still fails, and says which

Three outcomes, decided from facts already recorded rather than from a new guess:

| what happened | how it is told apart | verdict |
|---|---|---|
| the harness's own code came back | `rc == 21` | pass |
| tmux never ran the pane's array | the server option is unset | retry, then skip after `_HOOK_TRIALS` |
| tmux ran the array and charter still answered `1` | the option says `ran` | **fail** — an exit code lost between tmux and `cmd_launch` |
| charter recorded no exit for the frame at all | `state.exit_code(fid) is None` | **fail** — the pane vanished rather than dying askably, or the launch returned before the harness ran |

Plus a guard against the instrument going quiet: after arming, the test asserts its own probe is in that pane's array. If charter ever starts arming a `pane-died` hook on the harness pane in an operator's tmux, `set-hook -p` with an index silently *replaces* ours — and every void check below would then read "tmux ran no hook" and skip on a healthy machine. That now fails loudly instead.

## Each outcome was driven by injecting its own fault

Not waited for. Three throwaway subclasses, each forcing one fault into the real test:

```
A_StatusHiddenButArrayRuns  -> FAIL: tmux ran this pane's `pane-died` array — so it HAD the
                                     child's status — and charter still answered 1 instead of
                                     the harness's own 21. That is an exit code lost between
                                     tmux and `cmd_launch`, not a trial that measured nothing
B_TheArrayNeverRan          -> SKIP: none of 3 deaths on this machine reached a `pane-died`
                                     hook at all …
C_ThePaneVanished           -> FAIL: None != 1 : charter answered its fallback code and
                                     recorded no exit for this frame — the harness pane
                                     vanished rather than dying askably …
```

(A hides the status from `_pane_state` while tmux really has it; B adds the observable a void produces; C removes **both** `remain-on-exit` armings, which is the regression the message names. C's first draft removed only the pane-scoped one and the test still passed — `_draw_panels` arms the window-scoped one even with `[frame] slots` empty, which is worth knowing.)

## No production change

`commands_frame.py` is untouched. This is a test-honesty defect and the fix is in the suite.

One thing deliberately **not** changed, stated rather than left implicit: charter answers `_UNKNOWN_DEATH_CODE` for a dead pane with an empty `#{pane_dead_status}`, and cannot tell that from a harness killed by a signal — which also has an empty status and for which `1` is the stated answer. `_pane_state`'s own docstring argues that, and the only way to separate them is to spend a `pane-died` index on the operator's server, which this path exists not to do. That is why the *test* carries the probe and charter does not.

## Runs

* full `tests/test_frame_tmux_integration.py` — **52 tests, OK**, tmux 3.7c
* suite counts and CI run counts are in a follow-up comment (one green run proves nothing here)

Do not merge.

### DIFF
--- a/docs/news/unreleased-the-last-tmux-test-that-could-not-say-so-can-say-so.md
+++ b/docs/news/unreleased-the-last-tmux-test-that-could-not-say-so-can-say-so.md
@@ -1,0 +1,66 @@
+---
+version: unreleased
+headline: The last tmux test that had to assert where its siblings skip now measures the same thing they do, and the obstacle written down for it turned out not to be its own
+---
+
+Charter's tmux integration module has one rule for a death it could not measure: a trial
+that says nothing about charter is spent again, and after three of them the test skips
+naming what it could not get. On a loaded tmux 3.4 a pane's fd can close before tmux has
+the child's status, and for some deaths it never gets it — `#{pane_dead}` `1`,
+`#{pane_dead_status}` empty, the pane's `pane-died` array never run, permanently. Measured:
+7 of 118.
+
+One test was exempt, and said so in the module's own header: the one that reads the exit
+code a whole real `cmd_launch` returns. A void death there came back as
+`_UNKNOWN_DEATH_CODE`, which is `1`, and the assertion read
+
+```
+AssertionError: Lists differ: [1] != [21]
+      : the harness's own exit code did not come back
+```
+
+— a red about the runner, in the one test with no way to say so. It happened: `test (3.11)`
+went red at a branch head whose *other* workflow run, started two seconds later on the same
+image, went green at 3.11, with the test running in both.
+
+## The reason it was exempt was true of a different code path
+
+The header said that pane carries charter's production `pane-died` pair — `[0]` the write
+hook, `[1]` `kill-session` — so there was no free index for a probe. That pair belongs to
+charter's own private server. This test drives the *operator's tmux* path, whose own
+docstring says it installs no hooks on the harness pane at all, because it is awake for the
+whole life of the frame and reads the status itself. `show-hooks -p` on that pane during a
+real launch confirms it: empty. The index was free the whole time.
+
+What is true of that path is the other half, and it is why the probe still needed something
+new: this launcher runs `kill-window` the instant it has the code, so a probe's evidence
+written on the dying pane is destroyed microseconds later by the very path under test.
+
+## The evidence moves to the server, and it was checked against the test's own subject first
+
+The probe's action is a tmux command, not a shell command line — #507's point, kept: it is
+set by tmux in the same act as running the array, with no shell, no fork and no filesystem
+in between, so "the array ran" and "a marker got written" cannot be confused. All that
+changes is the option's scope. Measured against tmux 3.7c: a `pane-died` hook whose action
+is `set-option -s` fires on the dying pane, and the option survives that `kill-window` and
+the frame's whole session.
+
+The test it lives in asserts that a launch writes *nothing* of the operator's tmux, so the
+first thing checked was whether the probe could answer that question for it. It cannot: a
+server option appears in none of the five readings that test snapshots — `show-options -g`,
+the operator's session, `show-options -g -w`, their own window, and `list-keys` — and it is
+armed after the snapshot is taken.
+
+## Every other `1` still fails, and each says which one it is
+
+A void is now retried up to three times and only then skipped. Anything else that comes
+back `1` is a failure with its own sentence, decided from facts already recorded rather
+than from a new guess:
+
+* **the array ran and charter still answered `1`** — tmux had the child's status and the
+  code was lost between them. That is the failure this test exists to report.
+* **charter recorded no exit for the frame at all** — the harness pane vanished rather than
+  dying askably, or the launch returned before the harness ever ran. Both are charter's.
+
+Each of the three outcomes was driven by injecting its own fault and checking the verdict,
+rather than waiting for a loaded runner to produce one.
--- a/tests/test_frame_tmux_integration.py
+++ b/tests/test_frame_tmux_integration.py
@@ -52,13 +52,24 @@ STATUS and is exactly what a void destroys — retries a trial that measured not
 everything else here (`_a_measured_death`), and a menu is now WAITED for on the terminal
 it is drawn on rather than slept at (:class:`_Screen`).
 
-**One test here still cannot carry a probe, and says so rather than pretending**:
+**The last test that could not carry a probe now carries one — #609, and it took a
+measurement to find out that the obstacle written here was not this test's.**
 `WindowInsideAnOperatorsTmux.test_nothing_of_the_operators_tmux_is_written_by_a_whole_launch`
-reads the code a whole real `cmd_launch` returns, and that pane carries charter's
-production `pane-died` PAIR — `[0]` the write hook, `[1]` `kill-session`. There is no free
-index: `[1]` is where a probe would go and is charter's teardown, and measured against tmux
-3.7c a `[2]` armed beside them never runs at all, because `kill-session` at `[1]` takes the
-server down first. A void death there is still read as an exit code of `1`.
+reads the code a whole real `cmd_launch` returns, and this paragraph used to say that pane
+carried charter's production `pane-died` PAIR — `[0]` the write hook, `[1]` `kill-session`
+— leaving no free index. That pair is the PRIVATE-SERVER path's. This test drives
+`_launch_in_operator_tmux`, whose own docstring says it installs no hooks on the harness
+pane at all ("the harness's exit code travels without hooks here"), and `show-hooks -p` on
+that pane during a real launch confirms it: **empty**. `_PROBE_INDEX` was free the whole
+time, and the test was asserting where its siblings skip for a reason that was never true
+of it.
+
+What IS true of it is the other half, and it is why the probe needed something new: this
+launcher runs `kill-window` the instant it has read the code, so the dying pane — and any
+option written on it — is gone microseconds later. So the probe's evidence moves to the
+SERVER for this one caller (`_VoidDeaths._ON_THE_SERVER`), which outlives the window and
+appears in none of the five readings the test snapshots. A void death there is now a
+retried trial and then a skip, exactly like everywhere else.
 
 Every test gets its own tmux SESSION (hooks are per-pane, so a shared session would let
 one test's hook leak into another's pane) on the ONE socket this module owns, and every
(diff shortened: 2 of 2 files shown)


diazoxide:
## Measured, paired, on the environment that went red

`main`'s test and this branch's, **on `ubuntu-latest` with tmux 3.4**, 250 runs of the test each, all three legs on the same runner image and the loaded legs under the identical load (12 spin loops, the recipe #487 used). Run from a temporary branch carrying a throwaway workflow, the same way `tmp-ci-tmux-probe` was used before; the branch is deleted.

| leg | pass | **fail** | error | non-21 attempts | of those, the array RAN |
|---|---|---|---|---|---|
| this branch, idle | 250 | **0** | 0 | 0 | — |
| this branch, loaded | 243 | **0** | 7 | 2 | **0** (both #487 voids) |
| `main`, loaded | 239 | **7** | 4 | — | — |

`main`'s seven failures are, verbatim:

```
    : the harness's own exit code did not come back
```

which is #609's report, reproduced 7 times in 250 on the runner it was reported from. This branch, under the same load, fails **none** — the two voids it met were detected as voids (the probe says tmux never ran the array) and spent as retries.

**Load is the whole variable**, which is what #609 said and could not show: 0 voids in 250 idle runs, 2 in 250 loaded runs, on the same image minutes apart. That is exactly "two concurrent runs at one sha disagreed".

The `error` rows are the harness's own environment, not the test: `OSError: [Errno 14] Bad address: 'tmux'` — a runner that cannot fork while 12 spin loops and 250 tmux servers have been through it. They appear in **both** loaded legs (7 here, 4 on `main`) and are not a difference between them.

Reproduced independently on a local Ubuntu 24.04 container, tmux 3.4, `--cpus=1` against the same 12 spin loops:

* this branch: 60 runs, **60 pass**, 5 non-21 attempts, **all 5 voids**, 0 skips
* `main`: 60 runs, **1 fail** — `the harness's own exit code did not come back`

## Suite

* `python3 -m unittest discover -s tests`, 3.14.4 — **7319 tests, OK** (numbers refreshed at the head sha in a follow-up)
* full `tests/test_frame_tmux_integration.py`, tmux 3.7c — **52 tests, OK**; the launch test alone, 5 consecutive runs — 5× OK

## Hand-check, since `tools/sweep.py` has nothing to say here

The CI gate ran and reported `diff against 8d4b42b905dc: 0 file(s), 0 added line(s)` — this branch changes no line under `charter/`, so the sweep's operators have nothing to mutate. What needed checking instead is the shared probe plumbing every death-dependent test in the module now routes through, and whether breaking it produces a red or a **silent skip**. Five mutations, against a verified unmutated baseline, each asserting it applied, each restore verified byte-for-byte before the next:

| mutation | verdict |
|---|---|
| `_probe_where` drops the `-t <pane>` target | pinned (2 failures) |
| `_the_array_ran` always answers no — every death looks like a void | **pinned (12 failures)** — the module does not silently skip |
| the probe's evidence goes back to the pane, where `kill-window` destroys it | pinned (1 failure) |
| the green-path check that the probe still works is deleted | survived |
| the guard that the probe is still in this pane's array is deleted | survived |

The two survivors are guards whose subject is a *future* change, so nothing today can pin them — but the first is not equivalent, and the check that shows it is the third row. Mutation 3's single failure is **the green-path check itself**:

```
AssertionError: False is not true : the harness's own exit code came back, so tmux had
the child's status and ran this pane's `pane-died` array — but the probe armed on it
reports nothing. Its silence on a void would then mean nothing either, and every void
would skip a test that could still have been measured
```

So deleting that line is what makes mutation 3 invisible: the two are a pair, and the line is load-bearing rather than dead. It is the thing that converts a broken probe from a permanent silent skip into a loud failure on the first run.

The remaining survivor — `assertIn(probe_action, show-hooks)` — is a tripwire for a production change: if charter ever arms a `pane-died` hook on the harness pane in an operator's tmux, `set-hook -p` with an index silently replaces this probe. No test can reach that state without changing production, and the green-path check above catches the same failure mode anyway, just later and with a worse message. Kept deliberately, reported rather than deleted.

Do not merge.

diazoxide:
## Verification at the head sha (`a4e7c1a`)

**Full suite, two environments:**

| environment | result |
|---|---|
| 3.14.4 | **7319 tests, OK** (348s) |
| 3.12.7, with every `CHARTER_*`/`CLAUDE_*`/`ANTHROPIC_*`/`TMUX*` unset (`env -i`) | **7319 tests, OK** (354s) |

**CI, and "one green run proves nothing" is the whole point here — so: eight.**

Read from `repos/diazoxide/charter/commits/a4e7c1ad.../check-runs`, not `gh pr checks` (#561):

```
test (3.11)=success   test (3.12)=success   test (3.13)=success   test (3.14)=success
Every guard this branch adds is a line a test goes red without=success
A dev-channel install from main actually installs=skipped     (push-to-main only, by design)
```

At that sha the `tests` workflow was executed **8 times**: run `33184844567` (1 attempt) and run `33184848356` (**7 attempts**, confirmed by its `run_attempt` field), each attempt a fresh matrix of 3.11/3.12/3.13/3.14 — **32 job executions, all green**. Each rerun was waited for before the next was triggered, so they are eight sequential executions rather than eight views of one.

That is still a weak instrument by itself, which is why it is not the argument. The 250-run paired probe above is: on the same runner image and the same load, `main` fails 7 times in 250 and this branch fails 0.

## One red on the way, and it is not this branch

The first 3.14 run came back `FAILED (failures=1)`:

```
FAIL: FourEdgeIntegration.test_the_resize_hook_really_fires_and_charter_really_recomputes
AssertionError: 1 != 0 : installing the resize hook failed
```

Reported rather than quietly re-run, because a suite that hides a red is the thing this PR is about. Three independent reasons it is not reachable from this diff:

* **Different class, no shared code.** `FourEdgeIntegration(PersonaIso, unittest.TestCase)` does not inherit `_VoidDeaths`. Every hunk here is in the module docstring, `_VoidDeaths` (736-859), one line of `EarlyDeathIntegration`, and `WindowInsideAnOperatorsTmux` (3626+). The failing assertion is at line 2562.
* **5/5 OK** re-running that test alone.
* **7319 OK** on a full re-run of the same suite, and 7319 OK at the previous head, whose only difference is one placeholder-less `f` prefix dropped from a failure message.

It is a real `set-hook` returning 1 on a box that was at load ~30 with four sibling agents running their own suites — the same class of environment failure this PR exists to stop being read as charter's, one class over.

## No tmux leaked

Checked after every run: of the socket files in `/tmp/tmux-<uid>/`, exactly one server is live, and it is the operator's own `charter`. Every server this module starts is killed and its socket unlinked in one cleanup, and the throwaway CI branch used for the probe has been deleted.

Do not merge.

Issue #512, opened 2026-08-25
frame: the launcher discards the gather cache and never refills it, so the repo table is blank until the first tool call


diazoxide:
Reported by the operator: opening a new session showed no repos in the frame; after resizing the window, the repo table appeared.

## Root cause — not a resize bug

`commands_frame.cmd_launch` calls **`gather.discard(fid)`** (`commands_frame.py:1440`) before drawing anything, and **nothing in the launcher ever calls `gather.refresh`**. The only caller is `frame/notify.py:108`, reached from `hooks.py` on `posttooluse*`.

So the frame launches with a deliberately empty cache, and `slots._bottom` (and formerly `_left`) render from `gather.read(fid)` — which has nothing to read. The table stays blank until the operator's **first tool call** fires a hook. The reported resize simply coincided with one.

`panel._watch` is innocent: `resized = {"flag": True}  # the first pass always paints`. It paints immediately, against an empty cache.

## Why the discard exists

It is deliberate — a new frame must not inherit a dead frame's rows. The bug is that nothing fills the cache afterwards, so "start clean" became "start blank and stay blank".

## What a fix needs to settle

- **Who refreshes, and when.** The launcher is the obvious place, but a cold `gather.scan()` costs ~35 ms and three git invocations, and it would sit on the launch path before the harness appears.
- **Or: the panel refreshes on its first paint** — but `slots` reading only the cache is a deliberate rule (`_left`'s docstring: "Reads ONLY `gather.read(fid)` … never a repo directory listing, a `git status`, or a `glstate.read_for` of its own"), and the idle-tick property pins a panel at exactly one `stat`. Breaking that to fix a first-paint is the wrong trade.
- **Or: a detached refresh at launch**, the way `update.maybe_spawn` and `glstate.maybe_spawn` already work — the frame draws immediately with whatever is there, and the rows arrive a beat later without blocking the launch.

The third is the shape charter already uses twice for exactly this problem. Whatever is chosen, the first paint must not be a lie: an empty table on a plane that has repos reads as "no repos", which is the same confidently-wrong output `_left` was retired for.

diazoxide:
Fixed in #525, but the root cause stated above is only half of it — and the half it names
would not have produced a blank table on its own. Recording that here so the next reader
does not re-derive it.

## `gather.read` does not have "nothing to read"

> `slots._bottom` (and formerly `_left`) render from `gather.read(fid)` — which has
> nothing to read.

[...]

Issue #472, opened 2026-08-24
persona list and persona stats build a table row out of an unbounded committed directory name


diazoxide:
`charter persona list` and `charter persona stats` build their table rows out of the persona **directory** name, unbounded. A directory under `personas/` is a committed name charter did not mint — `persona.list_personas()` (`charter/persona.py:207`) globs `personas/*/`, asks only for a leading underscore and a `persona.md`, and never asks `valid_name`; a filesystem forbids `/` and NUL and nothing else. So a commit can add a directory whose name holds a line separator, and one persona then renders as **two table rows**, the second one entirely attacker-chosen and wearing charter's own column layout.

This is the #453 mechanism — a value crossing into a format with structure without being escaped for it — on the two surfaces that report the roster.

## Where

- `charter/commands_persona.py:156` — `print(fmt.format(mark, n, roles[n][:rw - 2], vaults[n], _vault_status(vaults[n])))`
- `charter/commands_persona.py` `cmd_persona_stats` — the same `n` pasted into its row.

## Reproduction

Two committed persona directories, `good` and `evil<U+2028>  fake     Fake role   vault2`, both with an ordinary `persona.md`. Driving `cmd_persona_list` and `cmd_persona_stats` on charter `b713884`:

```
=== persona list ===
  PERSONA                             ROLE  VAULT  VAULT STATUS
  evil
  fake     Fake role   vault2        —      no vault
  good                                Good  v      not set up (local)

=== persona stats ===
PERSONA                       MEM  RECENT  VERIFY   DUP  DISP  STATUS
_shared                         0       —       —     —     —  ✗ dormant
evil
  fake     Fake role   vault2    0       —       —     —     0  ✗ dormant
good                            0       —       —     —     0  ✗ dormant
```

Two personas on disk, three rows in each report. Works for `\n`, `\r`, U+2028, U+2029, U+0085, and every other separator `str.splitlines` honours.

## Why it is not simply `contain.one_line(n)` at the print

Both commands size their columns from the raw names (`nw = max(len(n) for n in names) + 2`, and the stats renderer likewise). Bounding the name at the `print` alone leaves the widths computed from a string longer than the one that gets printed, so the columns stop lining up for every persona in the table. The name has to be bounded once, **before** the widths are measured, and both renderers have to use the bounded form throughout — including wherever a name is matched back against `active`.

## Relationship to #453

[...]

PR #621, opened 2026-08-28
A component can be told its pane was focused, blurred or resized (#607)

### LINKED ISSUE #607: Component.events is validated and read by nothing — no dispatcher exists, so no click, scroll, focus or blur ever reaches a component
Found by the Textual experiment (#605) while asking what a widget framework would need.

## The gap

`Component.events` is validated at `charter/frame/component.py:320` against `EVENT_KINDS = ("key", "click", "scroll", "focus", "blur", "resize")` — and then **read by nothing**. There is no dispatcher. **No event kind fires for any component, ever.**

The vocabulary is complete and the decoder exists: `overlay.decode()` parses SGR mouse sequences into `Event(kind, name, row, col, pressed)`, and `overlay.py:502-515` consumes `KEY`, `SCROLL` and `CLICK` — but only for the overlay's own surface (the palette and the pickers). Nothing routes an event from a **panel's** pane to the component that owns it.

So a provider can declare `events = ["scroll"]` today, pass validation, and receive nothing. The Textual experiment's interactive component got keys and clicks only by **taking the tty itself**, outside the contract.

## Why it matters now

The operator's standing ask is *"clicks, scrolls, etc. — full interactiveness, whatever is possible in a terminal should be supported"*, and this is the missing half. #559 turned `focus-events` on; nothing consumes `focus` or `blur` either.

## The measurement that constrains the design

Mouse reporting is enabled from the **active pane's** mode alone, so turning `[frame] mouse` on takes the wheel from the harness — which is why it defaults to `False`. The experiment tested whether a component requesting mouse only while active dissolves that trade, against real tmux 3.7c:

> The modal behaviour is real, but **tmux routes by position and a click does not focus the pane** — the trade is relocated, not dissolved.

That corrects an earlier assumption of mine. A click on an unfocused panel is delivered to that panel *without* focusing it, so "only the active pane requests mouse" does not describe what an operator would experience. Whatever dispatcher gets built has to decide what a click on a non-active pane means.

## Two more contract gaps from the same experiment

- **`render(ctx) -> list[str]` is a renderer contract.** A framework that owns the terminal has no seam. The obvious workaround — run the app on a thread and return `[]` — is **refuted by measurement**: `LinuxDriver.__init__` calls `signal.signal(SIGTSTP, …)`, which only works on the main thread; Textual swallows the `ValueError` and `run()` returns having done nothing, with nothing raised.
- **`Registry.draw` escapes every provider line, so a provider component is monochrome.** Charter's own components can emit SGR; a third party's cannot. That is defensible (it is containment) but it is not written down as a limit anywhere, and it means §4b's "draws its own rectangle" is narrower than it sounds.

diazoxide:
Closes #607.

`Component.events` was validated at `component.py:320` and read by nothing. A provider could
declare `events = ["scroll"]`, pass every check, and receive nothing, ever. There was no
dispatcher. There is one now, and it delivers **three of the six kinds**: `focus`, `blur`,
`resize`.

## What a provider writes

```python
component.Component(
    id="acme.metrics", title="Metrics", edge="right", size=Fixed(12),
    events=("focus", "blur"),
    on_event=lambda ev: remember(ev.kind),   # truthy = repaint me
    render=lambda ctx: [line_for(remembered())],
)
```

The two halves are refused apart, in both directions. A declaration with nothing behind it
*is* #607's defect written by the provider instead of by charter, and a handler with no
declaration is a callable charter would never call.

## The four decisions worth reviewing

**1. Input is read off the pane charter already claimed (fd 1), never off `sys.stdin`.**
A tmux pane runs its command with fds 0/1/2 on one pty, so `frame/pane.py`'s claimed
descriptor *is* where the pane's input arrives. Measured against tmux 3.7c with the probe's
own stdin closed (`0<&-`): `fd=1 isatty=True`, then `READ b'\x1b[I'` on select. That reuses
#606's fix rather than writing a second one for the other direction — `sys.stdin` is a
mutable global a provider's library can replace exactly as it can replace `sys.stdout`.

**2. The tty goes to cbreak, never to raw.** `tty.setraw` clears `OPOST`/`ONLCR` and
`panel._write` joins rows with `\n`. Four modes, one 40x10 tmux pane each, read back with
`capture-pane`:

```
cooked    -> ['AAA', 'BBB', 'CCC']
raw       -> ['AAA', '   BBB', '      CCC']      <- every panel paint, sheared
cbreak    -> ['AAA', 'BBB', 'CCC']
explicit  -> ['AAA', 'BBB', 'CCC']
```

`frame/palette.py` reaches for `setraw` and is right to — it joins with `\r\n` because it
owns the whole pane. A panel does not.

**3. A handler is a notification, not a second renderer.** Nothing `on_event` returns
reaches the pane; the return value is read for its truth and nothing else, and the repaint
that follows is the existing one — same `_write`, same `Registry.draw`, same rectangle. So
#472's containment order is kept *by construction*: there is no path from a handler's
return value to a terminal for an escape sequence to travel down. A handler is also handed
no `ctx`, so §4f's one-snapshot-one-timestamp survives; and a `resize` carries no size,
because `ctx.width`/`ctx.height` is where a rectangle is read and a number on the event
would be a second answer that could disagree.

**4. A handler that raises is retired, and the pane says so.** `registry.py`'s docstring
names three moments a provider can fail — on import, while building, in `render` — and
gives one answer to all three: the pane names the component and why. `on_event` is the
fourth and takes the same answer. **The cost is real and I want it read rather than
skimmed:** a component whose `render` still works loses its pane to a message because its
*handler* broke. I chose that over the alternative — a component that quietly stops being
interactive, indistinguishable from one nothing has happened to yet, which is #512's shape.
Retired rather than retried, because a handler that raised on one event will raise on the
next and delivering more spends the operator's repaints on a loop of identical failures.

**A handler that never returns wedges its own panel and nothing else.** No watchdog, and
that is deliberate: the panel is one process holding one pane, so a looping handler stops
that pane repainting and leaves every sibling pane, the harness and tmux untouched — §4b's
promise kept, and exactly what an infinite loop in `render` already does. A
`signal.setitimer` watchdog cannot interrupt a C-level block, so it would promise a bound
it does not have while adding a new way for the `select` and the `write` on either side of
a handler to fail half-way. The operator's way out is unchanged and runs no charter code:
`HATCH_KEY` is matched in tmux's own root key table.

## What I did NOT ship, and why

`key`, `click` and `scroll` are decoded and routed nowhere. They stay declarable, which is
what `EVENT_KINDS` already promises ("a declaration of what you HANDLE, never a promise
that it FIRES"), and a case pins that declaring one is still not an error.

- **`key`** — the harness owns the keyboard. tmux routes typing to the active pane, which
  is the harness's, so the only keystrokes a panel could see are the ones the operator typed
  into the wrong pane. There is a case pinning that a keypress never reaches a handler that
  asked for `focus`, because once a panel is reading its pane those bytes really do arrive.
- **`click`/`scroll`** — see below. This is the one thing I could not decide alone.

## The decision I am asking for — one word

**What does a click on a non-active pane mean?**

The measurement I was given: tmux routes pointer events by POSITION, and a click does not
focus the pane. So "only the active pane requests mouse" relocates the trade rather than
dissolving it, and a click on an unfocused panel arrives at that panel without moving the
operator's keyboard. Three answers:

| | behaviour | cost |
|---|---|---|
| **(a) point-to-act** | deliver where the pointer is; never move focus | the pointer and the keyboard address different panes at once |
| (b) click-to-focus | deliver, and `select-pane` to it | a misclick mid-prompt moves your typing off the harness; F12 to get back |
| (c) active-only | drop clicks on unfocused panes | the mouse is dead in the common case, since the harness is nearly always active |

**My recommendation is (a), and this PR is what makes it defensible.** (b) is hostile — the
frame exists to keep the harness the thing you type into, and a stray click must not take
that away. (c) makes the pointer useless unless you first pay (b)'s cost by hand. (a) is
also what an operator already experiences from tmux with `mouse on`: scrolling a pane you
are not focused on works.

The reason (a) was ambiguous before and is not now: the objection to point-to-act is that
the pointer becomes a second, silent focus disagreeing with the keyboard's. **After this
PR a component can see which one it is** — `focus`/`blur` tell it whether the keyboard is
on it, so it can render "you are pointing at me" and "you are typing into me" as different
states instead of guessing. Delivering `click` on top of that is then routing, not a new
question.

Worth noting either way: `[frame] mouse` still defaults off and costs the operator their
terminal's text selection when on, so under any of the three, clicks fire only for someone
who opted in with their eyes open.

Answer **a**, **b** or **c** and I will open the follow-up.

## Coordination — `overlay.py`

`overlay.py` is shared machinery, so this only **adds** to it. Five executable lines; **no
executable line edited, none deleted**:

- `overlay.py:199` — `FOCUS, BLUR = "focus", "blur"` (beside the existing `KEY, CLICK,
  SCROLL, RESIZE`)
- `overlay.py:248` — `_CSI_FOCUS = {b"I": FOCUS, b"O": BLUR}`
- `overlay.py:401-405` — three statements (plus a two-line comment) in `decode`'s existing
  CSI branch, emitting those kinds for a parameterless `CSI I`/`CSI O`

The only deletions in that file are six lines of comment prose on the `KEY, CLICK,
SCROLL, RESIZE` constant, which said `focus`/`blur` "do not exist yet" because
`focus-events` ships off in tmux — #559 made that stale.

Both sequences were previously consumed by `_CSI` and dropped, so the palette and the
pickers see exactly what they saw. `Surface.handle` returning `None` for both, with `mouse`
on and off, is pinned so this cannot start moving a palette's selection.

Nothing here touches `charter/hooks.py` (#608).

## Measurements (real tmux 3.7c, sockets killed and unlinked in one cleanup, nothing left)

```
focus-events on    pane that asked        READ b'\x1b[I' on select, b'\x1b[O' on leave
                   pane that did NOT ask  nothing, ever
                   client's own terminal  loses/regains focus -> same pair to active pane
focus-events off   pane that asked        nothing, ever
resize-window      every pane             SIGWINCH, focused or not
```

`commands_frame.conf_text` writes `set -g focus-events on` (#559), so this works on
charter's own server and does not inside an operator's existing tmux, where charter sources
no config at all. That is the "never fires" direction `EVENT_KINDS` asks for, and
`docs/frame.md` says it to the operator.

**Two defects the measurements caught before review did:**

- `termios.error` is **not** an `OSError` (`mro: (termios.error, Exception, BaseException,
  object)`). A guard written against `frame/pane.py`'s `_CANNOT_SAY` set alone would have
  turned `charter panel acme.metrics > /tmp/log` into `panel._hold` painting a refusal for
  a component that did nothing wrong.
- `TCSADRAIN` waits for the terminal to consume what is already written, and on a pane
  whose far end is not reading it never returns — it hung this branch's own suite in a
  cleanup. `close()` uses `TCSANOW`; it changes no output flag, so a drain protects nothing
  and can only wedge, in the `finally` that holds the operator's tty mode. Pinned on a
  thread with a deadline, so a regression is red rather than a hang.

## Cost

A panel whose component declares nothing charter delivers builds **no dispatcher**: the
tty keeps its mode, no `\x1b[?1004h` is written, and `_watch` sleeps exactly as before.
Charter's own four panels are in that group. A component declaring only `resize` opens no
input path either, and does not even measure the pane — pinned, because that gate's
consequence is a syscall per paint and is otherwise masked by the delivery filter.

## Behaviour change

A component that declares `events` and supplies no `on_event` now fails to load. That is a
real break for a component written against the old contract — but only for one that
declared events it was never receiving, which is the defect being fixed.
`API_VERSION` is **not** bumped: it asks "can charter honour what this provider declared",
and a component built before this field existed declares nothing charter cannot still
honour. Bumping would refuse every installed provider at once to fix a class whose events
had never fired. The reasoning is in the field's own docstring so the next reader can
disagree with it in one place.

## What an adversarial read found afterwards

The dispatcher was written, pinned and green before it was reviewed adversarially. The
review found **nine reachable defects**, all now fixed with a case that goes red without
the fix — recording them because they are the interesting part of this branch:

1. **`close()` destroyed its own restore state on the first line**, so "idempotent" meant
   "the second call is a guaranteed no-op": anything that raised or blocked in between lost
   the `tcsetattr` for good, since `_watch`'s `finally` would find `_fd` already `None`. And
   something *can* block — `open()` leaves `IXON` on, so Ctrl-S in the pane parks the flush
   inside the withdrawal. **The restore goes first now**, pinned on a thread with a deadline
   so a regression is red rather than a hang.
2. **`_watch`'s `finally` ran the fallible `close()` ahead of the infallible
   `signal.signal()`**, so a raising close leaked the SIGWINCH handler that function was
   split out to hand back. Nested now.
3. **`_CANNOT_SAY` was missing `TypeError`** — what `termios.tcgetattr` and `select.select`
   raise for a stream stand-in whose `fileno()` is not a descriptor. Same class as the
   `termios.error` miss one step out, and it sent a piped panel to `_hold` for doing
   nothing wrong. A `MagicMock` stdout reaches it.
4. **`os.read`'s bare `except OSError` folded `BlockingIOError`/`InterruptedError` into
   end-of-input**, which is *permanent*. One `O_NONBLOCK` race — anything putting stdout on
   an asyncio loop creates one — or one `SIGWINCH` landing in the read cost a component
   every later event, silently, with no `failure` to paint.
5. **`note_resize` stored an unmeasurable reading**, spending the comparison: a pty answers
   "no size" between a resize and its `TIOCSWINSZ`, and one such reading in the middle meant
   the resize never fired at all.
6. **`open()` called twice captured cbreak as `before`**, so `close()` would "restore" the
   pane to `ECHO` off, for good.
7. **`wanted()` asked truthiness**, so a handler written as an instance with a falsy
   `__bool__` silently got no dispatcher — this issue's defect with a new spelling.
8. **The handler-failure message was truncated to one line** instead of wrapped like every
   other provider failure. `" charter: <id> stopped taking events — "` is 45 characters, so
   in a 30-column side panel the exception text was *always* what got clipped — and
   `_wrap`'s own docstring says why that is not cosmetic. It also ignored the operator's
   `pad`.
9. **A composite's PART could declare events.** A part is never placed on the frame, so
   `_dispatcher` never asks `wanted()` of it: declared, and read by nothing — #607 one level
   down. Refused at registration now, naming both components; widening to deliver to
   children later costs nothing, which is `EVENT_KINDS`'s own argument about `drag`.

Two docstring claims were also overstated and are now accurate: a wedged handler **also
leaves its pane in cbreak** (the `finally` that restores it never runs), and the overlay
does not receive focus reports because its pane is split fresh for it — not because nothing
could ever deliver one.

## Verification

- `python3 -m unittest discover -s tests` green on **3.14** and on **3.12** with
  `CHARTER_*`/`CLAUDE_*`/`ANTHROPIC_*`/`TMUX*` unset (`env -i`, all 15 such variables
  dropped). Counts in the comments below.
- `tools/sweep.py` runs as this repo's CI gate on every PR; its verdict is on this PR.
- Hand-checks ran through a harness that asserts the mutation **applied** (anchor present,
  exactly once, digest moved) and that the restore **landed** (digest back to the original),
  with the restore in a `finally` — #585/#586, and the false pin that followed a script
  crashing before its restore.
- Falsification of the integration suite before trusting it: six mutations, five killed it,
  and the sixth (`_deliver`'s declared-kind filter) survived *there* because `wanted()`
  gates it — it is killed by the unit suite, and the sharper consequence it guards (a
  keypress reaching a `focus` handler) now has its own case.
- Hand-checks in total: 6 falsifying the integration suite, 2 on guards the first pass had
  left masked, 12 on the constants and failure paths the sweep is weakest on, and 9 on the
  review fixes above — 29, all killed.

### Known limit, stated rather than left to be found

An operator switching away from their whole terminal window also blurs the focused panel —
tmux forwards the client's own focus report to the active pane, measured. That behaviour is
recorded with its measurement in `events.py` but does not yet have an automated case; adding
one is a small follow-up, not a gap in the mechanism.

### DIFF
--- a/charter/frame/component.py
+++ b/charter/frame/component.py
@@ -88,6 +88,16 @@ EDGES = ("top", "bottom", "left", "right")
 #: The rule that follows, and the one thing this constant is asking of you: **degrade to
 #: "never fires", never to "fires wrongly".** A component that treats a missing ``blur``
 #: as "still focused" is correct; one that infers focus from elapsed time is not.
+#:
+#: **Which of these charter actually delivers is `frame/events.py`'s `DELIVERED`, and it
+#: is three of the six** (#607). ``focus``, ``blur`` and ``resize`` fire; ``key``,
+#: ``click`` and ``scroll`` are decoded and routed nowhere, because the harness owns the
+#: keyboard and because tmux routes a pointer by POSITION without focusing the pane it
+#: lands in — so a click would be a second focus, disagreeing with the keyboard's. Both
+#: lists are needed and neither can be the other: this one is what a provider may DECLARE
+#: and is closed by the spec, that one is what charter can carry this release. Declaring a
+#: kind charter does not deliver is not an error and does not become one — it is the
+#: paragraph above, live.
 EVENT_KINDS = ("key", "click", "scroll", "focus", "blur", "resize")
 
 #: The slices of the plane snapshot a component may declare in ``needs`` — and therefore
@@ -271,6 +281,27 @@ class Component:
     render: Callable[[Any], list[str]]
     needs: tuple[str, ...] = ()
     events: tuple[str, ...] = ()
+    #: What receives an event of a declared kind — one `frame.events.Event`, answering
+    #: truthy for "repaint me" (§4f, #607). ``None`` for a component that declares no
+    #: events, and refused alongside a non-empty ``events`` below.
+    #:
+    #: **It is handed no ``ctx``, and that is the contract rather than an omission.** A
+    #: component reads the plane at the repaint that follows, out of the one snapshot §4f
+    #: gives it; a handler that could read a second one would be the second timestamp that
+    #: doctrine exists to prevent. So a handler notes what happened and `render` draws it.
+    #:
+    #: **Nothing it returns reaches the pane.** The rows on screen are ``render``'s, which
+    #: `registry.Registry.draw` escapes, measures and clips; the return value here is read
+    #: for its truth and nothing else. That is why the containment order #472 is about
+    #: needs no second guard on this path — there is no path.
+    #:
+    #: This field was ADDED to a shipped contract without moving :data:`API_VERSION`, and
+    #: the reason is what that constant is for: it asks "can charter honour what this
+    #: provider declared", and a component built before this field existed declares nothing
+    #: charter cannot still honour. A bump would refuse every installed provider at once to
+    #: fix a class of provider — one that declared ``events`` — whose events had never
+    #: fired in the first place.
+    on_event: Callable[[Any], Any] | None = None
     #: The leaf ids this component draws inside its own pane, in the order it draws
     #: them, or ``()`` for a leaf. Charter never splits a pane (§4d), so this is how N
     #: things share one — and it is ids rather than components because the rules that
@@ -319,3 +350,27 @@ class Component:
             self, "events",
             names("events", f"component {self.id}", self.events, EVENT_KINDS))
         object.__setattr__(self, "children", tuple(self.children))
+        # After the normalisation above, so both refusals below quote the tuple charter
+        # will actually dispatch against rather than whatever shape arrived.
+        if self.on_event is not None and not callable(self.on_event):
+            raise ComponentError(
+                f"component {self.id}: on_event must be callable, not "
+                f"{contain.one_line(repr(self.on_event))}")
+        # **The pair is refused in BOTH directions, and that is this issue's whole shape.**
+        # `events` was validated here and read by nothing for two phases (#607), so a
+        # provider could declare `scroll`, pass every check, and receive nothing ever. A
+        # declaration with no handler is that defect written by the provider instead of by
+        # charter; a handler with no declaration is a callable charter would never call.
+        # Both are refused at construction, where the message names a fixable thing —
+        # which is the argument this module's docstring makes for checking anything here.
+        if self.events and self.on_event is None:
+            raise ComponentError(
+                f"component {self.id}: declares events "
+                f"{', '.join(self.events)} and supplies no on_event to receive them — "
+                f"give it one, or drop the declaration. A kind declared with nothing "
+                f"behind it is indistinguishable from one charter never fires")
+        if self.on_event is not None and not self.events:
+            raise ComponentError(
+                f"component {self.id}: supplies on_event and declares no events, so "
+                f"nothing would ever reach it — name what it handles in events "
+                f"({', '.join(EVENT_KINDS)})")
--- a/charter/frame/events.py
+++ b/charter/frame/events.py
@@ -132,7 +132,6 @@ import select
 import termios
 import time
 
-from .. import contain
 from . import overlay, pane
 from .registry import _because
 
@@ -216,27 +215,27 @@ _CANNOT_SAY = (AttributeError, OSError, TypeError, ValueError, termios.error)
 def wanted(c) -> tuple[str, ...]:
     """The kinds charter will actually deliver to component *c*, in declared order.
 
-    The intersection of what it declared and what :data:`DELIVERED` says charter carries —
-    empty for a component with no handler, which `component.Component` already refuses to
-    build alongside a non-empty ``events``, and empty for one whose every declared kind is
-    a kind charter does not fire. That second emptiness is the one that matters: a
-    component declaring only ``click`` gets no dispatcher, so its pane's terminal is left
-    exactly as it was found.
+    The intersection of what it declared and what :data:`DELIVERED` says charter carries.
+    Empty for a component whose every declared kind is one charter does not fire, and that
+    emptiness is the one that matters: a component declaring only ``click`` gets no
+    dispatcher, so its pane's terminal is left exactly as it was found.
 
-    ``c.on_event`` rather than a ``getattr`` with a default: every `Component` carries the
-    field, so a default here would be a branch nothing can reach and no test can falsify —
-    which this project deletes rather than keeps (#568). What makes it unreachable is
-    itself pinned, by the case that asks a component declaring neither for its ``on_event``
-    and gets ``None``.
+    **It does not ask whether there is a handler, and the deleted check is why this
+    docstring is long.** Two versions of that guard shipped here — a ``getattr`` default,
+    then ``c.on_event is not None`` — and the sweep proved the second one equivalent on
+    `main`. Both were: `Component` refuses ``events`` without ``on_event`` and refuses
+    ``on_event`` without ``events``, so a component with no handler has no events either
+    and the intersection below is already ``()``. A branch that cannot change an outcome is
+    not documentation of an intent — it is a second, weaker answer to a question
+    `component.Component.__post_init__` has already answered (#568's argument, and the
+    sweep's own definition of a line that should not be there).
 
-    ``is not None`` rather than truthiness, which is the same question `Component`'s own
-    validation asks one module over. A callable object defining a falsy ``__bool__`` or an
-    empty ``__len__`` — a handler written as an instance of a class rather than as a
-    function — passes construction and would silently get no dispatcher here, which is
-    #607's defect with a new spelling.
+    What makes it unreachable is itself pinned, which is the condition for deleting rather
+    than keeping: `TheDeclarationAndTheHandlerAreOneThing` asks for the refusal in both
+    directions, so the day either half is relaxed, those cases go red here rather than this
+    function silently starting to matter.
     """
-    return tuple(k for k in c.events if k in DELIVERED) \
-        if c.on_event is not None else ()
+    return tuple(k for k in c.events if k in DELIVERED)
 
 
 class Dispatcher:
@@ -514,8 +513,14 @@ class Dispatcher:
         except KeyboardInterrupt:
             raise
         except BaseException as exc:
-            self._failure = (f"{contain.one_line(self._c.id)} stopped taking events — "
-                             f"{_because(exc)}")
+            # No `contain.one_line` on the id, and that is the sweep's verdict rather than
+            # an oversight: `Component.__post_init__` holds every id to `_ID_RE`
+            # (`[a-z0-9_.]`), so there is nothing here for it to contain and the call could
+            # not change an outcome. What the exception SAID is a different matter and
+            # `_because` contains it; and `panel._component_text` contains the whole line
+            # again before any width arithmetic touches it, which is the guard that has a
+            # consequence and a case behind it.
+            self._failure = f"{self._c.id} stopped taking events — {_because(exc)}"
             self.close()
             # True, so the pane repaints and says so on the very next tick rather than
             # whenever something else happens to move the frame's version.
--- a/charter/frame/events.py
+++ b/charter/frame/events.py
@@ -115,6 +115,14 @@ not have, and it would add a new way for the `select` and the `write` on either 
 handler to fail half-way. The operator's way out is unchanged and is not charter's code:
 `overlay.HATCH_KEY` is matched in tmux's own root key table before any byte reaches this
 process.
+
+**"And nothing else" has one exception and it belongs in the sentence rather than under
+it**: a handler that never returns never reaches `panel._watch`'s ``finally`` either, so
+that pane keeps the mode :meth:`Dispatcher.open` installed — ``ICANON`` and ``ECHO`` off —
+for the life of the process. Typing into it echoes nothing. That is a WEDGED pane looking
+wedged rather than a second failure, and it is bounded to the same one pane; but a
+docstring that said the pane was simply frozen would be describing a tidier failure than
+the one this design actually has.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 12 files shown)


diazoxide:
## Local verification at `8b9ae75`

**Full suite, two environments, both counts:**

| environment | result |
|---|---|
| python 3.14.4, normal env | `Ran 7404 tests in 353.396s` — **OK** |
| python 3.12.13, `env -i` (all 15 `CHARTER_*`/`CLAUDE_*`/`ANTHROPIC_*`/`TMUX*` dropped) | `Ran 7404 tests in 351.452s` — **OK** |

**Real tmux 3.7c**, `tests/test_frame_focus_reaches_a_component.py` — 6 cases, a real
server, a real attached client on a real pty, a real installed `.dist-info`, three real
`charter panel <component-id>` processes. Sockets killed and unlinked in one cleanup;
`ls /tmp/tmux-$(id -u)/ | grep focus-integ` finds nothing afterwards, and the measurement
harness's own socket directory is removed by its own `finally`.

**Hand-checks: 29 mutations, 29 killed.** Every one ran through a harness that refuses to
report unless the mutation is proven to have APPLIED (anchor present, present exactly once,
file digest moved) and the restore is proven to have LANDED (digest back to the original),
with the restore in a `finally` — #585/#586, and the false pin that followed a script
crashing before its restore.

Two rounds are worth naming because they changed the branch:

- **Falsifying the tmux integration suite before trusting it.** Six mutations; five killed
  it. The sixth — deleting `_deliver`'s declared-kind filter — survived *there*, and the
  reason is a second unpinned line hiding the consequence: `wanted()` gates the dispatcher,
  so the negative component (`acme.deaf`, declaring only `click`) never gets one and the
  filter is unreachable from that suite. It is killed by the unit suite. Rather than call it
  equivalent I asked what it actually guards and found the sharper case — once a panel is
  reading its pane, an operator typing into that pane produces `key` events, and a handler
  that asked for `focus` must not be handed them. That now has its own case.
- **The two masked cost guards.** `note_resize`'s early return and `_FROM_INPUT`'s gate both
  produce identical *output* either way; what they buy is a syscall per paint and an
  untouched terminal. Both are pinned on the cost rather than the result.

**Not pinned, stated instead:** `_CHUNK = 4096` bounds how many events one tick can carry
and nothing observable changes at another value — `overlay.decode` holds partial sequences
back regardless. It is a tuning constant, not a guard.

diazoxide:
## The sweep gate's survivors, answered — `07f0b6b`

The gate is **reporting-only**, so it went green at `8b9ae75` with **8 survivors** under
it. A survivor is a guard with no test behind it or a line that should not be there, so
none of them is left in the report. At `07f0b6b`:

```
gate: 8 unpinned, 2 in masked cluster(s)  ->  gate: 1 unpinned, 0 in masked cluster(s)
pinned: 43  ->  49
```

**Two were deleted, because that is what an equivalent mutant earns:**

- `wanted()`'s handler check. `Component` refuses `events` without `on_event` and
  `on_event` without `events`, so a component with no handler has no events either and the
  intersection is already `()`. **Two spellings of that guard shipped on this branch** — a
  `getattr` default, then `is not None` — and both were a second, weaker answer to a
  question `__post_init__` had already given. What makes it unreachable is itself pinned,
  which is the condition for deleting rather than keeping.
- `contain.one_line` on the component id in `_deliver`. `_ID_RE` holds every id to
  `[a-z0-9_.]`, so there was nothing there to contain. `contain` became an unused import
  and went with it.

**Five were pinned, each with a consequence the standing cases could not see:**

- `component.py`'s containment of `repr(self.on_event)` — a provider's object answers
  `repr` with whatever it likes, and a refusal that can forge a second line is worse than
  the value it refuses.
- `close()`'s `fd is None` guard — without it a resize-only component's pane, or one that
  is not a terminal, gets `\x1b[?1004l` written into it having never asked.
- **`poll()`'s two catches, which the sweep itself flagged as masking each other.** The
  `os.read` one is the half that matters: closing a pty's far end answers `b""` on macOS
  and raises `OSError: [Errno 5]` on Linux, so the standing case reached one arm only — a
  suite green twice over on this machine would still have been red on every Linux
  operator's, which is the measurement `palette._reader`'s docstring records costing a red
  CI once already.
- `panel`'s containment **before** the width arithmetic (#472). Neither escaping nor width
  could show it — `_fit` escapes *and* truncates, so both questions are answered whichever
  order the two run in, which is precisely why it survived with two cases standing over it.
  What the order decides is where `_wrap` breaks, and getting it wrong **drops text**.
  Measured at width 32 on `'stopped: ' + '\x1b[2J' * 3 + ' ENDMARK'`:

  ```
  tui.width(raw) = 17        tui.width(contained) = 38
  contain first  -> ('stopped: \\x1b[2J\\x1b[2J\\x1b[2J', 'ENDMARK')
  contain second -> ('stopped: \\x1b[2J\\x1b[2J\\x1b[2J …',)
  ```

  `ENDMARK` is gone, and in a real failure that tail is the exception's own message.

**One is declared rather than changed.** `_CHUNK = 4096` bounds how many events one tick
can carry; any positive value behaves identically because `overlay.decode` holds partial
sequences back regardless. It is a tuning constant, not a guard, and it is the single
remaining `unpinned`.

**The 3 UNRESOLVED are timeouts on a loaded runner, and a timeout is not a pin.** All three
are `drop-conjunct` on the new refusals in `component.py`. Resolved locally — both
conjuncts of all three, six mutations, six killed.

## Final state at `07f0b6b`

| | |
|---|---|
| python 3.14.4, normal env | `Ran 7409 tests` — **OK** |
| python 3.12.13, `env -i` | `Ran 7409 tests` — **OK** |
| CI 3.11 / 3.12 / 3.13 / 3.14 | success |
| sweep gate | success — 49 pinned, 1 declared tuning constant |
| real tmux 3.7c integration | 6 cases, green; no socket left behind |
| hand-checks, cumulative | **40 mutations, 40 killed** |

PR #630, opened 2026-08-28
A green sweep says "no survivors", and one too big to finish says that instead (#617)

### LINKED ISSUE #617: The sweep gate times out at 60 minutes on a 334-line diff, and that is before --enforce
Found by #608, whose diff hit the ceiling twice.

## What happened

`.github/workflows/sweep.yml` ran `09:03:44Z→10:04:20Z` and `10:17:02Z→11:17:21Z` — the full `timeout-minutes: 60` both times, reported as **`cancelled`**.

The diff was **334 added lines**, which is **62 mutations** once #615's `retune-string` operator counts string constants. The job's header sizes itself for roughly half that.

## Why it matters now rather than later

The gate **blocks nothing today** — `--enforce` is deliberately absent, and that was the right call. So a timeout costs a report, not a merge.

But the whole point of shipping it reporting-first was to make its numbers visible on real branches before turning it on, and **a run that times out produces no numbers at all**. A gate that silently reports nothing on the largest diffs is reporting on exactly the branches least likely to be fully pinned.

And #615 was explicit that `UNRESOLVED` is its own outcome for precisely this reason: a timeout is not a red and not a survivor, it is *no verdict*. A whole-job timeout is that failure one level up — the job itself returns no verdict, and `cancelled` reads like infrastructure noise rather than a missing answer.

## The arithmetic

Measured on `main` earlier this week: selection map 118s (cached), unmutated baseline 262s, ~24s per mutation at `--jobs 3` on a loaded box. Phase 2's PRs generated 34–38 mutations, so `262 + 37×24 ≈ 20 min` — comfortably inside the hour, which is what the job was sized against.

`retune-string` changed that input. It is the operator family that closes the blind spot every hand-check this week was compensating for, so it is not something to give back — but it roughly doubles the mutation count on a diff that touches string constants, and guard tables are nothing *but* string constants.

## Options, none of them obviously right

- **Raise `timeout-minutes`.** Cheapest, and postpones the question rather than answering it.
- **Cap mutations per run**, reporting what was dropped. #565's spec is explicit that silent truncation reads as "covered everything" — so this only works if the cap is loud.
- **Shard by file** across matrix jobs. The selection map is cached and content-addressed, so the marginal cost of a second job is the baseline run, not the map.
- **Scope `retune-string` further.** It is already restricted to read positions; the measurement showed mutating every string took `charter/` from 7,006 to 14,801 mutations, nearly all prose.

Worth deciding before `--enforce`, because at that point a timeout stops being a missing report and becomes a blocked merge.

## Related
- **#569** — the harness's remaining gaps; the renderer row is still thin (106 of 168 mutations survive) and is named there.

diazoxide:
Closes #617.

The sweep gate had two defects. The quieter one was worse.

## 1. A green check was not "no survivors"

The gate reports and blocks nothing — `--enforce` is deliberately absent and **stays absent** — but a job that blocks nothing still has to *say* something, and this one concluded `success` whatever it found. #621's branch went green with **eight survivors** under it.

**GitHub's conclusion vocabulary cannot express the difference, so the answer is not a conclusion.** A `run:`-driven job concludes `success` or `failure` and nothing else. The one conclusion that means *I looked, here is the count, this is not a pass or a fail* — `neutral` — exists only on the Checks API, which needs `checks: write` and a check run created by hand. `sweep.yml`'s header already refuses that in as many words: this job runs the repository's own code against mutated copies of it and is the last place to hold a writable token. Trading that for a prettier icon is not a trade worth making.

So the **check's name** carries it. A job name may interpolate `needs.<job>.outputs.*`, so a final one-step job renders on the pull request as one of:

```
deletion sweep / no survivors
deletion sweep / 8 survivors
deletion sweep / no verdict: 2 of 3 shards did not report
deletion sweep / no verdict: the sweep never sized itself
```

**All of them green.** Distinguishing the states is not the same as failing on them. This is live on this pull request already — the check below is literally named `no survivors`.

Every survivor is also an **annotation on the line it is about** — in the margin of the guard, on the Files-changed tab, carrying the question it failed. Platform-deferred survivors are `::notice` rather than `::warning`, because the gate never fails on one of those and should not shout about one either. GitHub draws ten annotations of a level per step and then stops silently, so the tenth slot is reserved for the line that says how many were not drawn.

## 2. It ran out of time

#608's 62 mutations were cancelled at the hour twice; #626's 78 three times, each reaching about two thirds of its plan. The job was sized against Phase 2's 30–52, and `retune-string` roughly doubles the count on a diff full of string constants.

**Nothing is dropped to make a branch fit.** A cap on the mutation count would read, downstream, as "covered everything" — the silent truncation the spec refuses. What is capped is the **fan-out**. A `plan` job counts the mutations in one `ast` pass over the diff (no test runs), sizes the sweep from the real number, and the shards run in parallel. Past even the ceiling the plan emits a warning annotation and still sweeps every mutation.

**Dealt one mutation at a time, and not split by file** — the correction the numbers force on the issue's own proposal. Both over-budget diffs were dominated by a *single new file*, so a job-per-file hands one job all 78 of `gitconfig.py`'s mutations and the others nothing. The masked-cluster bucket survives round-robin because `--verdict` classifies the **merged** results, not one shard's; there is a test for exactly that.

### Measured on `ubuntu-latest`, not on a workstation

| | measured |
|---|---|
| checkout at `fetch-depth: 0` + interpreter | 3 s |
| selection map, traced | **250 s** (`Size the sweep` step, 4 m 13 s end to end) |
| selection map, restored from cache | under 5 s |
| sandbox clone | ~15 s |
| unmutated baseline | ~240 s — the same 7,693 tests `test.yml` runs in 4 min |
| per mutation | **58 s**, read off #626: ~53 of 78 inside the hour, over a 515 s no-cache fixed cost |

`SHARD_BUDGET` 40 min, `SHARD_FIXED` 12 min → **28 mutations a shard**. 62 → 3 shards, 78 → 3 shards, 28 → 1. `MAX_SHARDS` 8, so the ceiling is 224 and past it the plan says so out loud. 78 over three shards is 26 each: 515 + 26×58 ≈ **34 min** with no cache, under 30 with one — inside the budget and well inside the hour backstop.

The map is warmed once by `plan` and restored by every shard through `actions/cache`. That the map was cached was the issue's premise for sharding, and it **was not true**: `sweep.yml` had no cache step at all, so every run rebuilt it. `SHARD_FIXED` is still the no-cache figure and rounded up past it, because a fork's pull request cannot write that cache and a budget that only holds on a hit fails on exactly the runs nobody is watching.

## What this rests on

One distinction the tool did not make: **a sweep that ran and found nothing, and a sweep that did not run, are different answers.**

- A branch with nothing under the swept paths now writes an **empty** result set rather than none — a file that is absent is indistinguishable from a job that was cancelled. (This branch is that case, which is why its check says `no survivors` and not `no verdict`.)
- A result file that will not parse is a shard that did not report. There is no third reading of a truncated upload.
- An empty shard count from a failed `plan` job reads as *the sweep never sized itself*, not *no shards were needed* — and not as `1 of 1 did not report` either, because a denominator nobody computed is not a denominator.
- `as_json` now writes `"naming": null` rather than `[]` when the evidence pass never ran. Caught by the round-trip test: "nothing measured executes this file" and "one module executes it and does not name the symbol" are different findings, and flattening them reported the wrong one.

## The sweep swept the gate that reports it

119 mutations against `tools/sweep.py`. It found two lines of its own that no test went red without, both now pinned and both hand-checked against a green baseline with a sha256-verified restore:

- **The unsharded path had stopped being tested.** Forcing `if shard is not None` to always-true left every test green — once the shard tests existed, nothing in the suite ran a sweep *without* one. That is how anyone uses `tools/sweep.py` from a terminal.
- **The sizing arithmetic could divide by zero.** The `max(1, …)` in `per_shard` is what stops a fixed cost that outgrew the budget from making mutations-per-shard zero. Deleting it changed no answer today and would have made the plan job raise tomorrow — no plan, no shards, no numbers, which is the #617 failure arriving out of the arithmetic written to prevent it.

A third survivor is not a finding about this code: it is a string inside a **type annotation**, which `from __future__ import annotations` means the interpreter never evaluates, so no test can ever go red without it. That is a blind spot in `retune-string`'s scoping rather than a guard, and it is filed as **#632** instead of being worked around here.

## Not regressed

Survivors still sort into unpinned / masked cluster / platform-deferred / unresolved / not-applied; platform-deferred still never fails; `UNRESOLVED` is still its own outcome; a survivor still reports what the covering tests assert; every mutation still asserts it applied; sandboxes are still private per run. `--enforce` is still the single flag, now on the `collect` step, and a test refuses it in any non-comment line of the workflow.

## Verification

- **270 cases in `tests/test_sweep.py`** (from 193), literals pinned rather than recomputed from the constants under test. Eleven of them read `.github/workflows/sweep.yml` as a tree through `tests/test_workflows.py`'s own loader, because the half #617 was about is whether anything ever hears the answer.
- **Full suite in two environments** — reported in a comment below.
- `actionlint` clean. `tests/test_workflows.py` parses the new four-job graph; it caught an unquoted `: ` in a job name.
- **This pull request is its own end-to-end test.** The workflow it changes runs on it: `Size the sweep` → `Sweep shard 1 of 1` → `Add up what the shards found` → a check named `no survivors`.

### DIFF
--- a/.github/workflows/sweep.yml
+++ b/.github/workflows/sweep.yml
@@ -5,6 +5,13 @@ on:
 
 # Read-only, declared rather than inherited. This job runs the repository's own code
 # against mutated copies of it, which is the last place to hold a writable token.
+#
+# It is also the reason the answer below is carried by a check's NAME rather than by its
+# conclusion. GitHub gives a `run:`-driven job two conclusions, `success` and `failure`;
+# the third one this sweep needs — `neutral`, meaning "I looked, here is the count, this
+# is not a pass or a fail" — exists only on the Checks API, which needs `checks: write`
+# and a check run created by hand. That is a writable token, in this job, for a prettier
+# icon. Not a trade worth making, so `verdict` below is a job whose name is the answer.
 permissions:
   contents: read
 
@@ -19,23 +26,33 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  sweep:
-    name: Every guard this branch adds is a line a test goes red without
+  # ------------------------------------------------------------------------------------
+  # 1. How big is this? Answered before anything is spent on answering it.
+  # ------------------------------------------------------------------------------------
+  plan:
+    name: Size the sweep
     runs-on: ubuntu-latest
     # Stage C of `docs/superpowers/specs/2026-08-27-deletion-sweep-harness.md`.
     #
     # **Scoped to the lines this branch ADDED**, never to the tree. The whole-tree number
     # is stage B — roughly 5,672 mutations and fourteen hours — and it is not a
-    # pull-request check and never will be. Measured on real branches from this phase: 30
-    # to 52 mutations each, a 262 s unmutated baseline, and about 24 s a mutation, so a
-    # gate run is twenty-odd minutes.
+    # pull-request check and never will be.
     #
-    # **It reports and blocks nothing.** `--enforce` is the flag that changes that and it
-    # is deliberately absent here. The spec's staging argument — a gate whose baseline
-    # nobody has seen gets disabled the first time it is inconvenient — applies to this
-    # job's own credibility too. Turn it on by adding `--enforce` to the step below, once
-    # the numbers on a few real branches have been read and believed.
-    timeout-minutes: 60
+    # This job used to be the whole workflow, sized against the 30 to 52 mutations Phase 2
+    # produced. #608 brought 62 and #626 brought 78, and five runs across the two were
+    # cancelled at the cap having measured about two thirds of the plan (#617). So the
+    # count is now measured rather than assumed: one `ast` pass over the diff, no test
+    # runs, and `tools/sweep.py` decides how many jobs it needs from the real number.
+    #
+    # **Nothing is dropped to make it fit.** A cap on the mutations would read, to every
+    # reader downstream, as "covered everything" — the spec says exactly that about silent
+    # truncation. What is capped is the fan-out, and when a diff is past even that the
+    # plan says so out loud instead of quietly measuring less.
+    timeout-minutes: 30
+    outputs:
+      mutations: ${{ steps.size.outputs.mutations }}
+      shards: ${{ steps.size.outputs.shards }}
+      matrix: ${{ steps.size.outputs.matrix }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
@@ -52,7 +69,19 @@ jobs:
           # #508's entire defect, out of the sweep's reach. `sweep.reach()` says so in the
           # report rather than letting a smaller sweep read like a clean one.
           python-version: "3.12"
-      - name: Sweep the guards this branch adds
+      # The selection map is one trace of the whole suite — 350 s measured, and the
+      # largest fixed cost a shard pays. It is content-addressed on the tree it measured,
+      # so this step saves it once here and every shard below restores it instead of
+      # measuring it again. A fork's pull request cannot write this cache and will simply
+      # measure its own; `SHARD_FIXED` in `sweep.py` is the no-cache figure for that
+      # reason, because a budget that only holds when the cache hits is a budget that
+      # fails on exactly the runs nobody is watching.
+      - uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        with:
+          path: ${{ runner.temp }}/sweep/cache
+          key: sweep-map-${{ hashFiles('charter/**/*.py', 'tests/**/*.py') }}
+      - name: How many mutations, and how many jobs they need
+        id: size
         env:
           # Through the environment and not interpolated into the script. A `${{ }}` inside
           # `run:` is textual substitution before bash ever sees it, which is how a value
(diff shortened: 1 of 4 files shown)


diazoxide:
## Verification at `72f58ff`

### The workflow ran on itself, end to end

```
Size the sweep                :: success   (4m13s — 250s of it tracing the selection map)
Sweep shard 1 of 1            :: success
Add up what the shards found  :: success
no survivors                  :: success   ← the check's NAME is the answer
test (3.11 / 3.12 / 3.13 / 3.14) :: success
```

`gh api repos/diazoxide/charter/commits/72f58ff7c72eceb829681de7f88daff2dc7dea4a/check-runs` — nine checks, nothing failed, nothing cancelled.

That last row is the whole of the first half of #617. Before this branch the same run would have shown a green tick beside a name that says the same thing on every branch, whatever it found.

It says `no survivors` and not `no verdict` because of the smallest change here: this diff touches nothing under `charter/`, so the shard has nothing to sweep — and it now writes an **empty** result set rather than no file at all. A missing file and an empty one were indistinguishable before, and one of them means "cancelled".

### The shard's artifact merges on another machine

The `[]` that GitHub's runner uploaded, downloaded and merged by the local tool:

```
$ python3 tools/sweep.py --verdict <downloaded> --shards 1
merged 0 result(s) from 1 of 1 shard(s)
gate: 0 unpinned, 0 in masked cluster(s), 0 platform-deferred, 0 unresolved, 0 not applied, 0 pinned
gate: no survivors
gate: reporting only — nothing here blocks. Pass --enforce to make it.
```

Same answer, other machine — which is what "one question answered on several machines" has to mean.

### Full suite, two environments, both at `72f58ff`

| | |
|---|---|
| macOS, CPython **3.12.13** | **Ran 7713 tests — OK** (472 s) |
| macOS, CPython **3.14.4**, `CHARTER_*` / `CLAUDE_*` / `ANTHROPIC_*` / `TMUX*` unset | **Ran 7713 tests — OK** (442 s) |
| Linux, CPython 3.11 / 3.12 / 3.13 / 3.14 (CI) | green |

The clean environment was *verified* empty rather than assumed — the run is preceded by a check that no `CHARTER`/`CLAUDE`/`ANTHROPIC`/`TMUX` name survives into `os.environ`, which printed `leaked: []`.

`tests/test_sweep.py` is at **271 cases** (from 193). `actionlint` clean.

### The sweep swept the gate that reports it

`tools/sweep.py --gate --path tools` — **127 mutations** against the file this whole change is about. Two of its own lines had no test going red without them:

- **The unsharded path had stopped being tested.** Forcing `if shard is not None` to always-true left every test green: once the shard tests existed, nothing in the suite ran a sweep *without* one — which is how anyone uses this tool from a terminal.
- **The sizing arithmetic could divide by zero.** The `max(1, …)` in `per_shard` stops a fixed cost that outgrew the budget from making mutations-per-shard zero. Deleting it changed no answer today and would have made the plan job raise tomorrow: no plan, no shards, no numbers — the #617 failure arriving out of the arithmetic written to prevent it.

Each was hand-checked against a **green unmutated baseline** with a **sha256-verified restore** (#579's rule, #586's):

```
BASELINE (unmutated):  rc=0  Ran 13 tests  OK
MUTANT  `shard is not None` -> `True`:   rc=1  FAILED (errors=1)     restored 8d62e1a25ad1 == 8d62e1a25ad1
MUTANT  `shard is not None` -> `False`:  rc=1  FAILED (failures=1)   restored 8d62e1a25ad1 == 8d62e1a25ad1

BASELINE (unmutated):  rc=0  Ran 11 tests  OK
MUTANT  per_shard() without its max(1, …):  rc=1  FAILED (failures=1)  restored 8d62e1a25ad1 == 8d62e1a25ad1
```

…and the re-run sweep now reports all three of those mutants `pinned` on its own, which is the check that matters:

```
[1/127] pinned  tools/sweep.py:1630 [disable-branch] shard is not None
[2/127] pinned  tools/sweep.py:1630 [disable-branch] shard is not None
[8/127] pinned  tools/sweep.py:2094 [unclamp] max(1, (SHARD_BUDGET - SHARD_FIXED) // SECONDS_A_MUTATION)
[9/127] pinned  tools/sweep.py:2094 [unclamp] max(1, (SHARD_BUDGET - SHARD_FIXED) // SECONDS_A_MUTATION)
```

A third survivor is `"Pair"` inside a return annotation — unevaluated under `from __future__ import annotations`, so no test can ever pin it. Filed as **#632** rather than worked around; unquoting one forward reference would fix one line and none of the others.

**The self-sweep's first run was cancelled at 40 of 119**, on a box carrying four other agents' sweeps at load average 25. It is running again and still going. That is the same failure this pull request is about, one level down, and it is worth saying rather than rounding off: an incomplete sweep is **no verdict**, not a clean one — which is now the answer the tool gives instead of silence.

diazoxide:
## What the sweep found on the branch that changes the sweep

Adding to the verification above. `tools/sweep.py --gate --path tools` — **127 mutations** against the file this whole change is about. Three of its own lines had no test going red without them.

**1. The unsharded path had stopped being tested.** Forcing `if shard is not None` to always-true left every test green: once the shard tests existed, nothing in the suite ran a sweep *without* one — which is how anyone uses this tool from a terminal.

**2. The sizing arithmetic could divide by zero.** The `max(1, …)` in `per_shard` stops a fixed cost that outgrew the budget from making mutations-per-shard zero. Deleting it changed no answer today and would have made the plan job raise tomorrow: no plan, no shards, no numbers — the #617 failure arriving out of the arithmetic written to prevent it.

**3. The annotation escaping was asserted by absence.** The tests checked that a newline was *gone* from a workflow command, never that it had become `%0A`. So all three encodings survived:

```
[41/127] SURVIVED  tools/sweep.py:2253 [retune-string] "%0D"   (1 module(s), then all 7712)
[42/127] SURVIVED  tools/sweep.py:2253 [retune-string] "%0A"   (1 module(s), then all 7712)
[43/127] SURVIVED  tools/sweep.py:2253 [retune-string] "%25"   (1 module(s), then all 7712)
```

A wrong encoding there breaks nothing visibly — the runner prints a mangled message, which is the kind of wrong nobody files and nobody can read either. The bytes are pinned now, and pinning them pins the **order** too: `%` has to be escaped first, or the `%` of `%0A` is escaped a second time and `a\nb` comes out as `a%250Ab`. That last one is a hazard the sweep's own operators cannot generate, and the same literal assertion catches it.

Every one hand-checked against a **green unmutated baseline** with a **sha256-verified restore**, eight mutants across the three findings, **zero unpinned**:

```
BASELINE TheWorkflowAsksTheToolAndNotTheOtherWayAround: rc=0 Ran 13 tests  OK
   `shard is not None` -> True     rc=1 FAILED (errors=1)     restore e560780c82d2==e560780c82d2  PINNED
   `shard is not None` -> False    rc=1 FAILED (failures=1)   restore e560780c82d2==e560780c82d2  PINNED
BASELINE TheSweepIsSplitAcrossMachinesAndNothingIsDropped: rc=0 Ran 11 tests  OK
   per_shard() without max(1, …)   rc=1 FAILED (failures=1)   restore e560780c82d2==e560780c82d2  PINNED
BASELINE EverySurvivorLandsOnTheLineItIsAbout: rc=0 Ran 13 tests  OK
   "%0D" retuned                   rc=1 FAILED (failures=1)   restore e560780c82d2==e560780c82d2  PINNED
   "%0A" retuned                   rc=1 FAILED (failures=1)   restore e560780c82d2==e560780c82d2  PINNED
   "%25" retuned                   rc=1 FAILED (failures=1)   restore e560780c82d2==e560780c82d2  PINNED
   % escaped last                  rc=1 FAILED (failures=1)   restore e560780c82d2==e560780c82d2  PINNED
```

### And the third way a sweep lies, reproduced live — against my own hand-check

The first attempt at that table reported the `"%25"` mutant as **still green**, and the file on disk was provably correct: `git diff` clean, the line reading `"%25"`. The `.pyc` was the liar.

```
pyc records mtime=1787952595 size=153465
src has   mtime=1787952595 size=153465
```

CPython validates a cached `.pyc` on source **size and mtime truncated to whole seconds**, and nothing else. `retune-string` preserves length by construction — that is the operator's entire design — so a same-second restore of an equal-length mutation is byte-different and `(mtime, size)`-identical. The interpreter reused the mutant's bytecode and the *next* verdict was about the wrong bytes.

This is §4's third way a sweep lies, and **the harness is already immune to it**: `Sandbox.run` sets `PYTHONDONTWRITEBYTECODE=1`, with a comment documenting the exact measurement. Only my ad-hoc script was missing it. Every hand-check above was then re-run from scratch under `PYTHONDONTWRITEBYTECODE=1` and `python -B` with the caches cleared, which is the table you see.

Worth writing down because it is a rule about *reviewing* this tool, not only about running it: a hand-check of a length-preserving mutation is not evidence unless bytecode caching is off.

### The fourth survivor is not a finding about this code

`"Pair"` inside a return annotation, unevaluated under `from __future__ import annotations`, so no test can ever pin it. Filed as **#632** rather than worked around — unquoting one forward reference would fix one line and none of the others.

diazoxide:
### A fourth finding, and the reason to keep the sweep running

`[53/127]` and `[54/127]` — the **reserved annotation slot had no case on its boundary**:

```
[53/127] SURVIVED  tools/sweep.py:2312 [collapse-ifexp] ANNOTATION_CAP - 1 if n > ANNOTATION_CAP else ANNOTATION_CAP
[54/127] SURVIVED  tools/sweep.py:2312 [shift-boundary] n > ANNOTATION_CAP
```

When a level is over budget the tenth slot goes to the line that says how many were not drawn — otherwise that line is the first thing the cap eats. Every test had either far fewer than ten findings or far more, so **"always ten" and "always nine" both passed**, and "always ten" is the one that silently loses the note. There are cases at nine, ten and eleven now, and the boundary is pinned in both directions:

```
BASELINE EverySurvivorLandsOnTheLineItIsAbout: rc=0 Ran 14 tests  OK
   collapse -> always ANNOTATION_CAP       rc=1 FAILED (failures=4)  restore e560780c82d2==e560780c82d2  PINNED
   collapse -> always ANNOTATION_CAP - 1   rc=1 FAILED (failures=1)  restore e560780c82d2==e560780c82d2  PINNED
   boundary shifted to >=                  rc=1 FAILED (failures=1)  restore e560780c82d2==e560780c82d2  PINNED
```

That is the third finding in a row where the test asserted the *absence* of something rather than the *value* of it — a newline that was gone rather than encoded, a clamp that changed no answer today, a cap with no case on its edge. Which is the harness's own thesis, arriving about the harness.

**Running total on this branch: 127 mutations, five survivors.** Four were real and are pinned; the fifth is #632's annotation string, which no test can pin. The sweep is still running — under load average 30 from four sibling agents' sweeps on the same box — so **this is not a complete verdict and is not offered as one.** Whatever it finds next gets the same treatment.

Issue #632, opened 2026-08-28
`retune-string` mutates strings inside annotations, which PEP 563 never evaluates


diazoxide:
Found by the self-sweep on #630 (`tools/sweep.py --gate --path tools`), which reported this survivor against all 7,692 tests:

```
[8/119] SURVIVED  tools/sweep.py:1626 [retune-string] "Pair"   (1 module(s), then all 7692)
```

Line 1626 is a return annotation:

```python
def sweep(...) -> tuple[list[Result], list["Pair"]]:
```

Every module in this repository begins `from __future__ import annotations`, so that annotation is **never evaluated** — it is stored as source text and nothing reads it:

```
>>> type(sweep.sweep.__annotations__["return"]).__name__
'str'
>>> sweep.sweep.__annotations__["return"]
"tuple[list[Result], list['Pair']]"
```

Retuning `"Pair"` to some other four-letter string changes only the contents of a string nothing evaluates. It is an equivalent mutant by construction, and no test can ever pin it.

## Why it gets picked up

`read_positions` treats a `Subscript`'s slice as a read position, which is right for `d["components"]` and wrong for `list["Pair"]`:

```python
if isinstance(node, ast.Subscript):
    take(node.slice)
```

A forward reference in an annotation is exactly that shape.

## Why it is worth fixing rather than living with

It is not one line. `tools/sweep.py` alone carries four (`Mutation`, `Pair`, `Sandbox`), and the pattern is ordinary in this tree — any branch that adds a function whose signature forward-references a class it defines below gets a phantom survivor on the gate.

That is the specific kind of noise the harness cannot afford. The gate's whole argument is that a survivor is a finding worth a reviewer's minute; a survivor that provably *cannot* be pinned teaches the reader to skim the list, and there is deliberately no suppression list to put it on.

## The shape of a fix

Skip string constants that sit inside an annotation — a function's `returns` or an argument's `annotation`, an `AnnAssign`'s `annotation` — when the module carries `from __future__ import annotations`. Under PEP 563 that is provably unevaluated; without the future import a string annotation is still not evaluated at definition time, so the same argument mostly holds, but the narrow rule is the one with no judgement in it.

`read_positions`' docstring asks for a measurement behind any change to its predicate, and this one should have the same: how many read positions it removes across `charter/`, and how many mutations that is.

## Not blocking

[...]

PR #635, opened 2026-08-28
Recover #630's final commit, lost to a merge race

diazoxide:
**My error, twice over.** I merged #630 at head `72f58ff` while its author was still working; `bc8b724` landed afterwards and never reached `main`. This is the same mistake I made with #604 earlier — and #613 was that recovery.

## What was missing

```
 tests/test_sweep.py                                | 31 +++++++++++++++
 docs/news/unreleased-a-green-sweep-is-not-no-...md | 17 ++++++--
```

Four lines the sweep found unpinned **in the gate's own new code**, each hand-checked against a green unmutated baseline with a sha256-verified restore, under `PYTHONDONTWRITEBYTECODE=1` — plus the news entry saying what they were.

## Why losing these in particular would have been bad

The gate exists to make "this branch has unpinned guards" visible. Merging it with four unpinned lines of its own, reporting `no survivors`, is precisely the failure it was built to prevent — and it would have said so about itself, wrongly, on every PR from then on.

## The rule I keep relearning

**A PR head is not a finished state until its author says so.** Green checks are not the signal; the author's "done" is. I have now paid for that twice in one session, so I am writing it here rather than in a summary that scrolls away.

`tests.test_sweep` — 273 tests, OK.

### DIFF
--- a/docs/news/unreleased-a-green-sweep-is-not-no-survivors.md
+++ b/docs/news/unreleased-a-green-sweep-is-not-no-survivors.md
@@ -91,7 +91,7 @@ kind of pass.
 
 ## The sweep swept the gate that reports it
 
-119 mutations, against a change whose whole subject is the sweep. It found two lines of
+127 mutations, against a change whose whole subject is the sweep. It found four lines of
 its own that no test went red without.
 
 **The unsharded path had stopped being tested.** Forcing `if shard is not None` to
@@ -105,7 +105,20 @@ zero. Deleting it changed no answer with today's constants and would have made t
 job raise with tomorrow's — no plan, no shards, no numbers, which is the failure this
 whole change exists to prevent, arriving out of the arithmetic written to prevent it.
 
-Both are pinned now. A third survivor is not a finding about this code at all: it is a
+**And the annotation escaping was asserted by absence.** The tests checked that a newline
+was *gone* from a workflow command, never that it had become `%0A` — so retuning `%0D` or
+`%0A` or `%25` left the suite green. A wrong encoding there does not break anything
+visibly: the runner prints a mangled message, which is the kind of wrong nobody files and
+nobody can read either. The bytes are pinned now, and pinning them pins the order too,
+because `%` has to be escaped first or the `%` of `%0A` gets escaped a second time.
+
+**And the reserved annotation slot had no case on its boundary.** When a level is over
+budget the tenth slot goes to the line that says how many were not drawn — otherwise that
+line is the first thing the cap eats. Every test had either far fewer than ten findings or
+far more, so "always ten" and "always nine" both passed, and "always ten" is the one that
+loses the note. There is a case at nine, ten and eleven now.
+
+All four are pinned now. A fifth survivor is not a finding about this code at all: it is a
 string inside a *type annotation*, which `from __future__ import annotations` means the
 interpreter never evaluates, so no test can ever go red without it. That is a blind spot
 in the string operator's scoping rather than a guard, and it is filed as one (#632)
--- a/tests/test_sweep.py
+++ b/tests/test_sweep.py
@@ -2387,6 +2387,25 @@ class EverySurvivorLandsOnTheLineItIsAbout(unittest.TestCase):
         self.assertIn("5 of these 14 are not shown", said[-1])
         self.assertIn("all 14", said[-1])
 
+    def test_exactly_the_cap_is_drawn_whole_and_one_more_costs_a_slot(self):
+        """The boundary the reserved slot turns on, found unpinned by the sweep.
+
+        At exactly ten there is nothing to announce, so all ten are findings. At eleven
+        the note has to fit inside the same ten, so nine are. Without a case sitting on
+        the boundary, "always ten" and "always nine" both pass — and "always ten" is the
+        one that loses the note, which is the only line that distinguishes *these ten*
+        from *these ten of twenty-two*.
+        """
+        def drawn(n):
+            gate = sweep.classify([_result("survived", path=f"charter/{i}.py")
+                                   for i in range(n)])
+            said = sweep.annotations(gate)
+            return sum(1 for s in said if "file=" in s), len(said)
+
+        self.assertEqual(drawn(9), (9, 9))       # under the cap: no note
+        self.assertEqual(drawn(10), (10, 10))    # exactly the cap: still no note
+        self.assertEqual(drawn(11), (9, 10))     # one over: nine findings and the note
+
     def test_the_cap_is_a_levels_budget_and_not_one_kind_of_findings(self):
         """Masked clusters, lone survivors and unmeasured mutations are all warnings, so
         they share one budget of ten. Capping each family at ten instead would emit thirty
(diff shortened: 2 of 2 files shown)


diazoxide:
Thank you for catching this — and yes, `bc8b724` is exactly what was missing. I diffed this branch against it and my four files (`tools/sweep.py`, `tests/test_sweep.py`, `.github/workflows/sweep.yml`, the news entry) are identical, so the recovery is complete for what I had pushed.

**I have added one commit on top** (`964bba7`), because the self-sweep was still running when #630 merged and it found a fifth unpinned line.

## `[56/127]` — the wall clock was never read back

```
took = f"{elapsed / 60:.1f} min" if elapsed is not None else "merged from its shards"
```

Collapsing that to the second arm leaves **all 273 tests green**. Nothing asserted the minutes, so every page — measured or merged — could have said "merged from its shards" and no test would have noticed.

The conditional is worth keeping: a merge adds up several machines' results in about a second, and printing that second as the sweep's wall clock understates a forty-minute run by two orders of magnitude, on the one line a reader skims. But *"says so instead"* only means anything if the other arm carries a number. A test now reads both back.

```
BASELINE TheGateSaysWhichOfThreeThingsItFound: rc=0 Ran 18 tests  OK
   collapse -> always "merged from its shards"  rc=1 FAILED (failures=1)  PINNED
   collapse -> always the minutes               rc=1 FAILED (errors=2)    PINNED
   retune the minutes format                    rc=1 FAILED (failures=1)  PINNED
```

Hand-checked against a green unmutated baseline with a sha256-verified restore, under `PYTHONDONTWRITEBYTECODE=1`.

That last condition is not decoration, and it is worth writing down: **`retune-string` preserves length by construction**, so a same-second restore of an equal-length mutation is byte-different and `(mtime, size)`-identical — which is all CPython validates a `.pyc` on. My first hand-check reported a mutant as green while the file on disk was provably correct, because the interpreter was running the *previous* mutant's bytecode. `Sandbox.run` already sets `PYTHONDONTWRITEBYTECODE=1` and its comment documents the exact measurement; only my ad-hoc script was missing it. A hand-check of a length-preserving mutation is not evidence without it.

## And one stale finding that needs nothing

The same run's `[57/127] shift-boundary n > ANNOTATION_CAP` is **already pinned** by the boundary case this branch recovers — the sweep started before that test existed. Verified against the current tree: `note condition shifted to >=` → `rc=1 FAILED (failures=1)`.

## Where the sweep stands

`tools/sweep.py --gate --path tools`, **127 mutations**, still running under load average 25-30 from four sibling agents' sweeps on the same box. Seven survivors so far:

| | |
|---|---|
| `shard is not None` (×2) | real — the unsharded path had no test; pinned |
| `per_shard`'s `max(1, …)` | real — divide-by-zero guard; pinned |
| `%0D` / `%0A` / `%25` | real — escaping asserted by absence; pinned |
| `ANNOTATION_CAP` boundary (×2) | real — no case on the edge; pinned |
| `elapsed is not None` | real — this commit |
| `"Pair"` | not pinnable — annotation string under PEP 563, filed as #632 |

**This is still not a complete verdict** and I am not offering it as one. If the run finishes I will report what it says.

## On the merge

No harm done that this branch does not fix, and the check you shipped is working: the dispatch run at `bc8b724` went `Size the sweep` → `Sweep shard 1 of 1` → `Add up what the shards found` → a check named **`no survivors`**, which also exercises the merge-base path since a dispatch has no pull-request base sha.

diazoxide:
Two more from the same run, and a correction to my own method.

## `[59]` / `[60]` / `[61]` — the row that counts the missing shards

```python
if missing:
    w(f"| **did not report** | {f'{missing} of {shards}' if shards >= 1 else '?'} | "
      "a shard that never wrote a result — read nothing below as a count |")
```

Deleting the row outright left all 274 tests green, and so did collapsing the cell in either direction. The test looked for `"did not report"` **anywhere on the page** — and the section heading further down carries the same three words, so the phrase survived with the count gone.

That is the third time on this branch that an assertion about the *absence* of something stood in for one about its *value*: a newline that was gone rather than encoded, a cap with no case on its edge, and now a phrase that appears twice. Which is the harness's own thesis, arriving about the harness, repeatedly.

Fixed in `7699836`, asserted in each of the three shapes the cell takes — `2 of 3`, `1 of 1` for the one-shard plan whose only shard vanished, and `?` when the sweep never sized itself. **The middle one is the case `shards >= 1` actually turns on**, and without it the boundary shifted to `shards > 1` and survived too:

```
BASELINE: rc=0 Ran 19 tests  OK
[59] drop the whole row                rc=1 FAILED (failures=1)  PINNED
[60] collapse the cell -> count        rc=1 FAILED (failures=1)  PINNED
[60] collapse the cell -> '?'          rc=1 FAILED (failures=1)  PINNED
boundary shards >= 1 -> shards > 1     rc=1 FAILED (failures=1)  PINNED
boundary shards >= 1 -> shards >= 0    rc=1 FAILED (failures=1)  PINNED
```

## Correction: a verification run I had reported was void

An earlier full-suite run of mine came back `FAILED (errors=6)` on CPython 3.14 while CI's Linux 3.14 was green at the same sha. The six were all my own new tests, and the traceback was this:

```
took = f"{elapsed / 60:.1f} min" if elapsed is not None else "merged from its shards"
       ~~~~~~~~^~~~
TypeError: unsupported operand type(s) for /: 'NoneType' and 'int'
```

That is one of my **hand-check mutants**, mid-flight. I had run a mutation check in the same working tree the suite was reading, and the suite picked up the mutated file.

So: **a mutation hand-check must not share a working tree with a running verification suite.** It is the same family as the stale-bytecode trap earlier in this branch — the check and the thing being checked have to be isolated, or the check is measuring its own interference. Every run reported from here on is in a dedicated detached worktree that nothing writes to.

Both prior suite results at `bc8b724` are therefore withdrawn as evidence, and re-run clean at `7699836`. The findings above were each re-verified in the isolated tree.

## Where the sweep stands

**62 of 127, twelve survivors, zero unresolved, zero not-applied.** All twelve accounted for:

| survivor | |
|---|---|
| `%0D` `%0A` `%25` | escaping asserted by absence — pinned (this branch) |
| `ANNOTATION_CAP` boundary ×3 | no case on the edge — pinned (this branch) |
| `elapsed is not None`, `.1f` | wall clock never read back — pinned (`964bba7`) |
| `did not report` row ×3 | count asserted by a phrase — pinned (`7699836`) |
| `"Pair"` | annotation string under PEP 563 — not pinnable, #632 |

`shard is not None` and `per_shard`'s `max(1, …)` are absent from that list because this run started *after* they were fixed, and now come back `pinned` at `[1]`, `[2]`, `[8]`, `[9]`.

Still not a complete verdict, and still not offered as one.

diazoxide:
It happened again, on this PR: **#635 merged at `13e8852`, and two commits I pushed to `recover/617-final` afterwards did not go with it.** `964bba7` and `7699836` are still on that branch and are not on `main`.

They are rebased onto current `main` and opened as **#637**. `main` already has the escaping fix and the annotation-cap boundary case from `13e8852`; it does not have the wall-clock or table-row ones.

Not a complaint — the rule in your description is exactly right, and I would rather say plainly that the race caught the recovery too than let two real fixes sit on a merged branch. The branch tip and `main` disagree right now, which is the thing worth knowing.

For what it is worth, the mechanical version of the rule is cheap: `gh pr view <n> --json headRefOid` at merge time, compared against what the merge button is about to take. That is the same check `release.yml` makes about the version it is publishing, for the same reason — an irreversible action should name what it is acting on.

Issue #558, opened 2026-08-26
The dispatch path publishes with strictly weaker checks than the tag path, and nothing gates either


diazoxide:
Found while assessing whether the four dependabot bumps (#478–#481) could be rehearsed before merging, since all four actions sit on the publish path.

## Two facts, measured

**1. The tag/version check does not run on `workflow_dispatch`.**

`release.yml`'s `guard` job compares the tag to `pyproject.toml`, but the step is gated:

```yaml
- name: Compare the git tag to pyproject.toml
  if: startsWith(github.ref, 'refs/tags/v')
```

On a `workflow_dispatch` run `github.ref` is a branch, so the step is **skipped**, not passed. The only surviving check is "can the news entries render", which reads the version from `pyproject.toml` — the same file the skipped step exists to cross-examine.

The comment on the trigger says `workflow_dispatch` "lets a failed release be retried without a new tag", and the news step's own comment acknowledges "a workflow_dispatch retry has no tag to read". Both are true. The consequence is that the retry path has **strictly weaker checks than the path it retries**, and the check it drops is the one guarding the irreversible act.

**2. The `pypi` environment has no protection rules.**

```
$ gh api repos/diazoxide/charter/environments
  pypi: rules=NONE
```

`publish` declares `environment: name: pypi`. That name is load-bearing for Trusted Publishing — PyPI's OIDC claim includes it, so only a job declaring it can mint a valid token. But as a *gate* it is decorative: no required reviewer, no wait timer. Nothing pauses between `build` and an upload that cannot be undone.

## Why the combination matters

Normal release flow is: bump the version, merge, then tag. Between the merge and the tag, `main` carries a version that has never been published. A `workflow_dispatch` in that window skips the tag check (no tag), passes the news check (the entry exists — `news stamp` ran as part of the bump), builds, and publishes — with no tag, no version cross-check, and no human checkpoint.

Today it happens to land softly: `main` is at 0.53.0, already on PyPI, so PyPI would refuse the duplicate. That is PyPI declining, not charter refusing.

This also means the project's stated rule — releases are cut deliberately, together — is enforced by convention alone. The repo does not hold it.

## What this blocks

The four dependabot PRs bump `checkout` v4→v7, `setup-python` v5→v7, `upload-artifact` v4→v7, `download-artifact` v4→v8. Major jumps, all on the publish path, where a break surfaces at tag push and the tag is irreversible.

[...]

PR #637, opened 2026-08-28
Two more lines the sweep found unpinned in the gate it reports (#617)

diazoxide:
Follow-up to #630 and #635. **These two commits were pushed to `recover/617-final` after #635 had already been merged**, so they are on that branch but not on `main` — a second merge race, same shape as the first. Rebased onto current `main` here.

`main` already has the escaping fix and the annotation-cap boundary case. It does not have these.

## `[56]` / `[58]` — the wall clock was never read back

```python
took = f"{elapsed / 60:.1f} min" if elapsed is not None else "merged from its shards"
```

Collapsing that to the second arm left **all 273 tests green**. Nothing asserted the minutes, so every page — measured or merged — could have said "merged from its shards" and no test would have noticed. The `.1f` retune survived for the same reason.

The conditional earns its place: a merge adds up several machines' results in about a second, and printing that second as the sweep's wall clock understates a forty-minute run by two orders of magnitude, on the one line a reader skims. But *"says so instead"* only means something if the other arm carries a number.

## `[59]` / `[60]` / `[61]` — the row that counts the missing shards

```python
if missing:
    w(f"| **did not report** | {f'{missing} of {shards}' if shards >= 1 else '?'} | "
      "a shard that never wrote a result — read nothing below as a count |")
```

Deleting the row outright left the suite green, and so did collapsing the cell in either direction — because the test looked for `"did not report"` **anywhere on the page**, and the section heading below carries the same three words. The phrase survived with the count gone.

Asserted now in each of the three shapes the cell takes: `2 of 3`, `1 of 1` for the one-shard plan whose only shard vanished, and `?` when the sweep never sized itself. The middle one is the case `shards >= 1` actually turns on — without it the boundary shifted to `shards > 1` and survived too.

## The pattern

Three findings in a row, and all one mistake: **an assertion about the absence of something standing in for one about its value.** A newline that was gone rather than encoded; a cap with no case on its edge; a phrase that appears twice. Which is the harness's own thesis, arriving about the harness.

## Verification

Every mutant hand-checked against a **green unmutated baseline** with a **sha256-verified restore**, under `PYTHONDONTWRITEBYTECODE=1`, in a worktree no verification run was reading:

```
BASELINE TheGateSaysWhichOfThreeThingsItFound: rc=0 Ran 18 tests  OK
   collapse -> always "merged from its shards"  rc=1 FAILED (failures=1)  PINNED
   collapse -> always the minutes               rc=1 FAILED (errors=2)    PINNED
   retune the minutes format                    rc=1 FAILED (failures=1)  PINNED

BASELINE (with the row cases): rc=0 Ran 19 tests  OK
   [59] drop the whole row                rc=1 FAILED (failures=1)  PINNED
   [60] collapse the cell -> count        rc=1 FAILED (failures=1)  PINNED
   [60] collapse the cell -> '?'          rc=1 FAILED (failures=1)  PINNED
   boundary shards >= 1 -> shards > 1     rc=1 FAILED (failures=1)  PINNED
   boundary shards >= 1 -> shards >= 0    rc=1 FAILED (failures=1)  PINNED
```

Both conditions in that sentence are load-bearing, and each cost me a wrong answer first:

- **`PYTHONDONTWRITEBYTECODE=1`** — `retune-string` preserves length by construction, so a same-second restore of an equal-length mutation is byte-different and `(mtime, size)`-identical, which is all CPython validates a `.pyc` on. My first hand-check called a mutant green while the file on disk was provably correct, because the interpreter was running the previous mutant's bytecode. `Sandbox.run` already sets this and documents the measurement; only my ad-hoc script was missing it.
- **A worktree no suite is reading** — a later full-suite run came back `FAILED (errors=6)` on CPython 3.14 while CI's Linux 3.14 was green at the same sha. All six were my own tests, and the traceback was one of my hand-check mutants mid-flight. I had mutated the tree the suite was reading.

`tests.test_sweep` + `test_workflows` + `test_news`: **341 tests, OK.**

## Where the sweep stands

`tools/sweep.py --gate --path tools --base bcc1228` is running again on the fixed tree. Twelve survivors found so far across the earlier runs, every one accounted for: three escaping, three annotation-cap boundary, two wall-clock, three table-row — all pinned — and `"Pair"`, which is an annotation string under PEP 563 and can never be pinned (#632).

It has not finished, and I am not offering it as a complete verdict.

### DIFF
--- a/docs/news/unreleased-a-green-sweep-is-not-no-survivors.md
+++ b/docs/news/unreleased-a-green-sweep-is-not-no-survivors.md
@@ -91,7 +91,7 @@ kind of pass.
 
 ## The sweep swept the gate that reports it
 
-127 mutations, against a change whose whole subject is the sweep. It found five lines of
+127 mutations, against a change whose whole subject is the sweep. It found six lines of
 its own that no test went red without.
 
 **The unsharded path had stopped being tested.** Forcing `if shard is not None` to
@@ -124,7 +124,13 @@ instead, because a merge takes a second and printing that second would understat
 forty-minute run by two orders of magnitude. Nothing asserted the minutes, so collapsing
 the whole thing to "merged from its shards" left the suite green and every page saying it.
 
-All five are pinned now. A sixth survivor is not a finding about this code at all: it is a
+**And the row that counts the missing shards was asserted by a phrase that appears
+twice.** The test looked for "did not report" anywhere on the page — and the section
+heading below carries the same words, so deleting the table row outright left it green,
+and so did collapsing the cell that holds "2 of 3". A row is a number; the number is
+asserted now, in each of the three shapes it takes.
+
+All six are pinned now. A seventh survivor is not a finding about this code at all: it is a
 string inside a *type annotation*, which `from __future__ import annotations` means the
 interpreter never evaluates, so no test can ever go red without it. That is a blind spot
 in the string operator's scoping rather than a guard, and it is filed as one (#632)
--- a/docs/news/unreleased-a-green-sweep-is-not-no-survivors.md
+++ b/docs/news/unreleased-a-green-sweep-is-not-no-survivors.md
@@ -91,7 +91,7 @@ kind of pass.
 
 ## The sweep swept the gate that reports it
 
-127 mutations, against a change whose whole subject is the sweep. It found four lines of
+127 mutations, against a change whose whole subject is the sweep. It found five lines of
 its own that no test went red without.
 
 **The unsharded path had stopped being tested.** Forcing `if shard is not None` to
@@ -118,7 +118,13 @@ line is the first thing the cap eats. Every test had either far fewer than ten f
 far more, so "always ten" and "always nine" both passed, and "always ten" is the one that
 loses the note. There is a case at nine, ten and eleven now.
 
-All four are pinned now. A fifth survivor is not a finding about this code at all: it is a
+**And the wall clock was never read back.** The summary prints how long the sweep took,
+or — when the answer was added up from several machines rather than measured — says that
+instead, because a merge takes a second and printing that second would understate a
+forty-minute run by two orders of magnitude. Nothing asserted the minutes, so collapsing
+the whole thing to "merged from its shards" left the suite green and every page saying it.
+
+All five are pinned now. A sixth survivor is not a finding about this code at all: it is a
 string inside a *type annotation*, which `from __future__ import annotations` means the
 interpreter never evaluates, so no test can ever go red without it. That is a blind spot
 in the string operator's scoping rather than a guard, and it is filed as one (#632)
--- a/tests/test_sweep.py
+++ b/tests/test_sweep.py
@@ -2196,6 +2196,22 @@ class TheGateSaysWhichOfThreeThingsItFound(unittest.TestCase):
         text = sweep.gate_summary(gate, "a" * 40, "b" * 40, 60.0, enforce=False)
         self.assertIn("## Deletion sweep — 2 survivors", text)
 
+    def test_the_page_says_how_long_it_took_or_that_it_did_not_measure_that(self):
+        """`elapsed is None`, found unpinned by the sweep on the branch that added it.
+
+        A merge adds up several machines' results in about a second, and printing that
+        second as the sweep's wall clock would understate a forty-minute run by two orders
+        of magnitude, on the one line a reader skims. So the merged page says where its
+        number came from instead — and the *measured* page has to actually carry the
+        minutes, or "says so instead" is the only thing either page ever says.
+        """
+        gate = sweep.classify([_result("pinned")])
+        measured = sweep.gate_summary(gate, "a" * 40, "b" * 40, 1234.0, enforce=False)
+        merged = sweep.gate_summary(gate, "a" * 40, "b" * 40, None, enforce=False)
+        self.assertIn("20.6 min on ", measured)
+        self.assertNotIn("merged from its shards", measured)
+        self.assertIn("merged from its shards on ", merged)
+
     def test_a_page_missing_a_shard_refuses_to_call_the_branch_clean(self):
         gate = sweep.classify([_result("pinned")])
         text = sweep.gate_summary(gate, "a" * 40, "b" * 40, 60.0, enforce=False,
--- a/tests/test_sweep.py
+++ b/tests/test_sweep.py
@@ -2219,6 +2219,27 @@ class TheGateSaysWhichOfThreeThingsItFound(unittest.TestCase):
         self.assertIn("did not report", text)
         self.assertNotIn("Nothing added here is a line the suite would not miss", text)
 
+    def test_the_table_carries_the_missing_shards_as_a_row_with_a_count_in_it(self):
+        """`[59]` and `[60]` from the self-sweep, and both are one mistake.
+
+        The test above looks for "did not report" *anywhere on the page* — and the section
+        further down carries that phrase in its heading, so deleting the table row
+        outright left the suite green, and so did collapsing the cell that holds the
+        count. A row is a number; assert the number, in both of the shapes it has.
+        """
+        gate = sweep.classify([_result("pinned")])
+        counted = sweep.gate_summary(gate, "a" * 40, "b" * 40, None, False,
+                                     missing=2, shards=3)
+        self.assertIn("| **did not report** | 2 of 3 |", counted)
+        # One shard, and it was the one that vanished — the ordinary shape for a small
+        # diff, and the edge `shards >= 1` actually turns on.
+        lone = sweep.gate_summary(gate, "a" * 40, "b" * 40, None, False,
+                                  missing=1, shards=1)
+        self.assertIn("| **did not report** | 1 of 1 |", lone)
+        unsized = sweep.gate_summary(gate, "a" * 40, "b" * 40, None, False,
+                                     missing=1, shards=0)
+        self.assertIn("| **did not report** | ? |", unsized)
+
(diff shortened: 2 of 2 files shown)


diazoxide:
## Verification

**CI green at `7d5f03a`** — nine checks, nothing failed, nothing cancelled:

```
Size the sweep                   :: success
Sweep shard 1 of 1               :: success
Add up what the shards found     :: success
no survivors                     :: success   ← the check's NAME is the answer
test (3.11 / 3.12 / 3.13 / 3.14) :: success
```

**Full suite, two environments**, both on the identical content before it was split between `main` and this PR, and both in a **detached worktree nothing else was writing to**:

| | |
|---|---|
| macOS, CPython 3.12.13 | **Ran 7911 tests — OK** (401 s) |
| macOS, CPython 3.14.4, `CHARTER_*` / `CLAUDE_*` / `ANTHROPIC_*` / `TMUX*` unset | **Ran 7911 tests — OK** (454 s) |

The clean environment is *verified* empty rather than assumed — the run is preceded by a check that no such name survives into `os.environ`, which printed `leaked: []`.

## The sweep, re-run on the fixed tree

`tools/sweep.py --gate --path tools --base bcc1228` — the same 127 mutations, against the tree with all six fixes in. **Zero survivors so far**, and the mutations that survived the earlier runs now come back pinned:

```
[1/127] pinned  tools/sweep.py:1630 [disable-branch] shard is not None
[2/127] pinned  tools/sweep.py:1630 [disable-branch] shard is not None
[8/127] pinned  tools/sweep.py:2094 [unclamp] max(1, (SHARD_BUDGET - SHARD_FIXED) // SECONDS_A_MUTATION)
[9/127] pinned  tools/sweep.py:2094 [unclamp] max(1, (SHARD_BUDGET - SHARD_FIXED) // SECONDS_A_MUTATION)
```

It is at 10 of 127 and still running, on a box carrying four other agents' sweeps. **Not a complete verdict, and not offered as one** — which is the distinction this whole change exists to make sayable.

diazoxide:
Addendum — the two-environment suite has now also been run **at this PR's own head** (`7d5f03a`), not only on the pre-split content, in a detached worktree nothing else writes to:

| | |
|---|---|
| macOS, CPython 3.12.13 | **Ran 7911 tests — OK** (401.5 s) |
| macOS, CPython 3.14.4, `CHARTER_*` / `CLAUDE_*` / `ANTHROPIC_*` / `TMUX*` unset | **Ran 7911 tests — OK** (401.3 s) |

So both environments are green at `7d5f03a` and at `7699836`, four clean runs in total. Nothing outstanding on this PR from my side.

diazoxide:
## A seventh, from the sharded run — and it is why sharding was the right retry

`2998749`. The serial self-sweep was killed three times (harness background limits, then my own over-broad `pkill`, which matched `Sandbox.full()`'s `unittest discover -s tests` argv). Each restart re-measured the front of the file and never reached the back. So I restarted it as `--shard 1/3` — 43 mutations instead of 127, short enough to survive, and **round-robin, so it samples the whole file immediately** rather than crawling the first third again.

It reached line 2481 within seventeen mutations, and found this:

```python
def outcome(o: dict | None) -> Outcome | None:
    return None if not o else Outcome(o["green"], o["ran"], o["detail"])
```

Collapse it to `None` and **every page still matched**. The round-trip test compares a merged summary against the one a single machine would have written — and a subset's `detail` is printed in exactly one place, the *not-applied* section, which the fixture had no result for. So `--verdict` could drop every outcome its shards reported and nothing would have gone red.

That is the case that matters most for it. A shard saying "the edit never landed" is the sweep reporting a defect **in itself**, and the merge has to be able to say why. `as_json` wrote the detail; nothing read it back.

```
BASELINE TheAnswerSurvivesTheTripThroughAFile: rc=0 Ran 11 tests  OK
   collapse -> always None           rc=1 FAILED (failures=1)            PINNED
   collapse -> always Outcome(...)   rc=1 FAILED (failures=2, errors=3)  PINNED
   retune the 'detail' key           rc=1 FAILED (failures=2, errors=4)  PINNED
```

Same conditions as the rest: green unmutated baseline, sha256-verified restore, `PYTHONDONTWRITEBYTECODE=1`, in a worktree no suite was reading.

**Running total: seven unpinned lines in the gate's own new code**, every one of them an assertion about the *absence* of something standing in for one about its *value*. That is the harness's own thesis, and it keeps being about the harness.

`tests.test_sweep` + `test_workflows` + `test_news`: **342 tests, OK.** The shard is still running; shards 2 and 3 have not been run at all, so this is still not a complete verdict.

diazoxide:
## Verification at `7d7d165`

**CI green**, both workflows:

```
test (3.11 / 3.12 / 3.13 / 3.14)  :: success
Size the sweep                    :: success   (4m01s — 240s of it tracing the map)
Sweep shard 1 of 1                :: success
Add up what the shards found      :: success
no survivors                      :: success
```

The sweep chain needed a `workflow_dispatch` to run at all — see the note at the end.

**Full suite, two environments**, in a detached worktree nothing else writes to:

| | |
|---|---|
| macOS, CPython 3.12.13 | **Ran 7914 tests — OK** (396 s) |
| macOS, CPython 3.14.4, `CHARTER_*` / `CLAUDE_*` / `ANTHROPIC_*` / `TMUX*` unset | **Ran 7914 tests — OK** (376 s) |

## Shards 2 and 3 refused to answer, and that is the harness working

Both hit this:

```
baseline: full suite, unmutated…
Ran 0 tests — no tests ran in 419s
! the tree is RED before any mutation. Every mutation below will look
! pinned for a reason that has nothing to do with the guard. Fix first.
```

The tree is not red — the same commit passed 7914 tests twice, minutes earlier. The baseline subprocess produced no test output at all, on a box carrying **five other agents' sweeps** at load average 37. Starved, or killed by somebody else's `pkill`; `Sandbox.full()` runs `unittest discover -s tests`, which is a pattern more than one agent here matches on.

**The point is what the harness did with that.** It had a baseline of `ran=0`, and instead of treating "no failures observed" as green — which would have marked all 42 of its mutations `pinned` on the strength of a stopwatch — it refused to sweep. That is §4's second way a sweep lies, and the third of this branch's own guards to be demonstrated against the branch itself:

- the stale `.pyc` (length-preserving mutations, `PYTHONDONTWRITEBYTECODE`),
- the tree mutated under a running suite,
- and now the red baseline.

Each one cost me a wrong answer first, and each was caught by a guard already in the tool rather than by me noticing.

The over-subscription was my fault — I ran both shards at once alongside shard 1 and a sibling's sweep. Shard 2 is re-running alone. **The 84 mutations in shards 2 and 3 are still unmeasured, and this is still not a complete verdict.**

## A gap in the workflow I shipped

`sweep.yml` triggers on `pull_request` and `workflow_dispatch` only — no `push:`. Two pushes to this branch produced no `pull_request.synchronize`, so **the gate did not run at all**, and nothing said so: the checks list simply had no sweep row. `test.yml` has `push:` and ran both times, which is why the absence was easy to miss.

That is the same silence #617 is about, one level up — a branch with no gate run looks exactly like a branch whose gate found nothing. Adding `push:` closes it. I have not changed it here, because it is a separate decision about when the gate should run and this PR is already carrying eight fixes; say the word and I will open it as its own change.

diazoxide:
## A sixth way the machinery lies, and it is different in kind

The spec names five, and all five produce a wrong **answer**: exit-code scoring, a broken
baseline, stale bytecode, a hang, a mutation that never applied. This one destroys the
**work** while making the answers meaningless, which is why it does not fit the list:

> **A hand-check harness owns its worktree exclusively, start to finish.** Capturing the
> file at start and restoring it in `finally` is correct in isolation and wrong the moment
> anything else touches that tree.

It cost me both halves in one afternoon. The restore reverted an edit I had made while it
was running — silently, because a `finally` that writes back its own snapshot cannot know
somebody else's change is in it. And its verdicts had been measured against a file I was
editing underneath them, so they were meaningless anyway. I only noticed because a test
failure quoted `invalid literal for int() with base 10: ''`, which I recognised as my own
mutant rather than a regression.

The four earlier self-inflicted ones were each caught by a guard already in the tool — the
`PYTHONDONTWRITEBYTECODE` the sandbox sets, the red-baseline refusal, the green-baseline
assertion. This one nothing catches, because nothing is watching for it. Hence writing it
down.

## And the sharpest single illustration this project has

`parse_shard` refuses an argument that is not `N/M`. **Delete the refusal outright and
every test still passes**, because `int("")` and `int("two")` raise `ValueError` too:

```
with the refusal   --shard wants N/M, not '2'
without it         invalid literal for int() with base 10: ''
```

`assertRaises(SomeError)` is an assertion about **type**, and type is not the reason. That
is #558 — deleting `release.yml`'s `-z "$claimed"` left the run still exiting 1 for a
different reason — reproduced inside the harness built around #558, by the harness. It is
in the news entry alongside the fix.

## Where this stands

`f439cad` — merged with current `main` (which now carries #643, so #632's unpinnable
annotation survivor is gone), **383 tests green** on the merged tree.

**Fourteen findings.** Ten were tests asserting too little; three were lines answering
nobody and were deleted; one looked exactly like those three and was the opposite — the
`str()` in `expected_shards` is the only thing stopping `3.9` from being read as three
shards. Each of those four needed a measurement, not a reading.

The worst of them, `[42/42]`: `if not shards: missing = max(missing, 1)` deleted cleanly,
because the test covering it used an *empty* directory where `merge` already returns
`missing=1`. One stray artifact and that guard is the only thing between a failed plan job
and a page reading **no survivors** — #617 exactly, inside the change written to prevent it.

**Shard 3 is running now**, on a tree whose `tools/sweep.py` I verified byte-identical to
this head. Its 42 mutations have never been asked about. Not a complete verdict until it
reports, and I am not claiming one.

PR #639, opened 2026-08-28
A cross-repo landing is on the release floor, a revert is a new change, and divergence is named (Phase 4, Tasks 9-14)

diazoxide:
Phase 4, Tasks 9–14 of `docs/superpowers/specs/2026-08-28-phase4-cross-repo-change.md` — the last tranche. Builds on #629 (Tasks 1–4). **Tasks 5–8 are not in the tree**; where I needed something from them I built against the spec's stated shape and say so below.

## Task 9 — the floors

`("charter", "change", "land")` joins `hooks._PUBLISH_FORGE`, **and the reader widened with it**. The lookup lived only under `elif base in ("gh", "glab")`, so the tuple alone would have been a set member nothing could reach — a floor that never runs, in a phase whose thesis is that a guard nobody pins is a comment with a runtime cost. `_is_charter` rather than a bare name check, so `edm change land` and `python3 -m charter change land` are the same command to the guard. **Deny, never ask**: an unattended `_ask` falls back to `allow`.

Two mutations prove the pair, and neither is redundant: delete only the tuple (RED), and revert the base check while keeping the tuple (RED). Verified through the real `charter hook pretooluse` **and** across the process boundary in `tests/test_hook_process_boundary.py`, with a negative control (`charter change show` unattended is untouched).

`change` joins `toolgate._DANGEROUS["charter"]` and `["edm"]`. The grain is coarse on purpose — it declines `charter change show` too — and a test says so rather than leaving it implied, because the asymmetry is the argument: over-refusing costs one prompt, under-refusing auto-approves a merge, and the gate never denies.

`doctor` grows a `changes` row that names `charter guard ask --local 'charter change land *'` and **writes nothing** — asserted over the whole plane tree on all four of the check's branches, because the first version of that test covered one branch and a mutation writing from another survived it.

## Task 10 — `charter change revert`

A new change, `revert-<slug>`, whose members are the landed members of the original, each seeded off its clone's **own** default branch with a revert of the sha the landing log recorded.

- **`-m 1` only when git says that sha has more than one parent**, asked of git and never remembered.
- **The ordering is the original's, reversed.** If `charter-metrics` needed `charter` to land first, undoing it goes the other way — reverting the base while its dependent still needs the API it removes leaves the dependent broken. A revert that copied `needs` across would get this exactly backwards and look right in every single-member example.
- **A member with no landing record is named, not guessed** — no merge sha, so no revert, and charter will not pick the commit that looks about right.
- **It refuses over a dirty working tree**, by name: `git revert` commits into the checkout.
- **No force-push, no branch deletion, no default-branch reset, no request touched.** Not flags that default off — argv that is never built. Pinned two ways: a test that records every git invocation a real revert makes, and a source-level scan that also covers the paths that test does not walk.

## Task 11 — the surface

`gather.scan` carries the change records and the landing declarations — both file reads — in the one snapshot under its one timestamp. `component.NEEDS` and `ctx.SERVES` gained `changes` in the same commit, asserted against each other in one assertion.

The `changes` component is a **section of the persona sidebar**, not a pane, and both reasons were measured:

1. The frame supports **exactly one variable-height pane** by construction. `layout.slot_sizes` answers every member of `VARIABLE_ROW_SLOTS` with `layout.repos_rows`, and `_reassert_sizes` leaves that set unasserted so tmux's `resize-pane -y` — one boundary — has one remainder. Registered as `Content()` it was handed the **repo table's** height: 6 rows, for 6 repos, on a plane with one change.
2. A **placed** component has to be in `FRAME_SLOTS`, which is pinned to agree with the shipped `slots` default and with `density = full`. Placing it puts a pane saying "no changes in `<ws>`" on **every** operator's frame, for a feature most planes never use. `repos` saying "no clones" is a plane that is broken or new; a plane with no cross-repo change is the ordinary, permanent state.

As a section it costs those planes nothing: it returns no rows when there are none, exactly as the todos do. `F2` → the `change` row picks one; the picked one carries the picker's own mark and does **not** move to the top.

**Nothing on the repaint path starts a process**, measured by replacing every `subprocess` entry point with one that fails the test.

## Task 12 — divergence, named

Five checks, each with its own sentence, all at **FAIL**: a member landed outside charter; a declared landing the default branch no longer contains; a declared landing on a commit carrying no `Charter-Change` trailer; a member that landed while a blocker had not; and a member's branch name in a clone that is a member of no change. It reads what is already on this disk and never fetches — so it can under-report and can never invent.

## Task 13 — ADR 0020, docs, news

`docs/adr/0020-there-is-no-cross-repo-merge-loop.md`, `docs/changes.md` (force-included in the wheel, linked from the README), the floor entry in `docs/hooks.md`, the fourth store in `docs/workspaces.md`, and `docs/news/unreleased-cross-repo-change-surface.md` (`version: unreleased`, `adopt: workspace reinit --all`, no `check:`). No bump, no stamp, no tag.

## Findings — three places the spec or the tree turned out to be wrong

**1. `git revert -m 1` does NOT fail on a one-parent commit** (git 2.50.1, Apple Git-155). The spec's reason for the conditional is *"a squash landing is an ordinary commit and `-m` on one fails"*; measured, `-m 1` returns 0 on a one-parent commit, and so does `-m 2`. An outcome assertion therefore cannot see this at all, and hard-coding `-m 1` **survived** my first mutation round. The conditional still earns its place — older gits do fail with *"mainline was specified but commit is not a merge"*, and passing `-m` on a non-merge is a claim about the commit that is not true — so the test now pins the **argv**, which is a claim about charter and holds on every git.

**2. Task 11's `changes` component cannot be what the spec describes.** §3.6's example is a change with its member rows in a pane of its own; the two measurements above say that pane cannot exist without either mis-sizing itself or landing on every operator's frame. The section is what the frame can honestly carry, and `charter change show <slug>` is the full view §3.4 already names.

**3. `charter change revert` inherits the operator's commit signing.** On a machine using 1Password's `op-ssh-sign` the suite **hung** with no output, waiting for an approval nobody could give. Production deliberately does **not** pass `-c commit.gpgsign=false` — a revert is the operator's commit in their repository, ADR 0014 puts signing policy with the host, and `charter git-policy --apply` already writes `commit.gpgsign = false` into each managed clone's *local* config for exactly this reason. The fix is in the fixture.

## Not in this PR, and why

- **ADR 0002's amendment is deferred to whichever PR lands Task 6.** Its consequence — *"the reporting module is the only place in charter that writes to a forge"* — is still **true** in this tree, because there is no second write seam yet. Amending an ADR to describe code that is not here would be worse than leaving it.
- **`docs/forges.md` is untouched** for the same reason: `checks_at` and forge writes are Tasks 5/6.
- **`change.record_landing` / `change.landings`** are built here because `revert` and the divergence report read them. Task 8 is the writer's caller. If Tasks 5–8 land separately this is the one function likely to conflict.
- **The floor guards a command that does not exist yet.** That is the right order: a floor that arrives after the verb is a window.

## The deletion sweep, run by the repository

The gate ran twice. Both verdicts, because the delta is the point:

| | unpinned | masked cluster | platform-deferred | unresolved | pinned | headline |
|---|---|---|---|---|---|---|
| first push | 6 | 108 | 5 | 1 | 140 | `114 survivors, 1 not measured` |
| after working the list | 2 | 85 | 3 | 1 | 168 | `87 survivors, 1 not measured` |
| **final head** | **1** | **84** | **3** | **1** | **170** | **`85 survivors, 1 not measured`** |

`114 = 6 + 108`; platform-deferred and unresolved are counted separately. The gate is reporting-only.

**Five of the six unpinned were genuine gaps** and are pinned now. **One remains, and it is `GIT_TIMEOUT = 20 → 21`** — no observable behaviour changes between 20 seconds and 21, and a test asserting the constant against itself is a tautology, so what is pinned is that `_git` — the one funnel — always passes it. The second was pinnable and is fixed: `changes_section` checks its budget *before* `gather.read`, and the test asserted `== []` alone, which is satisfied either way — so `budget <= 0` could become `budget < 0` with the cache opened on every repaint of a pane with no room.

**Two findings that bear on `--enforce`:**

- **It found a dead line, not just a missing test.** `cmd_change_revert` checked that `revert-<slug>` is a change name; `_load` has already put the slug through `change_name_ok`, and `CHANGE_NAME_RE` has no length bound and no leading-character rule a `revert-` prefix could break. Deleted, property pinned on the derivation. Nothing but a sweep finds those.
- **It reports at least one mutation nothing can ever distinguish.** `split(".")[0]` and `rsplit(".")[0]` are the same expression without a maxsplit, and `pieces._host` on `main` carries the identical line.

**The 85 masked-cluster survivors need `--second-order`, which the gate does not run.** I have not resolved them and am not claiming they are equivalent. Working the bucket by feeding malformed input — rather than reasoning about it — is what moved `pinned` from 140 to 168. Full analysis in a comment below.

## Verification

- **8298 tests, OK, in two environments** at the head under review: CPython 3.14.4, and CPython 3.12.13 with `CHARTER_*` / `CLAUDE_*` / `ANTHROPIC_*` / `TMUX*` unset.
- **CI green at the head sha** — read through `gh api repos/diazoxide/charter/commits/<sha>/check-runs`, which cannot confuse "passed" with "did not run", rather than a rollup that can.
- **All 8 shards reported through the new `upload-artifact` v7 / `download-artifact` v8 pair** (#649): `merged 259 result(s) from 8 of 8 shard(s)` on the previous head. Nothing lost in the round-trip; that was the first 8-shard run on the new pair.
- **No test makes a network call.** Every git invocation the change surface can make is listed in full by a static test, and none of them touches a remote.
- Rebased onto `main` four times as it moved; the last carries #645, #647, #649 and #650. #645 touches `frame/gather.py` and `frame/state.py`, which this branch modifies heavily — the rebase is clean and both changes coexist, verified rather than assumed.

### One flake, chased down — and it was somebody else's, now fixed

Local full-suite runs failed intermittently in `tests/test_frame_input_reaches_a_component.py`, a real-tmux module this branch does not touch. Six different test methods failed across the runs, which is the signature of timing rather than logic. It is **pre-existing and not this branch's**:

- **It reproduced on plain `main` with this branch absent** — 1 failure in 4 full-suite runs at `a23cd61`, on the same test method that failed one of the runs here.
- **Pane geometry is byte-identical** between this branch and `main`: `FRAME_SLOTS`, `FRAME_DENSITY["full"]`, `SLOT_SIZE`, `SLOT_EDGE`, `VARIABLE_ROW_SLOTS` and `SLOT_OF` all compare equal, because `changes` is a *child* of the sidebar and absent from every placement table. The failing cases split real panes and wait on a five-second timeout; nothing here changes how much room they have.
- **This branch's own modules run immediately before it 4/4 clean**, so it was not interference from the new tests.

**#650 has since fixed it** — *"the input test stops reading its own client's arrival as a resize"* — and it is gone: both full-suite environments at this head are clean. The higher local rate on this branch before the fix was duration, not correctness; this branch adds ~155 tests and the machine was at load averages near 70.

One earlier 3.12 failure could not be named at all, because the run's output went through my own `tail -4` and the test name went with it. Recorded rather than quietly dropped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,6 +379,9 @@ copy wins, is compared to nothing, and drifts unwatched in both directions.
   constrain.
 - [docs/forges.md](docs/forges.md) — what GitLab and GitHub each need, self-hosted hosts,
   and the rule for a repo name that collides across forges.
+- [docs/changes.md](docs/changes.md) — one change spanning several repos: what the record
+  holds, why it holds no state, what charter refuses to do with it, and how a revert is a
+  new change rather than an undo button.
 
 ## Contributing
 
--- a/charter/change.py
+++ b/charter/change.py
@@ -50,6 +50,8 @@ merge request is a **request**. Neither name is renamed; both are shipped.
 from __future__ import annotations
 
 import json
+import re
+import socket
 from pathlib import Path
 
 from . import config, contain, instance, workspace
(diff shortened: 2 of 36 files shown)


diazoxide:
## Mutation testing — the mutation actually run, and its actual result

Every one below was applied to the real tree by a harness that asserts three things, because a hand-check without an unmutated baseline is not a verdict: the edit **changed the file** (sha256 before ≠ after), the named tests went **RED** with it in place, and the restore put the **byte-identical** file back (sha256 == the original). `__pycache__` cleared between runs. A mutation that silently did not apply looks exactly like a mutation nothing caught.

### Task 9 — the floors (5/5 pinned after a fix)

| mutation | result |
|---|---|
| remove `("charter","change","land")` from `_PUBLISH_FORGE`, keeping the widened base check | RED |
| revert the base check to `("gh","glab")`, **keeping** the entry | RED |
| make the release floor `_ask` instead of `_deny` | RED |
| remove `change` from `toolgate._DANGEROUS["charter"]` | RED |
| `doctor` writes the guard rule on the *no changes* branch | RED |
| `doctor` writes the guard rule on the *OK with changes* branch | RED |
| `doctor` stops naming `--local` in the prompt it recommends | RED |

The first two are the pair Step 3b asks for and neither is redundant: delete only the tuple and the base check stays green; revert only the base check and the tuple lookup stays green. **The doctor row cost a test fix.** The first version asserted "wrote no file" over a plane with one un-landed change, and a mutation writing the rule from the *no changes* branch **survived** it. It now snapshots the whole plane tree across all four of the check's branches.

### Task 10 — `charter change revert` (7/7 pinned after a fix)

| mutation | result |
|---|---|
| force-push during a revert | RED |
| delete a branch during a revert | RED |
| guess the merge sha from the branch name | RED |
| revert over a dirty working tree instead of refusing | RED |
| copy the original's ordering across instead of reversing it | RED |
| drop a member with no landing record silently instead of naming it | RED |
| **use `-m 1` unconditionally instead of asking git** | **SURVIVED → fixed** |

That last one is the finding in the PR body. `git revert --no-edit -m 1 <one-parent-sha>` returns **0** on git 2.50.1 (Apple Git-155) — and so does `-m 2` — so the spec's stated reason (*"`-m` on one fails"*) does not hold here and no outcome assertion can see the mutation at all. The test now pins the **argv**: no `-m` when git says one parent, `-m 1` when it says two. That is a claim about charter and holds on every git, and the conditional still earns its place because older gits do fail with *"mainline was specified but commit is not a merge"*.

### Task 11 — the surface (8/8 pinned after a fix)

| mutation | result |
|---|---|
| declare `changes` in `NEEDS` without serving it in `SERVES` | RED |
| `gather.scan` stops carrying the slice | RED |
| render the aggregate as the **best** member | RED |
| an empty member set reads as `landed` | RED |
| an unrecognised state ranks as `landed` rather than `unknown` | RED |
| the repaint calls the forge (a subprocess on the tick) | RED |
| a blocked member reads as `unknown` (the ordering derivation dropped) | RED |
| **skip `contain.one_line` on the change slug at the drawing site** | **SURVIVED → fixed** |

The second finding, and it is the sharper one. `tui.truncate` sanitises on the way out — it removes `\n`, `\r`, the escape and the NUL — **but it leaves U+2028 exactly as it found it**. My containment test split on `"\n"`, so it could not tell `contain.one_line` from its absence. It now uses `str.splitlines()` and asserts the absence of every codepoint a terminal treats as a line break, with a control that shows `tui.truncate` alone really does let U+2028 through. The guard is load-bearing; the test was not.

A deliberate no-op control (appending `+ ""` to the state lookup) **survived**, which is what says the harness can tell the two apart.

### Task 12 — divergence (7/7 pinned)

| mutation | result |
|---|---|
| downgrade the divergence row to WARN | RED |
| downgrade the unreadable-record row to WARN | RED |
| drop the out-of-order report | RED |
| drop the landed-outside-charter check | RED |
| drop the trailer check | RED |
| drop the git-no-longer-contains check | RED |
| drop the stray-branch check | RED |

diazoxide:
## The deletion sweep, and what its numbers say about `--enforce`

The gate ran twice on this branch. Both verdicts, because the delta is the interesting part:

| | unpinned | masked cluster | platform-deferred | unresolved | pinned | **headline** |
|---|---|---|---|---|---|---|
| first push | 6 | 108 | 5 | 1 | 140 | `114 survivors, 1 not measured` |
| after working the list | **2** | 85 | 3 | 1 | **168** | `87 survivors, 1 not measured` |

`114 = 6 + 108`; the platform-deferred and the unresolved are counted separately. **The gate is reporting-only — nothing here blocks.**

### The unpinned bucket is small, actionable, and it found real defects

Six, of which **five were genuine gaps in my tests** and are now pinned: `gather._empty` must carry `changes` as a list (that structure is what a renderer indexes into when everything else failed); `gather.scan` must survive `_change_rows` raising; `change.member_states(rec, None)` — `set(None)` raises, and `None` is what a caller with nothing landed passes, which is every change on the day it is created; `change._host` must take the **first** label, because a log named for a fully-qualified domain splits in two the first time somebody joins a VPN; `choose.current` must answer `""` and not `None`, which is its own annotation's claim.

The second run left two, and both are worth reporting rather than fixing:

**One was pinnable and the test name had already claimed it.** `changes_section` checks its budget before it reaches `gather.read`, so a pane with no room does not open a file it is about to discard. The test asserted `== []` alone — satisfied either way, because `_change_rows` refuses a zero budget too — so `budget <= 0` could become `budget < 0` with every assertion green and the cache opened on every repaint. Now fixed, both sides of the boundary. **A test whose name claims more than its body checks is the shape this suite keeps finding**, and the sweep is what found it.

**One cannot honestly be pinned at all.** `GIT_TIMEOUT = 20 → 21` changes no observable behaviour of any test, and a test asserting the constant against itself is a tautology. What is pinned instead is that `_git` — the one funnel — always passes it. If `--enforce` lands, this class needs somewhere to go that is not a suppression list; it is a real cost and it is small.

### Two findings that bear directly on whether the gate is worth enforcing

**It found a dead line, not just a missing test.** `cmd_change_revert` checked that `revert-<slug>` is a change name. It cannot fail: `_load` has already put the slug through `change_name_ok`, and `CHANGE_NAME_RE` has no length bound and no leading-character rule a `revert-` prefix could break — measured across the alphabet's edges. Deleted, with the property pinned on the derivation instead, where it goes red if the rule ever grows a length cap. **A guard nothing can reach is a comment with a runtime cost**, and nothing but a sweep finds those.

**And it reports at least one mutation nothing can ever distinguish.** `split(".")[0]` versus `rsplit(".")[0]` is the same expression without a maxsplit, and `pieces._host` on `main` carries the identical line. Not a gap, not fixable, and it will be reported on every branch that touches that shape.

### The masked-cluster bucket is 85 of the 87, and that is where the decision lives

The gate's own words: *"Another survivor shares this function. Two guards in sequence each look equivalent alone, so neither is pinned and they are read together."* They concentrate in the functions with several sequential refusals — `divergences` (13), `slots._change_rows` (11), `_seed_revert` (9), `cmd_change_revert` (7), `stray_branches` (6).

**Working them is what moved `pinned` 140 → 168, and it was not a suppression exercise.** The method that paid was feeding malformed input rather than reasoning about it: the drawing path is now driven with rows missing every field, carrying `None` where numbers belong, and holding a state charter has never heard of — and none of them may read as `landed`. The landing log is fed nine kinds of broken line, each skipped without taking the good lines around it down. Every way `_seed_revert` can refuse has its own case and asserts the absence of its neighbours' words.

What remains needs `--second-order`, which the gate does not run — pairs within one function, measured together. **I have not resolved those 85 and am not claiming they are equivalent.** On this branch the honest summary is: the unpinned bucket is worth enforcing today, and the masked bucket needs the second-order pass before a number from it means anything.

Issue #546, opened 2026-08-25
Two tests in test_vault_registration.py run the operator's real op binary against their real 1Password vault (and hang when it prompts)


diazoxide:
Two tests in `test_vault_registration.py` run the operator's REAL `op` binary against their REAL 1Password vault

`tests/test_vault_registration.py` reaches the 1Password CLI on `$PATH` — not a fake, not a stub — with the vault names the plane's own config carries:

```
--- op item get  charter-devops --vault Eng         --format json
    test_force_does_not_migrate_and_says_so
      (test_vault_registration.RegisteringOverAnExistingName)
--- op item list               --vault Eng         --format json
    test_force_does_not_migrate_and_says_so
--- op item get  charter-ops    --vault Engineering --format json
    test_the_account_pin_never_travels
      (test_vault_registration.SharedAndLocalHalves)
--- op item list               --vault Engineering --format json
    test_the_account_pin_never_travels
```

Measured by putting a logging stand-in for `op` first on `$PATH` and running the whole suite: those four invocations, from those two tests, and no others anywhere in 5652 tests.

## Why it has not been noticed

Both tests pass when `op` **fails fast** — an unauthenticated `op` exits non-zero in well under a second, the provider reports a vault it cannot read, and the assertions (which are about charter's refusal to migrate, and about an account pin not travelling) hold for the right reason by luck.

They do not pass when `op` **blocks**. `op` prompts for authorization when it has no usable session, and a shell that is not the one the desktop app integration was authorized for is exactly that: measured inside a fresh tmux pane, `op item get charter-devops --vault Eng --format json` sat waiting for 15 minutes with the suite parked in `subprocess.communicate` behind it — `python3 -m unittest discover -s tests` never finishes.

So the suite's answer here depends on the operator's 1Password auth state, which is neither charter's state nor the shell's. That is the same class #540 closed for `$EDM_*`/`$CHARTER_*` and #544 tracks for `$COLUMNS`, one layer out: the environment the suite inherits decides the result.

## The second problem, which is the worse one

Independently of the hang: a unit test should not be reaching a real credential store. Today it is a read of two named items in two named vaults on whoever's machine runs the suite. Nothing in the assertions wants a real `op` — every other `op` test in the tree (`test_op_reads_the_item_once.py`, `test_op_unreadable_*.py`, `test_onepassword_single_item.py`) drives a fake, and this module has the same fakes available to it.

## Suggested fix

[...]

Issue #323, opened 2026-08-21
A branch name reaches `gh api -F` unencoded, where a leading `@` means "read this file"


diazoxide:
`charter/forge/github.py`'s `ci_status` hands three values straight to `gh` as `-F` field
arguments, none of them encoded:

```python
# charter/forge/github.py:185-188
p = util.run([self.cli, "api", "graphql", "--hostname", self.host,
              "-f", f"query={_ROLLUP_QUERY}",
              "-F", f"owner={owner}", "-F", f"name={name}",
              "-F", f"ref={branch}"], check=False)
```

`open_change`, eight lines above in the same class, encodes the identical values with
`urllib.parse.quote(..., safe="")` — and so do both GitLab calls. `ci_status` is the only
one of charter's four forge status calls that does not. The asymmetry looks like an
oversight rather than a policy.

It matters because `-F` is not a string pass-through. From `gh api --help`:

> The `-F/--field` flag has magic type conversion based on the format of the value:
> […] if the value starts with `@`, the rest of the value is interpreted as a
> filename to read the value from. Pass `-` to read from standard input.

So a `-F` value is a **filename `gh` dereferences**, and charter is building those values
out of data it did not write:

* `branch` is `util.branch_of()`, read straight out of the tree's `.git/HEAD`.
* `owner`/`name` are `glstate._remote_path()`, parsed out of `git remote get-url origin`
  and split on the first `/` — so `name` carries everything after it, slashes included.

**Authority reached.** A value shaped as a path makes `gh` — a process holding the
operator's forge token — open an arbitrary local file the operator can read and place its
contents in the body of an authenticated request to the configured host. A value naming
stdin makes it block on the inherited descriptor. A value naming a file that never reaches
EOF makes it allocate without bound.

The request destination is not attacker-chosen: `--hostname` is always `self.host`, and
`registry.resolve_host` only ever returns a declared or default host (I tried nine
host-confusion URLs against `_host_of`; every one returned `None`, i.e. unmanaged). So this
is not an attacker-readable exfiltration channel. What it is: an unbounded local file read
driven by external data, and those file contents leaving the machine to a third party.

**Verified against 0.47.2**, with a shim on `PATH` recording argv. A real
`charter gl-refresh --workspace <ws>`, run against a throwaway clone, built:

```
['api', 'graphql', '--hostname', 'github.com', '-f', 'query=…',
 '-F', 'owner=acme', '-F', 'name=@<path>', '-F', 'ref=@<path>']
```

[...]

Issue #641, opened 2026-08-28
The suite hangs on any machine with signed commits: 30 of 41 modules run git commit with no shared neutralizer


diazoxide:
Found by the Phase 4 Tasks 5–8 agent, whose fixture repos inherited the operator's own git config.

## What happens

A fixture repo created by a test inherits `commit.gpgsign=true` from the machine's git config. With 1Password's `op-ssh-sign` as the signing program, `git commit` **parks on a biometric prompt** and the suite never returns.

Not a failure — a **hang**. The worst outcome available: not a pass, not a fail, no verdict, and no output saying why.

## Scope

**30 of 41 modules that run `git commit` have the same exposure**, and there is **no shared neutralizer**. The finder fixed it in their own helper; the other 29 remain.

## This is #546 one tool over

`util.run`'s docstring already says, in as many words, that it does **not** cover this. So it is a known gap that nothing has closed.

The family, all closed this week:

- **#546/#587** — two tests ran the operator's real `op` binary against their real 1Password vault, and hung when it prompted. Closed by `tests/_planeguard.RealVaultReach` refusing a `Popen` of `op`/`vault`/`npx`.
- **#545** — `test_frame_launcher` hung forever from a terminal: **126 blocked `input()` calls across 122 tests**, from 7 declaration sites. Closed by `tests/_ttyguard.py` installed **first** in `tests/__init__.py`, above the import that pulls charter in.
- **#638** — the suite validates the operator's real forge token against github.com 23 times per run.
- **#542, #564, #544, #529, #494, #537, #519/#521/#528** — PyPI, tmux servers, `$COLUMNS`, the filesystem, the clock, the cwd, the shell environment.

Every one is *the suite reads or spends the machine instead of the repo*. This is the same defect against **the operator's signing key and their thumb**.

## The prior art says what the fix looks like

Not 30 hand-patches. `#587` and `#590` both closed their family with **one guard installed before anything is collected** — and #590's measurement is the relevant one: *no exit-time cleanup can reach an interrupted run*, so the reaper runs at **start**. A fixture-repo helper that sets `commit.gpgsign=false`, `GIT_CONFIG_GLOBAL=/dev/null` or the `GIT_CONFIG_*` pair (which a dozen modules already carry for other reasons) should serve all 30.

Then **tighten a guard to refuse a signing invocation**, the way `RealVaultReach` refuses `op` — otherwise the exposure returns the first time someone writes a 42nd module.

## Why it has not bitten CI

[...]

PR #644, opened 2026-08-29
A head sha with no check run says NOT RUN rather than passing, and a change lands one member at a time (Phase 4, Tasks 5-8)

diazoxide:
Phase 4, Tasks 5–8 of `docs/superpowers/specs/2026-08-28-phase4-cross-repo-change.md`,
plus the floor from Task 9 that Task 8 cannot ship without. Builds on #629, which landed
while this was in flight; rebased onto `main`, so the diff is this half only.

## The failure it prevents

**A check that never ran reports CLEAN, and the merge button is offered anyway.** `gh pr
checks` says *"no checks reported"* and `mergeStateStatus` says `CLEAN` **identically when no
run was ever created**. That is #561, and it nearly merged an unverified branch on this
repository. Any script that lands a cross-repo change reads one of those two, because they
are what the CLI puts in front of you.

So `charter change land` reads neither. It reads the checks **at the head sha**, through a
protocol method that answers **two fields rather than one string** — a single string is what
made `ci_status`'s `None` mean six different things — and refuses on anything but `passed`:

| | `total` | `state` |
|---|---|---|
| charter could not ask, or could not ask *completely* | `None` | `unknown` |
| charter asked everywhere it knows to look; nothing there | `0` | `not_run` |
| ≥1 check, everything that concluded concluded well | `n` | `passed` |

`unknown` > `failed` > `running` > `not_run` > `passed`, because the only value meaning
"charter did not look" must never be outranked by one meaning "I looked and it was fine".

**The correction a reviewer already forced into the spec is implemented, not quoted.**
`commits/<sha>/check-runs` returns Check Runs *only*, so a repo reporting through Commit
Statuses — Jenkins, Buildkite, CircleCI — is `total_count: 0` at a fully green head: the same
bug from the other side, a permanent `NOT RUN` against a gate that offers no `--force`.
GitLab's mirror image is the merged-results pipeline, which runs against
`refs/merge-requests/:iid/merge` and whose sha is not the branch head. So GitHub reads check
runs **and** the combined commit status; GitLab reads the merge request's own head pipeline;
and **where either cannot enumerate completely the answer is `unknown`, never `not_run`.** A
false `not_run` costs a re-run. A false `passed` merges untested code.

## What this changes

- **`Forge` gains two reads** (`checks_at`, `request_for`) and **three writes**
  (`create_change`, `update_change_body`, `merge_change`) plus `change_body`. Writes raise
  `ForgeWriteError` and never route through `_api` — asserted against the syntax tree,
  because a swallowed write failure passes every happy-path test. `-f` never `-F` (#323),
  pinned on the argv of all six. `open_change` and `ci_status` are untouched.
- **`charter change push`** pushes to the clone's own origin, opens against the clone's own
  default branch (never a record field — a destination in a committed file is what §6.1
  forbids), and maintains a cross-link block between two markers. Absent, doubled, misordered
  or fenced markers → charter refuses to write, exactly as `render.splice_personas` returns
  `None` *"so a hand-written README is never appended to by surprise"*.
- **`charter change land`** takes one member: three gates, a merge, a read-back before
  anything is reported, the landing declaration written only for a merge charter confirmed,
  and a trace in a `finally` so refusals and exceptions are recorded as surely as successes.
- **There is no `--all`.** It does not parse; `--repo` is required so a bare `land` cannot
  mean "everything" by omission. It would have to answer what to do when member 3 of 5 is
  rejected mid-loop, and each available answer is wrong in a case the others handle. The
  shell loop somebody writes instead is five **gated** landings, because the gates are in the
  command. `--rebase` *does* parse and is refused with the reason — a rebase landing leaves
  charter no commit to carry the trailer and no single sha to revert.
- **The release floor is #639's, not mine.** Both branches had written it; theirs reaches
  the set through `_is_charter`, which catches `python3 -m charter` where a basename compare
  does not, so I dropped mine rather than merge a duplicate. `hooks.py`, `toolgate.py` and
  the `docs/hooks.md` bullet are out of this diff entirely.
- **ADR 0002 is amended**, not left quietly false: it said reporting was the only place
  charter writes to a forge. There are two seams now.
- **The suite refuses to reach a forge.** `tests/_planeguard` gains `RealForgeReach`, the
  `RealVaultReach` shape one host out, with the CLI names asked of `registry.KINDS`.

## The live #561 reading the exit criteria ask for

The suite can only prove charter reads a *fixture* correctly, and #561 is a claim about what
the forge does — so, measured against github.com on a real commit this repository has and CI
never ran on (a two-commit push: CI runs on the tip, the parent gets nothing):

```
raw check-runs   at 1590478  ->  total_count: 0        # GitHub HAS the commit
raw commit status at 1590478  ->  state: pending, total_count: 0

charter.checks_at 1590478  ->  total=0     state=NOT_RUN     # and it refuses the gate
charter.checks_at 1e5ca3a  ->  total=5     state=RUNNING     # the tip of the same push

statusCheckRollup at 1590478  ->  null                       # <- the whole bug
statusCheckRollup at 0986f47  ->  SUCCESS                    # a green head, for contrast
```

**`null` is what `ci_status` turns into `None` — the same `None` a rate-limited call, a
timeout, an auth failure and malformed JSON all produce.** Six worlds, one value, and the
merge button offered on all of them. `checks_at` answers `total=0, state=not_run` there and
`total=None, state=unknown` for every one of the others, which is the entire point of two
fields instead of one string.

**A second thing fell out of the live run, and it is why the criterion exists.** Probing a
sha GitHub does *not* have, the two endpoints disagree: `check-runs` answers **HTTP 422 "No
commit found"**, while the combined status answers **200 with `total_count: 0`**. A reader
trusting the status endpoint alone would call an unknown commit "no checks" — `NOT RUN` for a
commit that does not exist. Charter requires *both* reads to succeed, so it answers `unknown`
there; verified live. No fixture I wrote would have caught that, because I wrote the fixtures.

## Verification

**32 hand-written mutations, every one RED, every one verified applied and restored by
sha256** — the anchor must be found exactly once, the digest must change, and the restore
must return the file byte-for-byte or the harness aborts.

<details><summary>the 32</summary>

```
RED M1  total==0 renders as passed           RED M13  the blocker gate is skipped
RED M2  None and 0 collapse into one value   RED M14/15 not_run and unknown are passed
RED M3  read only check runs, no statuses    RED M16  landed reported without reading back
RED M4  an unknown conclusion maps to passed RED M17  the log line is written first
RED M5  a closed-unmerged request is "none"  RED M18  --rebase is permitted
RED M6  gitlab says not_run where it cannot  RED M19  the trace is dropped on refusal
RED M6b gitlab ignores a moved-on sha        RED M20  the _PUBLISH_FORGE entry is removed
RED M7  a write routes through _api          RED M21  the base check reverts, entry stays
RED M7b an unconfirmed merge returns a sha   RED M22  the _DANGEROUS word is removed
RED M7c -f becomes -F on a write             RED M23  the forge-CLI spawn guard is removed
RED M8  the base is read from the record     RED M23b that guard becomes a blanket allow
RED M9  an absent marker pair appends        RED M24  the per-file half of the log read
RED M9b a marker inside a fence is spliced   RED S1   the BEGIN marker is retuned
RED M10 containment is skipped               RED S2   the END marker is retuned
RED M11 the -- is dropped from the argv      RED S3   an unreached member is dropped
RED M12 --all is added                       RED S4   the sigil is hardcoded
```
</details>

**M2 survived first time, and that is the useful half.** Folding a failed check-runs read
into an empty list stayed green — the *statuses* read failed in the same fixture and produced
the `UNKNOWN` by itself, so the mutation was masked by its neighbour. The missing direction
is now pinned.

### `tools/sweep.py` — four real survivors, and a run that does not finish

**1,431 added lines under `charter/` → 443 mutations.** `sweep.over_budget(443)` says it in
its own words:

> *443 mutations is past the 224 that 8 shards can measure inside 40 minutes each. Every one
> of them is still swept — nothing here is dropped — but a shard may be cancelled at the job
> timeout, and a cancelled shard reports no verdict rather than a short one.*

So **read a cancelled shard here as unresolved, not as clean.** Locally on one box it reached
13 of 443 at ≈1/min, because a survivor costs a confirming run of the full suite and
survivors cluster in the function that has no test. **430 mutations are unexamined and I am
saying so rather than reporting a clean run.**

Of the 13 it reached, **four were survivors and every one was real** (S1–S4 above):

- **`paths.get(repo) or repo`** and **`sigils.get(repo, "#")`** — two fallbacks in
  `cross_link_block`, both *unreachable*: a repo only gets a number once its path and sigil
  are known. Three parallel dicts meant two lookups that could miss, so both grew a fallback
  nothing could reach. **Deleted, not pinned** — "equivalent mutant" and "dead code" are the
  same finding. One `seen` dict, one lookup, one outcome, and the `—` is a row a test asserts.
- **`body or ""`** in `splice_block` — the same value defended twice, since `change_body`
  already turns GitHub's `"body": null` into `""`. Deleted.
- **The `BLOCK_END` literal**, retuned, survived four modules *and* the whole suite — because
  every case in the file spelled it `commands_change.BLOCK_END`, so **the test's expectation
  moved with the thing under test.** That is the trap this project found three times in one
  branch this week; here is a fourth. Now pinned as literals, with the reason in the test:
  the markers are written into other people's request bodies and matched back, so retuning
  either orphans every block already published.

### A third finding: this suite hangs on a machine whose commits are signed

Not from the guard — from a run that stopped advancing at test 3,955 of 8,122 with an
`op-ssh-sign` child alive under it. The fixture repos in `test_change_push.py` run
`git commit`, which inherits the developer's `~/.gitconfig`; mine carries
`commit.gpgsign = true`, `gpg.format = ssh` and 1Password's `op-ssh-sign`, so every fixture
commit shelled out to 1Password and parked behind a biometric prompt.

That is **#546's failure exactly, one tool over**, and `util.run`'s own docstring predicted
it: its `GIT_TERMINAL_PROMPT=0` covers *"git's own prompts only — not a GUI credential
manager, and not an SSH signing agent, which is a separate way for a captured git call to
hang."* It is also invisible in CI, where there is no global config to inherit — green on the
runner, hung on the machine of anyone who signs their commits.

Fixed here by pinning `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` to `os.devnull` in the helper,
the idiom `test_doctor_forge_visibility.py` already uses.

**And it is not just mine.** There is no shared neutralizer — not in `tests/__init__`, not in
`_isolation`, not in `_planeguard` — so it is done per file or not at all: of the 41 modules
that shell out to `git commit`, **11 set it and 30 do not.** The fix is one line in the
`tests` package rather than 30 edits, but it changes how 30 modules see git config on the
strength of one green run, so it wants its own PR and its own review — the same reason I left
the 23-module `doctor` stub alone above rather than smuggling it in here.

### One way the guard made that flake likelier, found and fixed

Worth stating because it is the honest half. The first version of the forge tripwire called
the matcher **a second time** inside `Popen.__init__`, and that matcher resolves the program
against `$PATH` — so it doubled a `shutil.which` on **every subprocess the suite starts**.
The tmux tests are timing-sensitive by construction, and a guard does not get to become a
reason they flake. Both tripwires now share one matcher call against a union computed once at
install, and both still fire (`op item list` → `RealVaultReach`, `gh api …` →
`RealForgeReach`, `gh --version` → allowed).

### Two findings the suite guard produced, one still open

Running it flat reddened tests identically on 3.12 and 3.14:

1. **26 tests** ran `gh --version` through `doctor.check_forge_cli` — a local probe reaching
   nothing. Refusing it would have been the guard being wrong about its own subject.
2. **23 tests** ran `gh auth status --hostname github.com` through `doctor.run_all()`. **This
   suite validates the developer's own forge token against github.com, 23 times, every run.**
   It predates this branch and reads no repository; closing it means stubbing the preflight in
   23 modules, which is its own change and not a line smuggled into a forge feature. So it is
   allowed, the residual is written into the guard's docstring, and `auth login` / `auth
   token` are still refused — one opens a browser and waits, the other prints the credential.

Everything the change surface does — `api`, `pr merge`, `mr merge`, `repo`, `release` — is
refused, and that is pinned.

### Where the spec's plan does not survive contact

Reported rather than patched over:

1. **`checks_at(path, sha)` cannot satisfy §3.5's own GitLab correction.** Task 5 Step 1's
   test spells two arguments; Step 4b then needs the merge request's iid. The signature
   carries `number: int | None` — unused on GitHub, load-bearing on GitLab, and the reason
   GitLab's `not_run` is only reachable when the caller has a request.
2. **Task 6 lists two writes; Task 8 needs a third.** `merge_change` is not in the plan, and
   the alternative is `commands_change` shelling out to `gh` — a second seam with none of the
   protocol's disciplines. Same for `change_body`: Task 7 Step 5 asks charter to splice into
   an existing body and gives it no way to read one.
3. **Task 7 Step 5's refusal means charter can never annotate a request it did not open.**
   Implemented as written — charter's own `create_change` writes the markers in — but the
   consequence deserves naming: a hand-opened request gets the refusal with both markers
   quoted, and never a block.
4. **The block header says `request`, not `pull request`.** §4's example says the latter; §2's
   vocabulary rule says a member's PR/MR is a *request*, and one table can carry a GitLab row
   beside a GitHub one. The vocabulary rule wins.

## What deduplicating against #639 changed here

Four helpers were defined in **both** branches, in the same module, with different
signatures — and git merged them **without a conflict**: two `def _git`, the later one
silently winning. That is worth naming as a shape, not just fixing.

| | this branch had | adopted from #639 |
|---|---|---|
| `_git` | `(clone, args: list, check=False)` | `(clone, *args)`, one funnel, `GIT_TIMEOUT` |
| `_default_branch` | `origin/HEAD` only | `doctor._plane_default_branch` — `origin/HEAD`, then a local `main`/`master` |
| `_contains` | `(clone, sha) -> bool \| None` | `(clone, commitish, default) -> bool` |
| `_SHA_RE` | `[0-9a-fA-F]{7,64}` | `^[0-9a-f]{7,64}$`, lowercase, anchored |

Theirs are better in ways I would not have found: `_contains` takes the default branch
explicitly because every such question is *relative to* one, and a clone charter cannot
resolve a default for must contribute nothing rather than a guess — mine could not express
that, so `member_landed` now asks `_default_branch` separately and says *"charter cannot
resolve this clone's default branch"* where it used to fold that into a bare `False`. And
their `_SHA_RE` is anchored because Python's `$` matches before a trailing newline, so
`.match` admits `"e0c9d13\n"` — a newline on its way into a git argv, out of a file a hand
edit reaches. `member_landed` now holds the log's sha to that shape before it reaches git.

**Their static discipline changed my code, not just my call sites.** `test_commands_change.py`
reads every `_git` call off the syntax tree and requires the verb to be a *literal at the
call site*, so a helper returning a whole argv is a subcommand nobody is checking. `_push_argv`
is deleted and the push is spelled where it runs; what pins it is now stronger than the
helper test was — the entire argv, read statically (M11, M11b).

## Not in scope

Tasks 10–14 (revert, the frame surface, divergence, ADR 0020, the repo-run sweep), and Task 9
Step 5 (doctor naming `charter guard ask --local`), which belongs with the rest of Task 9.

## Checks

- [x] `python3 -m unittest discover -s tests` — **8,122 tests**, in two environments: 3.14,
      and 3.12 with `CHARTER_*`/`CLAUDE_*`/`ANTHROPIC_*`/`TMUX*` unset. Clean `main`
      measured alongside in its own worktree: **7,950, OK** — so this adds 172.
      **One caveat, measured rather than waved away:** some runs here show a failure in the
      tmux-driven modules (`test_frame_input_reaches_a_component`,
      `test_the_suite_reaps_its_own_tmux_servers`), and it moves between tests from run to
      run. It is **pre-existing and not from this branch**: the same test
      (`test_a_click_reaches_only_the_pane_it_landed_on`) fails on a clean `origin/main`
      worktree with none of this code present — 2 failures in 8 samples — on a machine
      carrying five live `charter` tmux servers, the oldest up five days. This branch has no
      diff under `charter/frame/`. See the note below for the one way it *did* touch them.
- [x] A test fails without this change — 32 of them, listed above.
- [x] Nothing reports success for a state it did not read back: `land` re-reads the request
      and prints the sha *that read* answered, and writes the landing log only after it
      (ADR 0013). A divergence charter can see, charter names — a member the forge calls
      merged with no log line is landed **and** reported as landed outside charter.
- [x] Docs updated in the same PR — `docs/forges.md` (the three disciplines, `checks_at`,
      the blind spot in each backend), `docs/hooks.md` (the new floor entry), ADR 0002's
      amendment.
- [x] `docs/news/unreleased-cross-repo-change-forge.md`, `version: unreleased`, no `check:`.
- [x] No new runtime dependencies — stdlib only, `unittest`, `gh`/`glab` as subprocesses.
- [x] Contradicts no ADR. It amends ADR 0002's *consequences* in writing, which §8.2 of the
      spec requires and §9 names as one of the two ADRs this phase touches.

**No version bump, no stamp, no tag, `release.yml` untouched.**


diazoxide:
Superseded by **#651** and **#652**, which are this PR split in two.

**Why.** Tasks 5–8 together are 1,393 added lines under `charter/` → **418 mutations**, against a deletion-sweep gate sized for 224. On this branch all eight shards were cancelled and the gate reported `no verdict: 8 of 8 shards did not report` — the one outcome it must never let pass quietly. Raising the ceiling or sharding finer would have made the *budget* bigger; splitting makes the *verdict* real. Measured, not guessed:

| | mutations | under the 224 ceiling |
|---|---:|---|
| #651 — Tasks 5–6, the forge protocol | **201** | yes, no over-budget warning |
| #652 — Tasks 7–8, `push` and `land` | **217** | yes, no over-budget warning |

**The stack, bottom to top:** #639 (Tasks 9–14, carries the release floor) → #651 → #652. Nothing in #651 has a caller, which is what makes it separately answerable; `charter change land` never exists without #639's floor beneath it.

Two other things that were in this branch are gone from both, because they landed elsewhere while this was open — and in both cases the other implementation was better than mine:

- **The release floor** (`hooks._PUBLISH_FORGE`, `toolgate._DANGEROUS`) is #639's. Theirs reaches the set through `_is_charter`, which catches `python3 -m charter` where my basename compare did not.
- **The suite's forge-token and signing-key guards** are #647's, which implemented both findings this branch filed (#638, #641) and went further: it *answers* the forge-auth probe locally rather than only refusing it, and measured 28 children across two forges where I had counted 23 on one.

Issue #564, opened 2026-08-26
Integration tests leak live tmux servers and their socket files, and nothing reaps them


diazoxide:
Found while verifying #554's fix, which added the missing `unlink` to the overlay hatch module's teardown. That fix is correct and merged — this issue is the wider leak it made visible.

## What is on this machine right now

```
$ ps -Ao pid,etime,command | grep 'tmux -L'
  28466  21:36:35     charter-integration-operator-21887
  19880  22:11:05     charter-integration-test-17119
  28800  21:36:33     charter-integration-test-22994
  64532  05:34:28     charter-integration-test-48505
  41377  1-20:49:47   charter-integration-test-40951
  93297  1-08:58:40   charter-integration-test-92517
  3442   1-10:52:47   charter-integration-test-95424
  80057  1-02:14:24   probe-menu-80053
```

**Eight live tmux servers left over from test runs**, the oldest running for nearly two days. Each is a resident process holding a session and a `cat`/`sh` child. They are not sockets on disk that nobody reopens — they are servers still running.

Separately, the socket directory:

```
$ ls /tmp/tmux-$(id -u) | wc -l
457
$ ls /tmp/tmux-$(id -u) | grep -c charter-overlay-hatch
317
```

317 dead `charter-overlay-hatch-*` files, one per historical run of `test_frame_overlay_escape_hatch.py` — the accumulated trace of the missing unlink that #554 fixed. None is held by any process (`lsof` → 0), so those are inert clutter rather than leaked servers.

The live leaks come from `tests/test_frame_tmux_integration.py`, which is the only module naming these sockets.

## Why it matters beyond tidiness

1. **A leaked server outlives the run that made it.** Two days of a resident tmux server per unlucky run is a real cost on a developer machine, and CI runners hide it by being ephemeral — so this only ever hurts the person running the suite locally, which is the person least likely to look.
2. **It undermines socket-name isolation.** The `-<pid>` suffix exists so concurrent runs do not collide. That works, but only because PIDs are eventually reused — at which point a *new* run can find a *live* server on the socket it just picked, holding sessions it did not create. That is a genuinely confusing failure, and it will present as an unrelated flake.
[...]

Issue #545, opened 2026-08-25
test_frame_launcher hangs forever when the suite is run from a terminal — _launch pins stdout's tty-ness and inherits stdin's


diazoxide:
`test_frame_launcher` hangs forever when the suite is run from a terminal — `_launch` pins stdout's tty-ness and inherits stdin's

#540 settled the rule: *the shell the suite was launched from is not a fixture*. One instance of it is still open, and it is not a wrong answer — it is a **permanent hang**.

## What happens

```
$ python3 -m unittest discover -s tests        # from an ordinary terminal, or inside a charter frame
...
test_the_frames_own_workspace_is_written_down_on_both_paths
(tests.test_frame_launcher.ALaunchFillsTheCacheItJustEmptied.…)
The other half of #512: a panel cannot resolve the workspace itself, so the ...
   ← blocks here, indefinitely
```

`sample(1)` on the stalled process: `builtin_input` → `PyFile_GetLine` → `_bufferedreader_raw_read`, with fd 0 on the pane's own `/dev/ttys037`. The suite is sitting at charter's own workspace picker prompt, waiting for a human who will never type.

Measured on `c1f9b95` (current `main`), reproduced by attaching a pty to stdin and nothing else. Same on a real tmux pane inside a live frame.

## Why

`commands_frame._picker_wanted` gates the #518 picker on **two** independent tty checks, by design:

```python
if not (sys.stdin.isatty() and sys.stdout.isatty()):
    return False
```

`tests/test_frame_launcher.py`'s two launch helpers pin the first of the pair and leave the second to the ambient shell:

* `_launch` (line ~1620) — `mock.patch("sys.stdout.isatty", return_value=True)`
* `_launch_inside` (line ~3873) — the same line

Nothing in the module mocks `sys.stdin.isatty`, and nothing sets `args.workspace` or `args.pick`. So the picker's answer is decided by whatever stdin the runner happened to have:

| stdin | `_picker_wanted` | result |
|---|---|---|
| pipe / `/dev/null` (CI, `... \| grep`, an agent's shell) | `False` | green |
| a terminal (a developer, a charter frame) | `True` | **hangs forever** |

CI has never seen it because CI's stdin is not a tty. It is invisible to exactly the environment that reports it, and unavoidable in exactly the environment a person uses.

[...]

Issue #542, opened 2026-08-25
The suite forks ~90 real charter _version-check children per run, each one a PyPI request — six modules stub it by hand and twelve do not


diazoxide:
Found while fixing #527. That issue was about **which plane** a test-spawned charter child
lands on, and it is closed: every child now carries a `$CHARTER_ROOT` naming a throwaway
plane, and `tests._planeguard.RealPlaneSpawn` refuses one that does not. This is the other
half — **how many children a unit test forks at all** — and it is filed separately because
the fix is a design decision, not a repeat of that one.

## What was measured

`subprocess.Popen.__init__` was wrapped for one full suite run (`python3 -m unittest
discover -s tests`, ambient `$CHARTER_SESSION_ID`/`$CLAUDE_CODE_SESSION_ID`/`$TMUX`/
`$TMUX_PANE` cleared) and every construction whose argv launches charter was logged:

```
174 charter children, of which:
  ~90  python3 -P -m charter _version-check      (update.maybe_spawn)
  ~36  python3 -P -m charter gl-refresh …        (glstate.maybe_spawn)
   …   the rest are deliberate CLI-driving tests
```

All of them now resolve a temp plane, so nothing touches the operator's state. But
`_version-check` is `update.fetch_and_store` — a **PyPI GET**, and on a dev plane a second
GET to GitHub. So a green suite makes ~90 outbound requests it never asserts anything about,
from ~90 processes it never waits for.

## Why they fire in a test and almost never in real use

Both `maybe_spawn`s are throttled by state in `config.STATE_DIR`. `PersonaIso` hands every
case a **fresh** temp state directory, so the cache is always absent and the cooldown lock
never exists — the throttles that make this rare on a real machine are exactly what a
throwaway plane removes. `test_statusline_brand` already says so in a comment: *"a temp
STATE_DIR always looks stale"*.

## The suite already knows it does not want this

Six modules stub it by hand, with the same comment twice over:

```python
spawn = update.maybe_spawn          # never fork a network child from the suite
update.maybe_spawn = lambda: None
self.addCleanup(lambda: setattr(update, "maybe_spawn", spawn))
```

`test_statusline_brand.py`, `test_statusline_gauge.py`, `test_statusline_inflight_chips.py`,
`test_statusline_session_block.py`, `test_statusline_plane_root_warning.py`,
`test_version_shows_channel.py`. Twelve other modules that call `statusline.render` do not.
[...]

PR #647, opened 2026-08-29
The suite stops spending the operator's forge token and their signing key (#638, #641)

### LINKED ISSUE #638: The suite validates the operator's real forge token against github.com 23 times per run
Found by the Phase 4 Tasks 5–8 agent while building a `RealForgeReach` guard in the shape of `RealVaultReach` (#546/#587).

## Measured

`doctor.run_all()` calls `gh auth status`, which **validates the operator's real token against github.com**. Reached **23 times per suite run**, identical on CPython 3.12 and 3.14.

## Why this is the #546 family

This project has spent the week closing "the suite reads the machine instead of the repo":

- the shell environment (#519/#521/#528), the plane (#402/#492), the filesystem (#529), the clock (#494), the working directory (#537)
- **the operator's real 1Password vault** — 4 `op` invocations from 2 tests (#546)
- **a real keyboard** — 126 blocking `input()` calls across 122 tests (#545)
- **PyPI** — 27 `_version-check` children per run, and 36 forge clients (#542)
- **leaked tmux servers** — 658 socket files, 14 live (#564)

This is the same shape, on a credential. It is not a leak — `gh auth status` does not print the token — but it is the operator's real authority reaching a real remote, from a test suite, on every run.

## And the multiplier is the part that matters now

`tools/sweep.py` runs the suite **once per mutation**, and the gate now runs on every pull request. 23 calls per run becomes 23 × the mutation count. #626's diff generated 78–82 mutations; that is roughly **1,800 authenticated requests to github.com from one PR's gate**, on the operator's own token.

#542 made exactly this argument about PyPI and it held: whatever the suite spends per run, a sweep spends again per mutation.

## What the finder did, and what is left

They allowed `auth status` in the new guard, refused `auth login` and `auth token`, and **wrote the residual into the guard's docstring** rather than hiding it. Closing it properly means stubbing the preflight in **23 modules**, which was correctly out of scope for a Phase 4 branch.

So the guard is honest and the reach is still there. That is the right state to file from, not to leave undone.

## Suggested

The six modules that already stub their providers are the prior art (#542 names them). The preflight is one call site in `doctor`; a fixture that answers it without a network call should serve all 23.

### LINKED ISSUE #641: The suite hangs on any machine with signed commits: 30 of 41 modules run git commit with no shared neutralizer

diazoxide:
Two issues, one pull request, because they are one defect on two credentials: the suite
spends the operator's forge token, and the suite spends the operator's signing key. Same
family as #546/#587 (the 1Password vault), #545 (a real keyboard), #542 (PyPI and forge
clients), #564 (leaked tmux servers) — *the suite reads or spends the machine instead of
the repository* — and the same prior art: **one fixture that answers, one tripwire that
refuses, both installed before any test module is collected.**

They are together rather than split because they share the file that refuses
(`tests/_planeguard.py`), the file that installs (`tests/__init__.py`) and the argument
for the shape. Splitting them would have meant two conflicting edits to the same wrapper.

---

## #638 — the forge token

`doctor.check_forge_auth` runs `gh auth status --hostname github.com`. Not a leak —
`auth status` never prints the token — but it is the operator's real authority reaching a
real remote from a unit test, on every run.

### Re-measured, and the issue's number was wrong

The issue said 23, all `gh`. Counted in-process (`Popen.__init__` wrapped before `tests`
is imported, because a `ps | grep` of a running suite matches its own command line — how
#542's "~90" turned out to be an artefact):

| child | before | after |
|---|---|---|
| `glab auth status --hostname gitlab.com` | 20 | 0 |
| `gh auth status --hostname github.com` | 8 | 0 |
| **forge-auth children per run** | **28** | **0** |

From **18** test modules, identically on CPython 3.12 and 3.14. The split is a property of
the fixtures, not the machine: a plane declaring no `[[forge]]` block falls back to the
single-GitLab default, and every plane `PersonaIso` hands out is one of those — so a fix
aimed only at the `gh` the issue named would have left twenty of the twenty-eight.

At 28 per run, the sweep gate spends that again per mutation: roughly **2,200 authenticated
requests from one pull request's gate** on a diff the size of #626's.

### Answered at the child, not at the check

`tests/_forgeprobe.py` wraps `charter.util.run`, recognises `<forge cli> auth status` and
nothing else, and returns a recorded reply on stderr — where `gh auth status` really writes
it, so `check_forge_auth`'s `(stdout or "") + (stderr or "")` stays load-bearing.

`mock.patch("charter.doctor.check_forge_auth")` would have been one line and would have
left that function's `"Logged in"` branch, its summary extraction, its concatenation and
its `ProcTimeout` arm dark for the whole suite — the trap #542 names. Every one of those
lines still executes, on a recorded reply. A test that wants a different answer patches
`charter.util.run`, which replaces the fixture and puts it back.

### And the allowance came out on the same commit as the reach

`RealForgeReach` (from the Phase 4 branch, converged here) shipped **allowing**
`auth status` and wrote the residual into its own docstring. It is now refused with
everything else. The rule is a one-item allowance, not a denial list: an argv naming no
subcommand at all (`gh --version` — the local probe reporting which CLI is installed) is
allowed; everything else, `api`/`pr merge`/`mr merge`/`repo clone`/`auth status` and any
subcommand charter has never heard of, fails the test that spawned it before the child
starts.

A guard that permits what nothing does any more is a guard that will quietly re-permit it.

**Residual, stated:** `gh --version` still runs 28 times a run, and `shutil.which` still
reads this machine. Both are local, and answering `which` would foreclose the tests about a
CLI that is not installed.

---

## #641 — the signing key

A fixture repository inherits `commit.gpgsign = true` from the machine's own
`~/.gitconfig`. With 1Password's `op-ssh-sign` behind it, `git commit` parks on a biometric
prompt. Not a failure — a **hang**: no pass, no fail, no verdict, no line saying why.

Reproduced on this machine, in a bare temp repository, with charter's own checkout (which
carries a local `commit.gpgsign=false`) out of the picture:

```
$ git init -q . && echo x > a && git add a && git commit -m probe
error: 1Password: failed to fill whole buffer
fatal: failed to write commit object
git commit -m probe  0.01s user 0.01s system 0% cpu 1:00.36 total
```

60.2 s and a failed commit with stdin closed; indefinite with a terminal attached. **A
runner has no signing config**, so CI cannot see this and never will.

### Measured

| | before | after |
|---|---|---|
| `git` children per run | 9,735 | 9,761 |
| …reading the operator's `~/.gitconfig` | **9,319** | **0** |
| `git commit` children per run | 1,384 from 31 modules | — |

Of those 1,384 commits, 1,068 carried `-c commit.gpgsign=false` on the argv, 208 relied on
a repo-local `git config` written in one module's `setUp`, and the rest on
`GIT_CONFIG_GLOBAL=/dev/null` in a hand-built environment. Three spellings, no shared
answer, and a module that remembers none of them does not fail — it hangs.

### One redirect, then a tripwire for what the redirect cannot reach

`tests/_gitguard.py` points `$GIT_CONFIG_GLOBAL` at a file this repository writes and
`$GIT_CONFIG_SYSTEM` at `os.devnull`, installed **above the import that pulls charter in**
(`charter.config` resolves a plane at import and reads a git worktree pointer on the way).
Signing off, an identity at a reserved `.invalid` domain, `init.defaultBranch = main`, no
global hooks path, and none of the machine's filter drivers — this machine has
`filter.lfs.*` globally, which every fixture repo was inheriting.

`_planeguard.AmbientGitConfig` then refuses a `git` spawned with an environment that drops
it. **Turning it on across 7,984 tests found exactly one module**: ten cases in
`test_frame_owns_the_surface`, which builds its environment with `clear=True` and so handed
`git remote get-url origin` an empty one. Those ten really were reading the operator's git
config; they now carry the redirect explicitly.

The suite also gains the thirty-second module on purpose — a fixture repo, a plain
`git commit`, no neutraliser of its own — which took 60.2 s and failed here before the
redirect and runs in 2.7 s after it, asserting the commit is unsigned, authored by the
suite, and on `main`.

---

## Verification

Full suite, before and after, in two environments — CPython 3.14 with the operator's own
shell, and CPython 3.12 with `CHARTER_*`/`CLAUDE_*`/`ANTHROPIC_*`/`TMUX*`/`COLUMNS`/`LINES`
removed (asserted, not assumed, before each run):

Deletion sweep over the diff: 

Both legs ran on a machine **with signing configured** (`commit.gpgsign = true`,
`gpg.format = ssh`, `gpg.ssh.program = op-ssh-sign`) — the environment #641 only reproduces
in, and the one CI can never provide.

Closes #638
Closes #641

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,6 +50,13 @@ command.
   stdout — `os.get_terminal_size()` answers what a pipe answers, so a render is the same
   width in your window as it is in CI. A test that wants a size states one with
   `mock.patch("os.get_terminal_size", …)`.
+  `tests/_gitguard.py` is the fifth, for the one config file charter never writes and every
+  `git` it spawns reads: **your own `~/.gitconfig`.** `$GIT_CONFIG_GLOBAL` points at a file
+  this repository writes, so a fixture repo commits as the suite, on `main`, and — the part
+  that matters — **unsigned**. Inheriting `commit.gpgsign = true` with a hardware or
+  1Password signer behind it does not fail a test, it *hangs* it, and no runner can ever
+  see that. A test that builds a child environment by hand carries
+  `tests._gitguard.environment()` into it, or is refused by name at the spawn.
 - **The suite spends nothing it was not asked to.** `tests/_planeguard.py` also refuses a
   charter child started with `start_new_session=True` — charter's own shape for a
   background refresh that outlives the process that started it. Those fire in tests and
@@ -60,7 +67,12 @@ command.
   `tests._isolation.no_background_refresh(self)`; if it is *about* the child, call
   `tests._planeguard.allow_background_children(self)`. A test that starts a real tmux
   server names its socket with `tests._tmuxreap.name("<slug>")`, so the next run can reap
-  it when this one is killed before its cleanup runs.
+  it when this one is killed before its cleanup runs. **Nor does it spend a credential.**
+  `doctor`'s preflight asks a forge whether your token is still good, and eighteen modules
+  reach that line — 28 authenticated round trips to github.com and gitlab.com per run, and
+  the sweep gate spends that again per mutation. `tests/_forgeprobe.py` answers the probe
+  with a recorded reply, and `tests._planeguard.RealForgeReach` refuses every other `gh` or
+  `glab` invocation: a forge test drives `charter.util.run` with a recorded reply instead.
 - **Comments explain *why*.** The codebase leans hard on this: a comment that restates the
   code earns nothing, one that records the failure a line prevents is worth several
   paragraphs of docs. Read a few modules before writing your first one.
@@ -148,18 +160,22 @@ outside the test — git config, `$PATH`, locale, the default shell, an env var,
 filesystem's case sensitivity — is either set explicitly by the test or is a bug waiting
 for someone else's machine.
 
-For git specifically, pin it on the invocation rather than trusting the environment:
+**For git specifically, that is now done for you** — and the fifth entry above is the
+reason it had to be. Thirty-one modules ran `git commit`, 1,384 children in one green run,
+each protected by hand in one of three different spellings or not at all; the module that
+remembers none of them does not fail, it hangs. `tests/_gitguard.py` points
+`$GIT_CONFIG_GLOBAL` at a file this repository writes — `init.defaultBranch = main`,
+signing off, no global hooks path, none of your filter drivers — before any test module is
+collected, and `tests._planeguard.AmbientGitConfig` refuses a git child that steps outside
+it. So a plain `subprocess.run(["git", "commit", …])` in a fixture repo is now correct on
+every machine, and the per-invocation `-c commit.gpgsign=false` several modules still carry
+is belt and braces rather than the only thing standing between the suite and a biometric
+prompt.
 
-```python
-_PINS = ["-c", "init.defaultBranch=main", "-c", "commit.gpgsign=false",
-         "-c", "tag.gpgsign=false"]
-subprocess.run(["git", *_PINS, "-C", str(cwd), *args], ...)
-```
-
-Signing belongs in that list even when the test has nothing to do with signing: a developer
-with a signing helper configured gets a fixture commit that stops to ask for a passphrase,
-and a suite with nobody at the keyboard hangs rather than fails. That is the same failure
-`docs/git-policy.md` exists to prevent in charter itself.
+Signing was always the one worth the paranoia: a developer with a signing helper configured
+gets a fixture commit that stops to ask for a passphrase, and a suite with nobody at the
+keyboard hangs rather than fails. That is the same failure `docs/git-policy.md` exists to
+prevent in charter itself.
 
 ### Reproducing the runner before you push
 
@@ -169,6 +185,9 @@ You can hand git a config the way the runner has it, without touching your own:
 GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=init.defaultBranch GIT_CONFIG_VALUE_0=master \
   python3 -m unittest discover -s tests
 ```
+
+`GIT_CONFIG_COUNT` is applied as if by `git -c`, so it still wins over the suite's own
+`$GIT_CONFIG_GLOBAL` redirect — the recipe means what it says.
 
 If a change touches git behaviour, run the suite that way once. It is cheaper than a red
 `main`, and it is how the last one was diagnosed.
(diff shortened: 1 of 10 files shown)


diazoxide:
Merging on the CI gate rather than on the author's local sweep, which died with the session (weekly limit, resets Sep 1).

**What is verified**: CI green on 3.11–3.14, and the sweep gate — scoped to the lines this branch adds — reads `no survivors`. The author reported the code final and rebased onto `f6d42f0` before stopping.

**What is not**: the author's own `tools/sweep.py --path tests --path charter` run over the whole diff (70 mutations, baseline green at 8073 tests) never finished. **The gate's scope is added lines only**, so a broader sweep asks a question the gate structurally cannot. That reading is outstanding, not clean.

Recording it here rather than implying the branch is fully swept.

Issue #648, opened 2026-08-29
test_frame_input_reaches_a_component flakes under load, and three agents have now spent time proving it is not theirs


diazoxide:
`tests/test_frame_input_reaches_a_component.py` arrived with #636 (click and scroll). It flakes on a loaded machine, and because it is a **new** module every branch in flight has had to prove the failure is not its own.

## Evidence it is the test, not any branch

**Same sha, opposite outcomes in two workflow runs** — read from `check-runs` on #644's head:

```
test (3.11): success   test (3.11): success
test (3.12): FAILURE   test (3.12): success     <- same commit
test (3.13): success   test (3.13): success
test (3.14): success   test (3.14): success
```

Independent reports from three different agents, on three different branches:

- **#644's author**: the `SIZE-0`/`SIZE-1` tmux assertion. Reproduces on a **clean `origin/main` worktree — 2 failures in 8 samples** — moves between tests run to run, and passes in isolation. Their branch has no diff under `charter/frame/`.
- **#643's author**: `test_a_focus_change_reaches_only_the_pane_it_is_about` failed on 3.12 locally. `origin/main` fails the identical case on the same machine; CI was green at their head; an earlier run on a quieter box was clean.
- **The coordinator**: ran the focus case in isolation on `main` just now — passes.

## Why it matters beyond the noise

Every one of those agents did the right thing: reproduced on `main`, checked isolation, compared runs at the same sha, and **said so rather than re-running until green**. That is the discipline this repo wants. But it cost each of them real time to establish the same fact independently, and it makes every merge decision a judgement call about whether a red is real.

There is also a merge hazard in it. Tonight a `mergeStateStatus` and a `gh pr checks` reading were both distrusted for exactly this reason (#561), and a flaky new module is the other way the same confusion arrives: a genuine regression in that file would now be indistinguishable from the known flake.

## What it looks like

Load-sensitive, tmux-backed, and it involves pane geometry (`SIZE-0`/`SIZE-1`) and focus delivery. This project has closed several of that shape already, and the measurements are on record:

- `kill-server` returns 0 with its socket **still bound** — measured window median 0.4 ms, max 1.3 ms. #554 passed 12/12 locally while CI was red at the same head for this reason.
- A **background tmux window keeps stale geometry** until `select-window` — measured on 3.7c and 3.2 (client 200×50 → 100×30, background window stayed 200×50).
[...]

PR #649, opened 2026-08-29
The artifact round-trip the sweep depends on is bumped in both workflows, not one (#480, #478)

diazoxide:
Supersedes #480 and #478.

Dependabot opened those two for `upload-artifact` v4.6.2 → v7.0.1 and `download-artifact` v4.3.0 → v8.0.1, and **both touch `release.yml` alone**. They predate `sweep.yml`, which uses the same pair for the thing the gate's verdict is made of — each shard uploads `deletion-sweep-<n>`, and `collect` downloads them back with `pattern:` + `merge-multiple: true`. Merging either as filed leaves the gate on v4 and the release on v7/v8.

## Neither PR could have tested what it changed

`sweep.yml` and `release.yml` are the only two files using these actions. On a workflows-only PR the sweep gate does not report at all, so #480 and #478 each went green on `test (3.11–3.14)` plus a skipped install check — nothing that touches an artifact. `release.yml`'s round-trip runs only during an actual release.

## So the inputs were read at the pinned shas

| pass | where | v7/v8 |
|---|---|---|
| `name` | sweep upload, release upload | present |
| `path` | both | present |
| `if-no-files-found` | sweep upload | present |
| `pattern` | sweep collect | present |
| `merge-multiple` | sweep collect | present |

**`merge-multiple` is the one that mattered.** `collect` reads a flat `shards/` directory. Had v8 dropped it, the download would have *succeeded* into a nested layout, `--verdict shards` would have found nothing, and every pull request would have read `no verdict` while every job stayed green — the #626 failure arriving by upgrade instead of by cancellation.

Both actions are `node24` at these versions. Tag→sha verified against `actions/*` directly; `tests/test_workflows.py` (45 tests) green.

## What is proven and what is not

The gate now tests itself: **the next PR touching Python** runs shards through v7/v8 and either produces a count or says `no verdict`. Four such PRs are in flight (#639, #644, #645, and #648's).

`release.yml` is **not** directly exercised — same action pair, used the same way, but the release itself remains its first real run. Worth a glance at the artifact steps when we cut it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -165,7 +165,7 @@ jobs:
         run: pipx run build
       - name: Validate metadata renders on PyPI
         run: pipx run twine check dist/*
-      - uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+      - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: dist
           path: dist/
@@ -185,7 +185,7 @@ jobs:
       # from the same run with the runtime token, needing no `GITHUB_TOKEN` scope at all.
       id-token: write        # REQUIRED: mints the OIDC token PyPI verifies
     steps:
-      - uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
+      - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
           name: dist
           path: dist/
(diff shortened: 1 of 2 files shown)


diazoxide:
**Correction to the PR body: the round-trip is directly proven, on this PR.**

I wrote that the gate would only test itself on "the next PR that touches Python." That was wrong — the sweep ran here:

```
success  Size the sweep
success  Sweep shard 1 of 1        <- uploaded deletion-sweep-1 at v7.0.1
success  Add up what the shards found  <- downloaded it at v8.0.1, pattern + merge-multiple
success  no survivors              <- a real count, not `no verdict`
```

`collect` found the flat `shards/` layout and `--verdict shards` read it. That is the `merge-multiple` question answered by execution rather than by reading `action.yml`.

It also explains why #480 and #478 showed no sweep check: their last push **predates `sweep.yml`**, which is the same reason their diffs touch `release.yml` alone. Two symptoms, one cause — they are simply stale.

`release.yml` remains covered only by the same-pair-same-usage argument, since nothing short of a release runs it.

Issue #646, opened 2026-08-29
sweep.yml runs on pull_request only, so a branch with no gate run looks identical to one whose gate found nothing


diazoxide:
Found by #630's own author, on their own workflow, and flagged rather than quietly patched.

## What happens

`sweep.yml` triggers on `pull_request` and `workflow_dispatch` only. **Two pushes to a branch produced no `synchronize` event, so the gate did not run — and nothing said so.** The checks list simply had no sweep row.

## Why this is the same defect the gate was built to fix

#617 was: *a green check was not "no survivors"*. #630 fixed it by making the check's **name** carry the verdict — `no survivors`, `114 survivors, 1 not measured`, `no verdict: 2 of 3 shards did not report` — because GitHub's conclusion vocabulary cannot express it.

This is that silence one level up. **A branch with no gate run looks identical to one whose gate found nothing**, because there is no row at all. The three states #630 carefully distinguished are undermined by a fourth nobody can see.

## And it is #561, which is still open

> **A check that never ran reports CLEAN**, and nothing on `main` requires it to have run.

#561 is about branch protection: a swallowed trigger makes a PR report `mergeStateStatus: CLEAN` and offer the merge button. That nearly merged an unverified branch here, and the lesson recorded from it was to read `gh api .../commits/<HEAD_SHA>/check-runs` rather than `gh pr checks`, because the former cannot confuse "none" with "green".

**The gate now has the same hole in its own trigger.** Reading check-runs per head sha does not help if the workflow never fired for that sha.

## Options, and why this is a decision rather than a patch

Adding `push:` closes it, but that is a decision about **when the gate should run**, not a one-line fix:

- The gate costs 8 parallel jobs and up to 40 minutes on a large diff (measured on #639: 443 mutations, 8 shards, 442 measured). Running it on every push to every branch is a different cost profile from running it per pull request.
- `pull_request` gives it a merge-base to diff against, which is what "scoped to added lines" means. A push to a branch with no PR has no obvious base.
- If it runs on push **and** pull_request, the same sha gets swept twice.

The narrower fix may be to make the **absence** legible instead: require the sweep check on `main` (which is #561's remedy), so a missing row blocks rather than passes silently.

## Related

- **#561** — the same property at the repository level; still open, needs a repo-settings change.
- **#630** — made the gate's verdict legible; this is the gap that verdict falls through.

diazoxide:
Closed by #679 and #561's branch protection together.

[...]

PR #659, opened 2026-08-29
A sweep a newer push cancelled says it was superseded, not that it found nothing (#654)

### LINKED ISSUE #654: A cancelled sweep publishes `no verdict` that outlives the run it describes, and reads exactly like a real one
## What happens

`sweep.yml` sets `concurrency: cancel-in-progress: true` so a pull request that gets three pushes pays for one sweep. When a run is cancelled that way, its `collect` job still executes — deliberately, under `if: always()`, because "a shard did not finish" is the case that job exists for — and publishes a check whose **name is the verdict**:

```
no verdict: the sweep never sized itself
```

That check stays on the sha. A later successful run publishes a *differently named* check (`81 survivors`, `no survivors`), so the two do not replace each other — GitHub shows both, and the stale one is indistinguishable from a genuine `no verdict`.

## Measured, on #652 at `74ccfdab40c2`

```
completed/cancelled  deletion sweep  run=86  2026-08-29T08:31:53Z
in_progress/-        deletion sweep  run=87  2026-08-29T08:31:53Z
```

Two runs created **in the same second**; 87 cancelled 86 by concurrency group. Run 86's checks on that sha:

```
cancelled  Size the sweep
cancelled  Sweep shard ${{ matrix.shard }} of ${{ needs.plan.outputs.shards }}
success    Add up what the shards found
success    no verdict: the sweep never sized itself     <- run 86, superseded
in_progress Size the sweep                              <- run 87, the real answer
```

Note the shard job name is unexpanded — `needs.plan.outputs.shards` was empty because `plan` was cancelled before it produced an output. That is the same emptiness `collect` reads, and it reads it correctly. **The job is right; the problem is that its answer outlives the question.**

## Why this matters rather than being cosmetic

`no verdict` is the one outcome the gate is built to make loud — #617's whole sentence, and the reason the verdict is carried by a check *name* rather than a conclusion. #644 carried a genuine `no verdict: 8 of 8 shards did not report`, and that was true and actionable; it led to the PR being split into #651 and #652.

A cancelled run manufacturing the same sentence dilutes exactly that signal. A reviewer seeing `no verdict` on a pull request cannot tell, from the check list alone, whether the sweep failed to answer or was simply superseded by their own next push. The spec's stated worry is a gate that gets disabled the first time it is inconvenient; a gate that cries `no verdict` on every second push is how that starts.

## The distinction to draw

`collect` currently cannot tell "nobody sized this" from "I was cancelled mid-size". Those are different facts and only the first is a finding. Some options, none of them obviously right:

- Skip `collect` when the run's own status is `cancelled` — but `if: always()` is load-bearing for the real case, so this needs to distinguish *this run was cancelled* from *a shard was cancelled*, which is #626 and must still report.
- Name the superseded verdict differently (`superseded: a newer push cancelled this sweep`) so it is still visible but not confusable.
- Keep the name stable across outcomes and carry the verdict in the summary — rejected once already, for the reason the workflow header gives at length: a stable name means the answer is only readable by opening something.

The second looks closest to the file's existing philosophy: say the true thing in the name, and let the name be different when the fact is different.

## Not urgent

Nothing is blocked by this. #652's run 87 is producing the real verdict. Filing so it is not rediscovered as a mystery by whoever next sees `no verdict` on a green branch.

diazoxide:
Closes #654.

`sweep.yml` sets `concurrency: cancel-in-progress`, so a PR that gets three pushes pays for one sweep. The runs it cancels still reach `collect` — which runs under `always()` **deliberately**, because *"a shard did not finish"* is the case that job exists for — and there it read its own empty `needs.plan.outputs.shards` and published:

```
no verdict: the sweep never sized itself
```

True of that run, and indistinguishable from the sentence #617 built this gate to say. The verdict is carried by the check's **name**, so two different names do not replace each other and **both rows stay** — a branch shows `no verdict` beside a perfectly good `81 survivors`, with nothing to say which describes it.

## Measured, on #652 at `74ccfdab40c2`

```
completed/cancelled  deletion sweep  run=86  2026-08-29T08:31:53Z
in_progress/-        deletion sweep  run=87  2026-08-29T08:31:53Z
```

Same second; 87 cancelled 86 by concurrency group; 86's `collect` completed and published its verdict beside 87's real one.

## The distinction, and why `cancelled()` is the right one

`cancelled()` is a fact about the **run**, not about a job. A shard that exceeds its own `timeout-minutes` is a cancelled *job* inside a run that was never cancelled — so **#626's sentence is untouched**: `no verdict: 8 of 8 shards did not report` still means what it meant, and #644 would still have been split on it. Only a superseding push reaches the new step.

The step **reaches for nothing** a cancellation would have skipped: no checkout, no interpreter, no artifacts — `printf` into `$GITHUB_OUTPUT` and nothing else. The single state it exists for is the one where every step above it was cancelled, so a step that needed the tree could not run in it.

`superseded`'s `if` and `say`'s are **exact negations**, which is what keeps the `||` in `outputs:` a choice between one value and one empty string rather than a precedence anybody has to reason about.

## Verification

Four tests in `TheWorkflowSaysTheAnswerWhereItCanBeSeen` (which reads the workflow as a parsed tree, not by grepping). Hand-swept — each mutation applied and restored with **sha256 verified both ways**:

| mutation | |
|---|---|
| drop `superseded`'s `if: cancelled()` | **RED** |
| drop `say`'s `if: ${{ !cancelled() }}` | **RED** |
| drop the `headline` fallback | **RED** |
| drop the `conclusion` fallback | **RED** |
| retune the word to `no verdict:` | **RED** |

The last one is the one that matters: the test asserts the superseded words are **not** the unanswered ones, checked against `verdict`'s own fallback string so the two cannot drift together.

## Still open, deliberately not fixed here

**#646** — the gate runs on `pull_request` only, so a branch whose trigger was swallowed has *no* sweep row at all, which reads the same as one that passed. That is the same silence one level up, and its own issue says why it is a decision (cost profile, merge-base, double-sweeping) rather than a patch. Its narrower remedy is branch protection, which is **#561** and needs a repo-settings change.

## Why this is worth fixing rather than shrugging at

`no verdict` is the one outcome this gate exists to make loud. #644 carried a real one and it was true, actionable, and led to that PR being split into #651 and #652. **A gate that cries the same sentence on every second push is how a true one stops being read.**

### DIFF
--- a/.github/workflows/sweep.yml
+++ b/.github/workflows/sweep.yml
@@ -205,27 +205,36 @@ jobs:
       # verdict beside 87's real one. Both rows stay, because the verdict is the check's
       # NAME and two different names do not replace each other.
       #
-      # `cancelled()` is the run-level fact and is exactly the distinction wanted: a shard
-      # that exceeds its own `timeout-minutes` is a cancelled JOB in a run that was never
-      # cancelled, so #626 still reports `no verdict: N of M shards did not report` and
-      # only a superseding push reaches this step.
+      # **`cancelled()` is NOT the discriminator, and that was measured rather than
+      # reasoned.** The first version of this step used it and did not fire: run 99 on this
+      # branch was cancelled by run 100 (`conclusion=cancelled`, `Size the sweep` cancelled,
+      # the shard cancelled) and `collect` still took the `say` branch and published
+      # `no verdict: 1 of 1 shard did not report` — the #626 sentence, from a run where no
+      # shard was ever asked to report. In a job running under `always()`, `cancelled()`
+      # does not answer for the run the way the docs' one-line summary suggests.
+      #
+      # `needs.plan.result` does. It is `cancelled` exactly when the sizing job did not
+      # finish, and that is the only state in which the shard arithmetic below is answering
+      # about shards nobody planned. #626's own case is untouched: a shard that exceeds its
+      # `timeout-minutes` leaves `plan` SUCCEEDED, so `no verdict: N of M shards did not
+      # report` still means what it meant and still blocks nothing.
       #
       # It depends on nothing a cancellation would have skipped — no checkout, no artifact,
       # no python. `printf` into `$GITHUB_OUTPUT` and nothing else, because a step that
       # needed the tree could not run in the one state it exists for.
-      - name: A superseded run says so rather than claiming a verdict
+      - name: A sweep that never sized itself says that, and nothing more
         id: superseded
-        if: cancelled()
+        if: ${{ needs.plan.result == 'cancelled' }}
         run: |
           set -euo pipefail
-          printf 'headline=superseded: a newer push cancelled this sweep\n' >>"$GITHUB_OUTPUT"
-          printf 'conclusion=superseded\n' >>"$GITHUB_OUTPUT"
-      # `!cancelled()` and not the default `success()`: the download above is
-      # `continue-on-error` precisely so that "no artifacts at all" reaches this step, and
+          printf 'headline=cancelled: this sweep did not size itself\n' >>"$GITHUB_OUTPUT"
+          printf 'conclusion=cancelled\n' >>"$GITHUB_OUTPUT"
+      # The exact negation of the step above, and not the default `success()`: the download
+      # is `continue-on-error` precisely so that "no artifacts at all" reaches this step, and
       # a condition that stopped at the first failure would take that back.
       - name: One answer for the whole sweep
         id: say
-        if: ${{ !cancelled() }}
+        if: ${{ needs.plan.result != 'cancelled' }}
         env:
           # Empty when the plan job never finished — which `expected_shards` reads as "the
           # sweep never sized itself", not as "no shards were needed". Those two are the
--- a/.github/workflows/sweep.yml
+++ b/.github/workflows/sweep.yml
@@ -171,8 +171,11 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     outputs:
-      headline: ${{ steps.say.outputs.headline }}
-      conclusion: ${{ steps.say.outputs.conclusion }}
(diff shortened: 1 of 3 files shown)


diazoxide:
## Measured, both ways, on this branch

The fix is proven by the same method that disproved the first attempt — a real superseding push, not a reading of the docs.

**Run 99, with the first version (`cancelled()`):**
```
conclusion=cancelled
  cancelled  Size the sweep
  cancelled  Sweep shard ${{ matrix.shard }} of ${{ needs.plan.outputs.shards }}
  success    Add up what the shards found
  success    no verdict: 1 of 1 shard did not report      <- #626's sentence, no shard ever asked
```
**`cancelled()` did not fire.** In a job running under `always()` it does not answer for the run the way its one-line summary suggests.

**Run 103, with `needs.plan.result`:**
```
conclusion=cancelled
  cancelled  Size the sweep
  cancelled  Sweep shard ${{ matrix.shard }} of ${{ needs.plan.outputs.shards }}
  success    Add up what the shards found
  success    cancelled: this sweep did not size itself    <- fires, and claims nothing it cannot see
```

Same cancellation, same job conclusions, different sentence — which is the whole point, because #626's sentence has to keep meaning what it means.

## Why the wording is narrower than the issue asked for

#654 proposed `superseded: a newer push cancelled this sweep`. That claims more than the check can observe: **`plan` carries `timeout-minutes: 30`**, so a cancelled sizing job has two possible causes and only one of them is a superseding push. The check names the observable fact instead. Either way the cancelled `Size the sweep` job sits directly beside it, so a timeout is not hidden by the wording.

## The mistake is pinned

Seven mutations, all RED, sha256-verified on apply and restore — including **regressing the condition back to `cancelled()`**, so the error this PR corrects cannot return silently. That one matters more than the rest: it is the difference between a guard that is tested and a guard that merely looks right.

PR #662, opened 2026-08-29
The frame's arithmetic stops reading the plane's own `charter.toml`, so `size = 15` can be committed (#661)

### LINKED ISSUE #661: `repos_rows` reads the plane's own `charter.toml`, so an operator who uses #660's `size` turns the suite red
**#660 shipped a feature that cannot be used on the plane that ships it.** Writing the documented config turns six tests red, and because `charter.toml` is a tracked file in this repository, committing it would turn **CI** red for everyone.

## Reproduce

Add the one line #660's own news entry tells an operator to add:

```toml
[[frame.component]]
use  = "repos"
size = 15
```

```
FAIL: test_a_small_plane_gets_exactly_the_rows_its_content_wants   (test_frame_layout.ReposIsSizedToItsContent)
FAIL: test_it_never_goes_below_one_row (content=0)                 (test_frame_layout.ReposIsSizedToItsContent)
FAIL: test_it_never_goes_below_one_row (content=1)                 (test_frame_layout.ReposIsSizedToItsContent)
FAIL: test_a_size_map_is_what_panel_argvs_splits_with              (test_frame_layout.SlotSizesAnswersEverySlotAtOnce)
FAIL: test_left_is_not_a_slot_this_module_will_size_or_split       (test_frame_layout.PanelGeometry)
FAIL: test_the_frame_it_splits_is_byte_for_byte_the_frame_it_split (test_component_id_is_the_currency.CharterOwnConfigIsUnchanged)
```

Delete the line and all six pass. Measured on `main` at `9e6df55`, 8590 tests, `env -u PYTHONSAFEPATH`.

## The mechanism

```
layout.repos_rows(content_rows=4, window_rows=50, slots=["top","bottom","repos"])  ->  15
```

Four rows of content, a 50-row window, neither bound binding — and the answer is 15, because `repos_rows` now consults `_arrangement()`, which reads the plane's live config. Its own docstring still says:

> **Pure arithmetic**, so it is pinned here with no tmux and no cache.

It is not pure arithmetic any more. The test passes explicit arguments and gets an answer that came from a file it never named.

## This is the tenth instance of one defect class

The suite reads or spends the machine: the shell environment, the plane, the filesystem, the clock, the cwd, `$COLUMNS`, PyPI, forge clients, tmux servers, the real 1Password vault, a real keyboard, the real GitHub token, a real signing key. **This one is new in kind, though** — the others read a machine that happens to differ. This one reads *the repository's own committed file*, so the failure is not environmental at all: it is deterministic, and it is triggered by following the documentation.

## Why the design choice was reasonable and still needs changing

#660 chose config-reading deliberately, and said so:

> Read from the resolved config rather than threaded through five signatures, for `_placed_here`'s reason word for word.

That reason is real — threading a value through five signatures to reach one leaf is its own defect. But the cost was not priced: **`repos_rows` was the module's one provably pure function, and its tests were written to that property.** `_placed_here` can read config because nothing asserts it as arithmetic; `repos_rows` cannot.

## Options

1. **Thread the pin through `slot_sizes` into `repos_rows`** as a keyword, defaulting to `None`. `slot_sizes` already reads the arrangement, so the read stays in one place and `repos_rows` goes back to being arithmetic. Five signatures is the objection; measure whether it is actually five.
2. **Keep the read, and give the tests a way to say which arrangement they mean** — a context manager or an injected arrangement. Honest, but it makes every future caller of a pure-looking function responsible for knowing it is not.
3. **Leave it and change the tests to expect the plane's config.** Rejected: it makes the suite's answer depend on a file any operator may edit, which is the defect above and not a fix for it.

**(1) is my recommendation** — it restores the property the tests were written against, and the "five signatures" cost should be measured rather than assumed.

## Related

- **#660** — shipped the feature.
- The nine prior instances of the suite reading or spending the machine.

Filed rather than patched around, and the `size = 15` line is **not** committed to this plane's `charter.toml` until this is fixed.

diazoxide:
Closes #661.

## Reproduced first, on `origin/main` at `9e6df55`

Adding the one line #660's news entry tells an operator to add:

```toml
[[frame.component]]
use  = "repos"
size = 15
```

```
FAIL: test_a_small_plane_gets_exactly_the_rows_its_content_wants   (test_frame_layout.ReposIsSizedToItsContent)
FAIL: test_it_never_goes_below_one_row (content=0) and (content=1) (test_frame_layout.ReposIsSizedToItsContent)
FAIL: test_a_size_map_is_what_panel_argvs_splits_with              (test_frame_layout.SlotSizesAnswersEverySlotAtOnce)
FAIL: test_left_is_not_a_slot_this_module_will_size_or_split       (test_frame_layout.PanelGeometry)
FAIL: test_the_frame_it_splits_is_byte_for_byte_the_frame_it_split (test_component_id_is_the_currency.CharterOwnConfigIsUnchanged)
```

Six, deterministic, and the mechanism confirmed at the leaf:

```
layout.repos_rows(content_rows=4, window_rows=50, slots=["top","bottom","repos"])  ->  15
```

## The measurement I was asked to take, and what it changed

The recommendation was option 1 — thread the pin from `slot_sizes` into `repos_rows`,
leaving `slot_sizes` to do the read. **Measured, that fixes three of the six.** Three of
the red tests are `slot_sizes`' own:

```python
layout.slot_sizes(["repos"], window_rows=50, content_rows=7)          # expects repos: 7
layout.slot_sizes(["top","left","bottom","repos"], ..., content_rows=3)  # expects repos: 3
layout.slot_sizes(f["slots"], window_rows=50, content_rows=6)         # expects repos: 6
```

They pass explicit arguments and assert the content, exactly as `repos_rows`' do. A
`slot_sizes` that still reads `config.FRAME` answers 15 to all three. So the read cannot
stop at `slot_sizes` — it has to leave `layout`.

**And the "five signatures" does not survive being counted either.** The path is:

| hop | what it needs |
|---|---|
| `layout.repos_rows` | the pin — one new keyword |
| `layout.slot_sizes` | carries it — one new keyword |
| `commands_frame._launch_sizes` | supplies it |
| `commands_frame._relayout` | supplies it |
| `commands_frame._reassert_sizes` | supplies it |

**Two signatures and three call sites.** `panel_argvs`, `repos_cols` and `harness_rows` —
three of the five `_placed_here`'s docstring names — never need it: two take a `sizes` map
that is already computed and one is about columns. `_placed_here`'s five is real *for
`_placed_here`*, which answers "where does this placed component sit and how many cells
does it cost" for a name none of those functions knows in advance. A pinned strip is one
number, for one slot, with one consumer. The count was borrowed, not measured.

The three call sites were already building `content_rows=frame_slots.repos_rows_wanted(fid,
pane_cols=…)` the same way three times over, so they collapse into one boundary function
and the change **removes** two copies of that while adding one read:

```python
def _slot_sizes(fid, slots, *, window_rows, pane_cols):
    return layout.slot_sizes(
        slots, window_rows=window_rows,
        content_rows=frame_slots.repos_rows_wanted(fid, pane_cols=pane_cols),
        pinned_rows=layout.pinned_repo_rows())
```

So the answer to "is threading worse than reading?" here is no, and it is not close: one
read, one place, and `layout` is arithmetic again.

**Option 3 was considered and is refused for #661's own reason** — but it is worth naming
what it would have cost, because one of the six red tests looks like it *wants* the
config. `CharterOwnConfigIsUnchanged.test_the_frame_it_splits_is_byte_for_byte_the_frame_it_split`
reads `charter.toml` off disk deliberately and asserts the whole launch argv. It is red
under #660 because it asserts `repos: 6` for `content_rows=6`. Changing it to expect 15
would have made this repository's committed frame a term in charter's geometry tests —
the defect, restated as the fix.

## What shipped

* `layout.pinned_repo_rows` (was `_pinned_rows`) — the read, made public **so that its
  callers can be counted**. One caller.
* `layout.repos_rows(..., pinned_rows=None)` and `layout.slot_sizes(..., pinned_rows=None)`
  — every term is an argument again. `None` is "this plane pinned nothing", which is every
  plane that does not say otherwise and every test asking about the arithmetic.
* `commands_frame._slot_sizes` — the boundary. `_launch_sizes` (both launch paths),
  `_relayout`'s splits and `_reassert_sizes` on every `window-resized` all go through it,
  so the three cannot disagree about what the plane committed.

`docs/frame.md` is **unedited**: nothing an operator sees changes, and the page already
describes the behaviour the code now has.

## The test that would have caught it

`TheCommittedFileIsReadAtABoundaryAndNotInTheArithmetic`, in the module #660 added:

* `test_this_repositorys_own_committed_file_with_a_pin_in_it_is_not_read_here` — reads
  **this repository's real `charter.toml`**, adds `size = 15` to its `repos` table,
  resolves it, and asserts `repos_rows(content_rows=4, …) == 4` and the whole `slot_sizes`
  map. #661's reproduction, as a test. It asserts the pin *survived the boundary* first,
  so a green cannot mean "the arrangement stopped resolving `size` at all". It adds the
  key to what is on disk rather than replacing the file, so the day charter's own frame
  gains a pin it stays green instead of needing a rewrite.
* two cases sweeping every arrangement a plane can commit — none, written out with no
  `size`, and pins at 2, 15 and 99 — against both `repos_rows` and `slot_sizes`.
  **Asserted as the content the caller named, not as "the same for all five"**: five
  identical wrong answers is exactly what a function reading the file would give.
* `test_the_number_the_arithmetic_was_handed_is_the_number_it_used` — the other half, and
  what stops the above being satisfied by a `repos_rows` that ignores pins altogether:
  `pinned_rows=7` passed explicitly while `config.FRAME` holds a `size = 15`. An argument
  quietly dropped in favour of a re-read answers 15; a pin never read at all answers 4.

Confirmed RED: reinstating the read inside `repos_rows` turns 7 of these subtests red,
including the reproduction. Restored, 24/24 green.

`TheBoundaryIsWhereThePlaneIsRead` pins the boundary itself — the pinned path, the default
path (so a `_slot_sizes` that passed `pinned_rows` and dropped `content_rows` cannot pass),
and `_reassert_sizes` on the resize path with no tmux, reading the strip back the way
`test_frame_density` does: the window's rows minus every asserted height minus one border
per split. `test_frame_tmux_integration.test_a_pinned_table_really_lands_on_its_committed_height`
asks real tmux the same question and is unchanged and green.

#660's existing cases now take the pin through `_pin(**repos)`, which calls
`layout.pinned_repo_rows` under a patched `config.FRAME` — so the number the arithmetic is
handed is still the boundary's answer and never a `15` written twice, and `config.FRAME` is
patched in exactly one place in the module.

## Verification

* **Baseline**, `origin/main` at `9e6df55`, unpinned: **8590 tests, OK**.
* **Acceptance**, `size = 15` committed to `charter.toml`: **8597 tests, OK** — see the
  comment below for the run, since a linked worktree resolves its plane to the main tree
  and the main tree is not mine to edit.
* `test_frame_tmux_integration` against **real tmux 3.7c**: 67 tests, OK.
* Deletion sweep: pushed for the sharded gate rather than run locally.

All measured with `env -u PYTHONSAFEPATH` — `PYTHONSAFEPATH=1` makes 31 subprocess tests
fail and reads as a red tree before any mutation. Unchanged from #660's note.

**`size = 15` is not committed to `charter.toml` in this PR.** That is the operator's call
now that it works.

No version bump, no stamp, no tag.

https://claude.ai/code/session_01KGTJ3Zg7dgCQEeQLXWWo1d

### DIFF
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -1568,6 +1568,46 @@ def _spawn_gather(fid: str, ws: str) -> None:
         return
 
 
+def _slot_sizes(fid: str, slots: list[str], *,
+                window_rows: int, pane_cols: int) -> dict[str, int]:
+    """`layout.slot_sizes` for a frame that has a plane behind it — the only place in
+    charter where a committed arrangement becomes a pane height.
+
+    **This function is the boundary, and it is one function because the boundary is one
+    line.** `layout` is arithmetic: hand it a row count and a slot list and it answers the
+    same thing on every machine, which is what makes `layout.repos_rows` testable with no
+    tmux, no cache and no plane. The two facts it cannot derive are what the table's
+    content wants (`frame_slots.repos_rows_wanted`, which reads this frame's gather) and
+    whether the operator pinned the strip to a height (`layout.pinned_repo_rows`, which
+    reads this plane's `charter.toml`). Both are read HERE, once, and passed down.
+
+    #660 put the second read inside `layout.repos_rows` instead, borrowing
+    `layout._placed_here`'s "rather than threaded through five signatures". #661 is the
+    bill: `repos_rows(content_rows=4, window_rows=50, slots=["top","bottom","repos"])`
+    answered `15` on a plane whose `charter.toml` carried `size = 15`, from a file the
+    caller never named, in the module's one provably pure function. On this repository
+    `charter.toml` is tracked, so committing the line the feature's own news entry tells
+    an operator to write turned six tests red for everybody.
+
+    And the five was not this path's number. `_placed_here`'s five is real for
+    `_placed_here` — an edge and a cell count for a name `slot_sizes`, `panel_argvs`,
+    `repos_cols`, `repos_rows` and `harness_rows` each have to ask about, none of which
+    knows it in advance. The pin is one number for one slot with one consumer, so its
+    path is two signatures (`slot_sizes`, `repos_rows`) and the three call sites below —
+    which were already building `content_rows` the same way three times over, and now do
+    not.
+
+    The three are `_launch_sizes` (both launch paths), `_relayout`'s splits for a density
+    the frame did not have, and `_reassert_sizes` on every `window-resized`. They differ
+    only in how the table pane's width is arrived at, which is why that arrives here as
+    *pane_cols* already measured rather than as a window to measure.
+    """
+    return layout.slot_sizes(
+        slots, window_rows=window_rows,
+        content_rows=frame_slots.repos_rows_wanted(fid, pane_cols=pane_cols),
+        pinned_rows=layout.pinned_repo_rows())
+
+
 def _launch_sizes(fid: str, slots: list[str], *,
                   window_cols: int, window_rows: int) -> dict[str, int]:
     """How big each of *slots* is split, in a *window_cols* x *window_rows* window —
@@ -1604,11 +1644,12 @@ def _launch_sizes(fid: str, slots: list[str], *,
     `gather.discard(fid)` before it draws anything, so this reaches `gather.row_count`
     with no cache and gets the directory-listing answer — an `iterdir`, never the git
     sweep `gather.scan` would run. See `gather.row_count`'s own docstring for both paths.
+
+    Through :func:`_slot_sizes`, which is where the plane's own pinned height is read and
+    why this is not a call to `layout.slot_sizes` directly.
     """
-    return layout.slot_sizes(
-        slots, window_rows=window_rows,
-        content_rows=frame_slots.repos_rows_wanted(
-            fid, pane_cols=layout.repos_cols(slots, window_cols=window_cols)))
+    return _slot_sizes(fid, slots, window_rows=window_rows,
+                       pane_cols=layout.repos_cols(slots, window_cols=window_cols))
 
 
 def _drawable_slots(cols: int, rows: int, configured: list[str] | None = None) -> list[str]:
@@ -4034,11 +4075,10 @@ def _relayout(socket: str, *, fid: str, harness_pane: str, panels: dict[str, str
         keep.update(_split_panels(
             socket, slots=missing, fid=fid, harness_pane=harness_pane, env=None,
             pane_env=pane_env, v=v,
-            sizes=layout.slot_sizes(
-                want, window_rows=window_rows,
-                content_rows=frame_slots.repos_rows_wanted(
-                    fid, pane_cols=layout.repos_cols(list(keep) + missing,
-                                                      window_cols=window_cols)))))
+            sizes=_slot_sizes(
+                fid, want, window_rows=window_rows,
+                pane_cols=layout.repos_cols(list(keep) + missing,
+                                            window_cols=window_cols))))
         _arm_panel_respawn(socket, fid=fid,
                            panes={s: keep[s] for s in missing if s in keep}, env=None)
 
@@ -4238,11 +4278,9 @@ def _reassert_sizes(socket: str, *, fid: str, panes: dict[str, str], harness_pan
     # here, and it has to land before anything is measured — see :func:`_variable_pane_cols`
     # for the tmux measurement that makes this an order rather than a preference.
     _apply_sizes(socket, panes=panes, sizes=layout.column_sizes(order), flag="-x")
-    sizes = layout.slot_sizes(
-        order, window_rows=window_rows,
-        content_rows=frame_slots.repos_rows_wanted(
-            fid, pane_cols=_variable_pane_cols(socket, panes=panes,
-                                               window_cols=window_cols)))
+    sizes = _slot_sizes(
+        fid, order, window_rows=window_rows,
+        pane_cols=_variable_pane_cols(socket, panes=panes, window_cols=window_cols))
     # Pass two: the ROWS, and the harness below them.
     _apply_sizes(socket, panes=panes, sizes=sizes, flag="-y")
     if _PANE_ID_RE.fullmatch(harness_pane or ""):
--- a/charter/frame/layout.py
+++ b/charter/frame/layout.py
@@ -264,10 +264,10 @@ def _arrangement():
 
     **One line, because there are two callers and the fallback in it is unpinned.**
     :func:`_placed_here` reads the arrangement for a component charter did not write and
-    :func:`_pinned_rows` reads it for the one built-in whose height a plane may commit; the
-    `or ()` is the same defence for both, and a second copy of it is a second line nothing
-    can fail without. `tools/sweep.py` reported exactly that copy the day it appeared and
-    `docs/news/unreleased-the-deletion-sweep-is-a-thing-the-repo-runs.md` records the
+    :func:`pinned_repo_rows` reads it for the one built-in whose height a plane may
+    commit; the `or ()` is the same defence for both, and a second copy of it is a second
+    line nothing can fail without. `tools/sweep.py` reported that copy the day it appeared
+    and `docs/news/unreleased-the-deletion-sweep-is-a-thing-the-repo-runs.md` records the
     original at this module's line 289 as a survivor it examined — so the answer to a
     second one is to have one, not to accept two.
 
@@ -275,7 +275,7 @@ def _arrangement():
     ``components`` before its own early return, deliberately ("a key present on one path
     and absent on another is two shapes for one answer"), so nothing in production reaches
     here without it — but `config.FRAME` is module state that a caller may replace whole,
-    and a `KeyError` out of :func:`repos_rows` costs a launch its entire frame. Pinning it
+    and a `KeyError` out of either reader costs a launch its entire frame. Pinning it
     would mean a test constructing a `config.FRAME` that `frame_of` cannot produce, which
     is a test of the line rather than of any property.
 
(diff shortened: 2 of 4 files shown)


diazoxide:
## The acceptance run, and how it was got

**`size = 15` written into `charter.toml`, full suite: green.** Measured twice, two ways,
because a linked worktree does not resolve its own plane.

### Why this needed setting up rather than just editing the file

`charter/root.py` redirects a linked worktree to the tree it was cut from
(`_plane_of` → `_outermost`). So a worktree under the scratchpad reads the **main
checkout's** `charter.toml`, not its own:

```
$ cd <worktree> && python3 -c "from charter import config; print(config.ROOT)"
/Users/aharon/IdeaProjects/charter
```

Editing the operator's checked-out tree to run a test is not mine to do, so the pin went
somewhere the walk would actually find it.

### Run 1 — `$CHARTER_ROOT` pointed at the worktree, `size = 15` in its `charter.toml`

```
Ran 8597 tests in 389.139s
FAILED (failures=2)
  test_plane_spawn_guard.WhatIsGuarded.test_the_guarded_root_is_this_machines_own_plane
  test_plane_write_guard.WhatIsGuarded.test_the_guarded_directory_is_this_machines_own_plane_state
```

Both are `$CHARTER_ROOT` artifacts and neither is about the pin — they assert the guarded
root is *the plane this process resolved*, and pointing `$CHARTER_ROOT` elsewhere is
exactly the disagreement they exist to catch. Proven independent, same tree, same commit:

```
with    CHARTER_ROOT:  Ran 106 tests  FAILED (failures=2)
without CHARTER_ROOT:  Ran 106 tests  OK
```

Before this change, the same setup gave **five** failures in `test_frame_layout` plus
`CharterOwnConfigIsUnchanged` — #661's six.

### Run 2 — two standalone planes, differing only in the pin

To take `$CHARTER_ROOT` out of it entirely, the tree was copied to two plain directories
(no `.git`, so no worktree redirect and the walk finds the copy's own `charter.toml`), one
with the pin and one without, and both full suites run:

| | pinned (`size = 15`) | control (no pin) |
|---|---|---|
| tests | 8597 | 8597 |
| result | 12 errors | 12 errors |

**The same twelve, name for name** — `test_workflows`, `test_plugin_freshness` and
`test_doctor_shadowed`, every one of which reads `git ls-files` and finds no repository,
because the copies deliberately have no `.git`. The delta between a plane that pins the
strip and one that does not is **zero tests**.

### Baselines

* `origin/main` at `9e6df55`, unpinned, in the worktree: **8590 tests, OK**.
* This branch, unpinned, in the worktree: **8597 tests, OK** (the +7 are this PR's).

All with `env -u PYTHONSAFEPATH`.

### The new tests were confirmed RED

Reinstating the read inside `repos_rows` (`pinned_rows = pinned_repo_rows() if
pinned_rows is None else pinned_rows`) and re-running the module:

```
Ran 24 tests
FAILED (failures=7)
  test_no_arrangement_a_plane_can_commit_changes_what_repos_rows_answers  (pinned small / large / past the cap)
  test_no_arrangement_a_plane_can_commit_changes_what_slot_sizes_answers  (pinned small / large / past the cap)
  test_this_repositorys_own_committed_file_with_a_pin_in_it_is_not_read_here
```

Restored: 24/24 green. `test_frame_tmux_integration` against real tmux 3.7c — including
#660's `test_a_pinned_table_really_lands_on_its_committed_height`, unchanged — 67 tests,
OK.

diazoxide:
## Deletion sweep — clean, decomposed, plus the two the operators do not reach

CI, [run 33265605495](https://github.com/diazoxide/charter/actions/runs/33265605495), on
linux / CPython 3.12.14:

```
baseline          : Ran 8597 tests — OK
mutations applied : 2
pinned            : 2
SURVIVED          : 0
UNRESOLVED        : 0

gate: 0 unpinned, 0 in masked cluster(s), 0 platform-deferred, 0 unresolved, 0 not applied, 2 pinned
gate: no survivors
```

**Decomposed: 0 unpinned, 0 masked-cluster, 0 platform-deferred, 0 unresolved.** Both
mutations are `charter/frame/layout.py:596 [collapse-ifexp]` — the ternary collapsed each
way — and both go red.

### Two is a small number, so here is what it does not cover

The diff is 110 added lines across two files and most of them are docstring. The two
lines that carry behaviour and are **not** shapes `OPERATORS` mutates are keyword
arguments, so the sweep never offered them. Applied by hand and run against a covering
subset (`test_a_pinned_repo_strip_keeps_its_height`, `test_frame_layout`,
`test_frame_density`, `test_component_id_is_the_currency` — 273 tests):

| mutation | result |
|---|---|
| `commands_frame._slot_sizes` drops `pinned_rows=layout.pinned_repo_rows()` — the read itself | **3 red**: the boundary, the launcher, and the resize hook |
| `layout.slot_sizes` stops passing `pinned_rows=pinned_rows` down to `repos_rows` | **4 red**: those three plus `test_the_pin_reaches_tmux_as_the_length_the_pane_is_split_to` |

So every line this branch adds that does anything is pinned, including the two the tool's
operators cannot express. That the resize hook is red in both is the point of adding
`test_the_resize_hook_sizes_the_strip_from_the_pin_and_not_the_clone_count`: before it,
the `window-resized` path was pinned only by the real-tmux case, which is platform-deferred
on a box without tmux.

### And the mutation that matters most was already confirmed red

Reinstating #660's read inside `repos_rows` turns 7 subtests red, including #661's own
reproduction against this repository's real `charter.toml`. That one is in the comment
above.

PR #664, opened 2026-08-29
The repo table's viewport moves over rows, so a plane with one clone and many worktrees can scroll it (#663)

### LINKED ISSUE #663: The repo table's viewport moves over repos, so a plane with one repo and many worktrees cannot scroll at all
#658 shipped scroll on the `repos` pane. **On this plane it is a complete no-op**, and the shape it fails on is this plane's ordinary state: one clone with fourteen worktrees.

## Measured

```
_scroll_limit(repos=1,  budget=6 ) = 0
_scroll_limit(repos=1,  budget=17) = 0
_scroll_limit(repos=1,  budget=3 ) = 0
_scroll_limit(repos=6,  budget=4 ) = 3     <- many repos: works
_scroll_limit(repos=14, budget=6 ) = 9
```

Live, on the frame: `repos` pane resized to 7 rows against 15 rows of content, three SGR wheel-downs injected into the pane's own pty (`\033[&lt;65;10;5M`) — **nothing moved**, and the seven worktrees past the pane's height are simply dropped. Clicks confirm the other half: `VIEWPORT.repo_at()` answers `'charter'` for the repo row and `None` for **every** worktree row, so a click below the first line selects nothing.

## This was stated, not hidden — and that is the point

`_scroll_limit`'s own docstring says it plainly:

> **The window moves over REPOS and not over PIECES, and that is a stated limit rather than an oversight.** Piece rows appear only in the one-repo shape […] where one repo always fits and this answers 0; pieces past the pane's remaining rows are dropped by that function as they always were.

The limit was declared honestly in #658's report. What nobody checked is that **the plane it was built on is the one-repo shape**, so the feature shipped switched off for its own operator. A stated limit that happens to cover the only configuration in front of you is not a limit, it is a no-op with documentation.

## Why the arithmetic is wrong in kind

`_scroll_limit` counts **repos**; the pane renders **rows**. Those are the same number only when no repo has pieces. The budget is in rows, the overflow reservation is in rows, and `repo_at` maps a row — so repos is the one unit in the expression that is not the pane's.

`_table_lines` already produces the rendered line list, and `VIEWPORT.publish` already receives one entry per drawn row. The quantity the viewport needs is `len(all_lines)`, not `len(repos)`.

## What a fix has to keep

- **Offset zero must stay byte-for-byte the current table.** #658's ranking (`_pick_rows`) is what decides which repos survive a starved pane; scrolling must move a window over that ranking, not replace it.
- **A pane tall enough for its content must still scroll nothing**, and that must stay a *tested* nothing — asserted at the boundary and one either side, per #658.
- **`repo_at` must resolve a piece row to something.** Today it answers `None` below the first repo. Either a piece row selects its parent repo (#658's stated behaviour, which is defensible) or pieces become selectable in their own right — but a row that is drawn and answers `None` is a dead row, and there would now be fourteen of them.
- A stale offset over a list that shrank must still be clamped by the next paint.

## Related

- **#658** — shipped the scroll, and stated this limit.
- **#660 / #661 / #662** — the pinned height, which makes overflow *more* likely rather than less: a plane that pins `size = 15` and has twenty worktrees now overflows permanently.

diazoxide:
Closes #663.

`slots._scroll_limit` counted **repos**; the pane renders **rows**. Those are the same
number only where no repo has pieces — so on the one-clone-many-worktrees plane, which is
a control plane's ordinary state and is the state of the plane #658 shipped from, it
answered `0` at every pane height. The wheel was inert, and the worktrees past the pane's
last row were dropped with nothing on screen saying so.

## Re-measured before anything was written

Reported numbers, confirmed on `d6318be`:

```
_scroll_limit(repos=1,  budget=6 ) = 0     _scroll_limit(repos=1, budget=17) = 0
_scroll_limit(repos=1,  budget=3 ) = 0     _scroll_limit(repos=6, budget=4 ) = 3
```

And on the real cache this plane carries (1 clone, 13 worktrees) rendered into a 7-row
pane: the offset changes nothing at any value, five worktrees drawn, eight dropped, no
`…(+N more)` at all, and no `⑂13` badge either — `shown_pieces` counted every piece in the
*cache* rather than the ones on *screen*, so the badge that means "there is more you cannot
see here" disappeared exactly when it was true.

**The pane is capped at `statusline._MAX_REPO_LINES` (14) however tall it is**, so this is
not only a starved-pane case: a plane with fourteen worktrees overflows permanently. This
plane has thirteen. It is one `worktree add` from the defect.

## The docstring's argument, taken seriously and then overturned

`_scroll_limit` gave two reasons, and both were true as stated:

> Scrolling them would need a second kind of offset — an index into a nested list rather
> than into the ranking — and one mechanism that answers for half the cases honestly is
> worth more than two that overlap.

**It is not a second kind of offset, and that is what makes the limit wrong rather than
conservative.** The table already spends its budget in one order — repo rows first, then
the piece rows nested under them (`statusline._repo_rows`' `wt_budget`). One offset into
*that* list is the same ordering read one window further along:

* while any repo row is in the window it is `_pick_rows(keys, room, …, offset=offset)` —
  literally the call that was already there;
* once every repo row is above the window, `offset - len(keys)` indexes the pieces.

`_pick_rows` answering *fewer rows, and eventually none* past the end — already its
documented behaviour, already deliberate, already refusing to clamp — is exactly what makes
the handover free. One mechanism, one offset, no nested index.

The second reason ("one repo always fits") was arithmetic about repos, and it is true;
what it does not survive is that the pane does not draw repos.

## What is in it

- **`_content_rows(data)`** — repo rows plus `_piece_rows(data)`, one function that both
  the composer and the bound are asked. A count above what is composed leaves an offset
  that scrolls onto nothing; a count below it leaves the bottom of the plane out of reach,
  which is the half that shipped. `test_the_counter_and_the_composer_agree_on_every_shape`
  is that as an equality over six shapes.
- **`_scroll_limit(rows, budget)`** — same three lines, a different argument.
- **`_table_lines`** windows over rows: `room = budget - (1 if capped else 0)`,
  `show = _pick_rows(keys, room, …, offset)`, `kids = pieces[start : start + room - len(show)]`
  with `start = max(0, offset - len(keys))`.
- **The `…(+N more)` row is reserved on the same condition every other shape reserves it
  on**, is composed LAST (it used to be appended above the piece rows, which could not be
  told apart while `capped` and a piece row were mutually exclusive — they no longer are),
  and counts hidden **rows**: hidden repos *and* hidden pieces, so `, all clean` is not
  said over a dirty worktree that is off screen.
- **`╰─` only when nothing follows.** A capped table's last drawn piece row keeps `├─`; the
  glyph and the admission under it were saying opposite things.
- **`⑂N` counts the pieces DRAWN.**

## Seen moving on a real frame

Real tmux, real attached client on a real pty, real `charter panel repos` and `charter
panel bottom` processes, `[frame] mouse = true` from charter's own `conf_text`, harness
active so the reports arrive because of charter's flag and not because the pane is
selected. SGR wheel reports written to the client's terminal:

```
--- offset 0 (as the pane paints it)
      ├─ solo ⑂9                           main
      │  ├─ piece0                         main
      │  ├─ piece1                         main
      …(+7 more, all clean)
--- after seven wheel-downs (the limit is 7)
      │  ├─ piece6                         main
      │  ├─ piece7                         main
      │  ├─ piece8                         main
      …(+7 more, all clean)
--- an eighth wheel-down: at the bottom, nothing moves
      │  ├─ piece6                         main
      │  ├─ piece7                         main
      │  ├─ piece8                         main
      …(+7 more, all clean)
--- seven wheel-ups: back to the table it started on
      ├─ solo ⑂9                           main
      │  ├─ piece0                         main
      │  ├─ piece1                         main
      …(+7 more, all clean)
```

`tests/test_a_real_click_reaches_the_real_repo_table.py` was the harness that could not
have caught this: it was built entirely on a many-clones plane, where a repo-counted limit
is right by coincidence. Its setup is now a base class and the **cache** is what a subclass
supplies, so "the other plane shape" is a two-line subclass rather than a reason not to
test it. Three shapes now run through it; **15 cases, all green**, and the five new
one-clone cases were **red before the change**:

| case | before | after |
|---|---|---|
| a wheel notch moves the window over the worktrees | **FAIL** — nothing moved | ok |
| the last worktree is reachable | **FAIL** | ok |
| the pieces that do not fit are admitted rather than dropped | **FAIL** — no note at all | ok |
| a click on a worktree row selects its clone (table **scrolled**, so the clone has no row) | **FAIL** | ok |
| the wheel comes back to the table it started on | **FAIL** | ok |

## The four constraints

1. **Offset zero is byte-for-byte the current table** wherever the ranking is what decides.
   Measured, not asserted: `_table_lines` output diffed old-vs-new over a grid of 320
   shapes (0/1/2/3/5/6/14/20 repos × 0/1/3/9/14 worktrees × 8 budgets). **Every shape is
   identical except those satisfying `1 + pieces > budget`** — the overflowing one-clone
   table, which is the defect. The plane this ships from, rendered from its own live
   `gather.json` into its own 15-row pane, is byte-identical and still `limit=0`.

   The one-clone overflow shape changing at offset zero **is the fix**, and it is the
   change this module's own rules demand: a table drawing 6 of 15 rows and admitting
   nothing is the false-clean reading it refuses everywhere else. Stating it plainly rather
   than hiding it under "compatible".
2. **A pane tall enough still scrolls nothing, and it is a tested nothing** — at the
   boundary and one either side in the row unit (`_scroll_limit(15, 15) == 0`,
   `(14, 15) == 0`, `(15, 6) == 10`, `(15, 3) == 13`), at the paint level (one clone, nine
   worktrees, an 11-row pane → `VIEWPORT.limit == 0` with the ninth worktree drawn), and
   **on a real frame**: `ARealWheelOverAPaneThatFitsItsPieces` lands a click and waits for
   its highlight first — that is the panel proving it repaints on this pty — and only then
   asserts three wheel notches cost nothing. Verified meaningful by mutating
   `rows <= budget` to `rows < budget`: that case goes **red**, on the real pane.
3. **No drawn row answers `None`** — asserted at four offsets, including offsets where the
   clone has no row of its own on screen and a piece row is the only thing that can answer.
4. **A stale offset is clamped by the next paint** — `settle` unchanged, and the bound now
   moves when a **worktree** is removed as well as when a repo is
   (`test_a_worktree_removed_under_a_scrolled_table_clamps_it_on_the_next_paint`: scrolled
   to 7, four worktrees removed, next paint puts it at 3).

## One claim in the report that did not reproduce

> `VIEWPORT.repo_at()` answers `'charter'` for the repo row and **`None` for every worktree
> row**

**It does not, and did not.** On `d6318be`, painting one clone with nine worktrees into a
7-row pane:

```
BASE   rows drawn: 7   repo_at: [None, 'solo', 'solo', 'solo', 'solo', 'solo', 'solo', None]
```

`gather.scan` has written `entry["repo"] = repo_for_wt` on every worktree row since #388,
`_Line` carries it, and a click on a piece row already selected its parent clone. The two
ways to see `None` there are a cache written before #388, or a `VIEWPORT` that never
painted the table at that height. **What was true is the half underneath it**: nine of the
fourteen rows did not exist to be clicked. Constraint 3 is met, and was met — it is now
asserted at offsets where a piece row is the only row that can answer, which is where it
was never tested.

## Deletion sweep

CI's own gate, sharded, on `a9b9efe`:

```
merged 23 result(s) from 1 of 1 shard(s)
gate: 2 unpinned, 2 in masked cluster(s), 0 platform-deferred, 0 unresolved, 0 not applied, 19 pinned
gate: 4 survivors
```

**None is suppressed and none is argued away — all four lines are gone.** Decomposed:

| where | verdict | what was done |
|---|---|---|
| `slots.py:585` `_piece_rows` `[no-fallback] data.get("repos") or []` | **masked cluster** (2 survivors in one function) | **Deleted** → `data["repos"]` |
| `slots.py:585` `_piece_rows` `[no-fallback] data.get("worktrees") or []` | **masked cluster** (same pair) | **Deleted** → `data["worktrees"]` |
| `slots.py:600` `_content_rows` `[no-fallback] data.get("repos") or []` | unpinned | **Deleted** → `data["repos"]` |
| `slots.py:815` `_table_lines` `[collapse-ifexp] … if capped else keys` | unpinned | **Deleted** — the branch is gone |

The first three are one finding, and it is one this file already answers ten screens down
in `_selected_detail`:

> `data["repos"]`, with no `or []` under it: `gather.cached` answers `None` for anything
> whose `repos` is not a LIST (`gather._shaped_like_a_scan`), so a fallback here could only
> ever stand in for an empty list with an empty list.

`_shaped_like_a_scan` checks `worktrees` on the same line it checks `repos`, so the
argument covers both keys and both new functions. Writing a test to redden them would have
been pinning a line that cannot change an outcome.

The fourth: the sweep is right that the arms are indistinguishable, and the reason is
worth stating because it is what makes the whole fix one mechanism. An **uncapped** table
has `room >= len(keys)` by definition and its offset is 0 (`_scroll_limit` answers 0 and
`_Viewport.settle` holds it there), and `_pick_rows` re-sorts its pick back into cache
order — so `[0:room]` of the ranking **is** `keys`, element for element. The branch was an
optimisation over two sorts of at most fourteen rows on a pane that repaints on a version
bump, and what it cost was a second shape for the line to have: with it, an offset over a
table that fits moved the pieces and not the repos. One window over one list, at every
offset, capped or not.

Rendering re-diffed after the four deletions: **byte-identical** to the pre-deletion tree
over the same 320-shape grid.

## Verification

- Full suite green on the final tree.
- `tests/test_a_real_click_reaches_the_real_repo_table.py` — 15 cases over three plane
  shapes, real tmux throughout.
- Sweep verdict above; the re-run on the final commit is in a comment below.
- No version bump, no stamp, no tag. News entry under `docs/news/` at `version: unreleased`.
- `docs/frame.md` gains the operator-facing half: rows, not clones, and what a scrolled
  tree looks like.

https://claude.ai/code/session_015BwnjrnCu9t4pMooDm7r6H

### DIFF
--- a/charter/frame/slots.py
+++ b/charter/frame/slots.py
@@ -457,13 +457,17 @@ class _Viewport:
     puts it back — the reset is the price of the singleton and is not hidden.
 
     **It holds an intent, and the renderer is what settles it against the data.** The
-    handler moves an offset knowing nothing about how many repos there are or how tall the
-    pane is — `Component.on_event` is handed no ctx by contract (§4f), and it must not grow
-    a second reading of a plane the repaint is about to read anyway. So :meth:`settle` is
-    called by `_repos` with the bound it just computed, and that is also the answer to *what
-    happens when the offset outlives a repo list that shrank under it*: the next paint
+    handler moves an offset knowing nothing about how many rows the table has or how tall
+    the pane is — `Component.on_event` is handed no ctx by contract (§4f), and it must not
+    grow a second reading of a plane the repaint is about to read anyway. So :meth:`settle`
+    is called by `_repos` with the bound it just computed, and that is also the answer to
+    *what happens when the offset outlives a table that shrank under it*: the next paint
     clamps it down and the pane draws the last window there is, rather than an empty table
     the operator has to guess their way back out of.
+
+    **The offset counts pane ROWS, not repos** (#663). A table is repo rows and the piece
+    rows nested under them, and a viewport whose unit was the repo could not move at all on
+    a plane with one clone — which was every control plane charter runs on.
 
     :attr:`limit` is that bound REMEMBERED, which is what lets the handler refuse a scroll
     that would change nothing. Zero until the first paint, and the first paint always
@@ -495,9 +499,10 @@ class _Viewport:
         """Record how far this table may scroll, and answer where it actually is.
 
         The clamp is here and nowhere else. `_scroll_limit` computes the bound from the
-        repo count and the pane's rows; this is what applies it, so an offset left over
-        from a longer list — the operator scrolled to the bottom, then a repo was removed —
-        comes back as the largest offset the CURRENT list has, on the very next paint.
+        table's ROW count (:func:`_content_rows`) and the pane's rows; this is what applies
+        it, so an offset left over from a longer table — the operator scrolled to the
+        bottom, then a repo or a worktree was removed — comes back as the largest offset
+        the CURRENT table has, on the very next paint.
 
         Never below zero, and that half is not symmetry: :meth:`move` already refuses to go
         under, so what this guards is a *limit* that has gone negative, which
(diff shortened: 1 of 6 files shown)


diazoxide:
### Deletion sweep, final — `4c8483d`

```
merged 18 result(s) from 1 of 1 shard(s)
gate: 0 unpinned, 0 in masked cluster(s), 0 platform-deferred, 0 unresolved, 0 not applied, 18 pinned
gate: no survivors
```

The first verdict, on `a9b9efe`, was **4 survivors**, decomposed in the PR body: two in one
masked cluster in `_piece_rows`, one unpinned in `_content_rows`, one unpinned in
`_table_lines`. **All four lines are deleted; none is suppressed and none is argued away.**
23 mutations became 18 because four lines are gone and the fifth site went with the
collapsed branch, and all 18 are pinned.

Also on this commit:

- full suite `Ran 8622 tests … OK` locally, and green on 3.11 / 3.12 / 3.13 / 3.14 in CI;
- `tests/test_a_real_click_reaches_the_real_repo_table.py` — 15 cases over three plane
  shapes on a real tmux;
- the `_table_lines` rendering re-diffed old-vs-new over the 320-shape grid after the
  deletions: unchanged from the pre-deletion tree, and still identical to `d6318be`
  everywhere except the overflowing one-clone shape.

PR #665, opened 2026-08-29
A release body GitHub will refuse is a release that publishes and then loses its notes

diazoxide:
Found while auditing release readiness for the next tag. **Do not merge without reading the second section** — this PR is red on purpose.

## What it adds

`tests/test_release_notes_fit_the_release.py`, three cases:

- `test_published_versions_fit` — **green.** Every stamped version's rendered body is under GitHub's limit. Permanent regression guard.
- `test_the_staged_release_fits` — **red.** The 69 entries staged for the next release render to **333,917 characters**; GitHub refuses a release body over **125,000**.
- `test_the_staged_release_has_headroom` — **red.** Under the limit is not the same as safe.

## Why this is a release-stopper

`release.yml`'s `announce` job pipes `charter news --for <version>` straight into `gh release create --notes-file`. GitHub's create-release API refuses a body over 125,000 characters (`body is too long`), and `gh` forwards the refusal rather than truncating.

`announce` is `needs: publish`, so the ordering is: guard → test → build → **irreversible PyPI upload** → Release creation fails.

The documented retry cannot repair it. `gh workflow run release.yml -f version=<X.Y.Z>` re-enters `publish`, which calls `pypa/gh-action-pypi-publish` with no `skip-existing`, so PyPI rejects a version it already has and `announce` is never reached again. The only remaining exit is the one the release charter forbids — writing the Release body by hand, which forks the published notes from the shipped entry that is meant to be their single source.

## Why the existing guard does not catch it

The `guard` job's reasoning is not at fault and is not weakened here. It refuses a version whose notes cannot be *rendered*, and asks `charter news --for` — the same call `announce` makes — precisely so that "the guard passed" and "the notes render" cannot disagree.

Rendering is not the failing step. That call exits 0 on a 300,000-character body, because producing the string is exactly what it was asked to do. The claim nobody was making is that the string **fits where it is about to be sent**.

## How close this already was

| version | entries | rendered body |
|---|---|---|
| 0.50.0 | 3 | 7,119 |
| 0.51.0 | 10 | 28,867 |
| 0.52.0 | 24 | **111,349 published — 3,651 short of the refusal** |
| 0.53.0 | 13 | 98,265 |
| staged | 69 | **333,917** |

Nothing noticed how close 0.52.0 came, because nothing was looking. Entries accumulate one PR at a time and each author sees only their own; the total crosses on an ordinary afternoon with no PR to blame. That is why the headroom case is separate rather than a stricter copy of the limit case.

## Measured in characters, not bytes

The API counts characters, and these entries carry `—`, `✗` and `⬢`. `len(body.encode())` would report a larger number than the one GitHub applies and would refuse a release it would in fact have accepted.

## What merging this does

It turns `main` red until the staged body is bounded. That is an accurate report of the tree's state, but it is a call for the release engineer and the operator, not for this PR — hence: opened, not merged. Two other agents have branches in flight.

The fix I would recommend lives in `news.render_body`: emit full bodies while they fit and headline-plus-link past the limit, so nothing is silently dropped and both views still come from one source — plus move this assertion into the `guard` job, so it fails at the cheap end instead of after the upload.

### DIFF
--- a/tests/test_release_notes_fit_the_release.py
+++ b/tests/test_release_notes_fit_the_release.py
@@ -1,0 +1,131 @@
+"""A release body GitHub will refuse is a release that publishes and then loses its notes.
+
+`release.yml`'s `announce` job ends with
+
+    python -m charter news --for "$version" > "$RUNNER_TEMP/notes.md"
+    gh release create "$tag" --title "$tag" --notes-file "$RUNNER_TEMP/notes.md"
+
+and GitHub's create-release API refuses a body over **125,000 characters** with
+``body is too long (maximum is 125000 characters)``. `gh` does not truncate to fit; it
+forwards the refusal. So a version whose entries render past that limit does not get
+shorter notes — it gets **no Release at all**.
+
+**Why that is a release-stopper and not a cosmetic bug.** `announce` is `needs: publish`,
+so the order is: guard, test, build, *upload to PyPI*, then create the Release. The
+upload is the irreversible step and it has already happened by the time GitHub refuses.
+Worse, the documented retry — `gh workflow run release.yml -f version=<X.Y.Z>` — cannot
+repair it: the retry re-enters `publish`, `pypa/gh-action-pypi-publish` is called without
+`skip-existing`, and PyPI rejects a version it already has, so `announce` is never reached
+on any subsequent run. The only way out is the one the release charter forbids — writing
+the Release body by hand, which forks the published notes from the shipped entry that is
+supposed to be their single source.
+
+**The existing guard cannot see this.** `guard` already refuses a version whose notes
+cannot be *rendered*, and does it by asking `charter news --for` — the same call
+`announce` makes — precisely so "the guard passed" and "the notes render" cannot disagree.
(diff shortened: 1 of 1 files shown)


diazoxide:
The fix for this is up as #672, and the three cases from this branch are in it — prose brought up to the shipped state, assertions kept, and still measured against the shipped tree rather than a fixture. They pass there because `news.render_body` is now bounded, which makes them the deletion-sweep pin on the bound rather than a report of a red tree.

Recommend closing this in favour of #672 once that lands, rather than merging it: merging this alone reddens `main`, which is what its own description says.

Follow-up filed separately as #673 — `publish` has no `skip-existing`, so *any* post-upload failure in `announce` is unrecoverable by the documented retry, not just the one fixed here.

diazoxide:
Superseded by #672, which lands the bound and the guard together.

This PR's three cases are in #672 with their prose updated and their assertions kept, still measured against the shipped tree. Merging this one alone would redden `main` — as its own description says, the failures are deliberate — so it goes as part of the fix rather than ahead of it.

Issue #502, opened 2026-08-24
news report lines contain the ordering VALUE but interpolate the committed FILENAME raw


diazoxide:
`news.ordering_errors` and `news.stamp` build a line of charter's own report out of a
committed filename, and interpolate it raw. `contain.one_line` is applied to the ordering
field's **value** and not to the **name of the file** the value came from, so an entry
whose filename holds a newline forges an extra line inside charter's own output.

A news entry is a committed file: its name is attacker-influenced in exactly the way
#453's MCP server name was, one surface over. Nothing about a `.md` filename is validated
before it reaches these messages.

## Reproduction

```python
import tempfile, pathlib
from unittest import mock
from charter import news

d = pathlib.Path(tempfile.mkdtemp())
name = "0.60.0-a\nEVIL: charter says nothing is wrong.md"
(d / name).write_text("---\nversion: 0.60.0\nheadline: h\nsecurity: yes\n---\n\nbody\n")
with mock.patch.object(news, "_PACKAGED", d):
    for why in news.ordering_errors(news.all()):
        print(why)
```

prints

```
0.60.0-a
EVIL: charter says nothing is wrong.md: `security: yes` is not a value charter reads — a news ordering field is `true` or `false`. Left unread, this entry sorts as though it never declared anything.
```

Two lines where charter emitted one, and the second is the attacker's sentence sitting in
CI output that a release engineer reads as charter's.

## The four sites

Two are new with #486 / PR #490:

* `charter/news.py:307` — `f"{e.path.name}: ..."` in the unreadable-value message
* `charter/news.py:316` — `", ".join(sorted(c.path.name for c in claimants))` in the
  duplicate-`lead:` message

Two are pre-existing and are why this is filed rather than folded into #490:

* `charter/news.py:899` — `f"{src.name} → {dst.name}: that name is already taken..."`
* `charter/news.py:904` — `f"{src.name}: no `version:` line in its frontmatter to stamp"`

`stamp()`'s two have been there since news entries existed, so a fix belongs in one change
that covers all four and can be reasoned about as a whole — the value guard was applied at
#490's two sites because it was in view there, not because the filename half was judged
safe.

## The property, and the next spelling

The property is **"every untrusted span in a report line is contained, not just the one
the author was thinking about"**. The value was contained because the value was the
subject of that commit; the filename carried the same trust level and was interpolated raw
three inches away.

The next spelling is a *third* untrusted span appearing in one of these messages — a
[...]

Issue #486, opened 2026-08-24
news: the Release body is ordered by slug, so a release cannot lead with its most important entry


diazoxide:
Found cutting 0.52.0.

`charter news --for <version>` renders entries in `sorted(d.glob("*.md"))` order (`charter/news.py:199`), which for a stamped release is alphabetical by slug. `release.yml`'s `announce` job pipes that straight into `gh release create --notes-file`, so slug order **is** the published order.

0.52.0 shipped 24 entries. The one that matters most — `a-server-name-cannot-declare-a-server`, a committed `mcp.json` server name interpolated raw into a generated sub-agent's YAML frontmatter, which could hand any vault on the machine to an attacker-chosen command (#453) — renders **eighth**, below `a-1password-vault-is-one-item`, a docs correction. A reader who opens https://github.com/diazoxide/charter/releases/tag/v0.52.0 and stops after two screens never learns they need to upgrade.

The ordering is not arbitrary today so much as accidental: it is whatever the filenames sort to, and the filename is a slug an author picked for a different reason.

**Why it was not fixed by hand at release time.** The release engineer's charter says the shipped entry is the single source for both the Release body and the offline `charter news` suggestion, and that hand-editing forks them — *"To change what a Release says, change the entry and the text follows."* That is right, and it means the fix belongs in the entry format or the renderer, not in a one-off edit to one Release.

## Shape of a fix

An explicit rank in the frontmatter, defaulting to "no opinion", so an entry can ask to lead and everything else keeps sorting as it does now — e.g. `lead: true`, or a numeric `order:`. Whatever the spelling, it should be:

- **opt-in**, so 24 entries do not each need a number;
- **checked**, so two entries cannot both claim the lead silently;
- **visible to `charter news`** as well as to the Release body, since they are deliberately the same call.

A `severity:`/`security: true` field that sorts security entries first would answer this specific case and also give `charter news` something to highlight offline, which slug order can never do.

Filed from the 0.52.0 release. Nothing about it blocked the publish — the notes are complete and correct, just ordered by an accident of naming.

PR #672, opened 2026-08-29
A release body GitHub will refuse is a release that publishes and then loses its notes (#665)

diazoxide:
Fixes the release blocker #665 found, at the end of the workflow where it can still be fixed.

## The blocker

`release.yml`'s `announce` job pipes `charter news --for <version>` into `gh release create --notes-file`, and GitHub's create-release API refuses a body over **125,000 characters** with `body is too long`. It does not trim, and `gh` forwards the refusal rather than truncating. So a version whose entries render past that limit did not publish shorter notes — it published **none**.

| | |
|---|---|
| raw size of the 69 unreleased entries | 339,334 bytes |
| rendered body for the next release | **333,917 characters** |
| GitHub create-release body limit | 125,000 characters |
| 0.52.0, published | **111,723 — 3,277 short of the refusal** |

`announce` waits on `publish`, so the refusal lands **after** the PyPI upload. And the documented retry cannot repair it: `gh workflow run release.yml -f version=<X.Y.Z>` re-enters `publish`, `pypa/gh-action-pypi-publish` is invoked with no `with:` block at all — so no `skip-existing` — and PyPI rejects a version it already has. `announce` is never reached again. Re-running the failed job alone cannot help either: the body is deterministically too big.

**The guard's reasoning was not at fault and is not weakened here.** It refuses a version whose notes cannot be *rendered*, and asks `charter news --for` — the same call `announce` makes — precisely so "the guard passed" and "the notes render" cannot disagree. Rendering was never the failing step: that call exits 0 on a 300,000-character body, because producing the string is what it was asked to do. The claim nobody was making is that the string **fits where it is sent**.

## What this does

**`news.render_body` is bounded** (`charter/news.py`). Whole while it fits `_BODY_BUDGET`; past that, the notes that fit render whole and the rest become their headline and a link to the note, with a heading between them stating how many, how long, and what the limit is. **Sixteen of the seventeen** stamped versions render byte-identically to before, measured rather than estimated — asserted as an equality, not as three `assertIn`s. 0.52.0 is the only one the bound reaches for: 111,723 → 95,288.

**Nothing is dropped and nothing is truncated.** Truncation is the dangerous shape here and it is the one thing this must not be: a body cut at 125,000 characters satisfies every length assertion and is indistinguishable, to a reader, from a release that shipped a dozen fewer things. So every entry keeps its own heading in its own place in the order; every elided entry carries a link to the note holding its text; and the elision is visible **at every point it applies**, not only at the cut — which is why one notice is enough.

**The cut is one point in the order, not a per-entry decision.** A greedy fill would give an ordinary note its body while a security note above it lost one — #486 wearing a size limit. And there is no second rule inside the bound that promotes security entries: there is one order, `news.all()`'s, and the notes that lead are the notes that keep their bodies. The consequence is asserted rather than assumed, in both its specific form (a security note is never demoted while an ordinary one keeps its body) and its general one (no note keeps its body below one that lost its own).

**The assertion moved to `guard`.** `charter news --for` refuses a body it cannot bound, so the failure arrives before `test`, `build` and `publish` instead of after the upload. It lives in the command rather than in a new workflow step because `guard` and `announce` run that exact command against that exact tree — a second copy of the limit in shell would be a second answer to one question, and it is the announce-side answer that decides whether the release exists.

## Two numbers, and why they are two

- `news.RELEASE_BODY_MAX = 125_000` — **GitHub's** number. Characters, not bytes: the API counts code points and charter's entries carry `—`, `✗`, `⬢`, so `len(body.encode())` reports 3,023 more than GitHub applies for the staged set and would refuse a release GitHub would have accepted.
- `news._BODY_BUDGET = 100_000` — **charter's**, deliberately below it. `charter news --for` `print`s the body, so the file `announce` redirects into is one character longer than what was measured; a ceiling reached is a ceiling with nothing left for the next change to the announce step, landing after the upload; and 0.52.0's 3,277-character margin is the evidence that "just under" is a state nobody notices. 100,000 is also above every release ever cut but two, so the bound does not begin eliding until a version is genuinely larger than any that has shipped.

The test states GitHub's 125,000 **independently** rather than importing charter's copy — a test that read charter's constant would agree with a typo in it — and asserts the two in the one direction with teeth: charter may be more conservative, never less.

## #665's tests, landed

All three of #665's cases are here and green, with their prose brought up to the shipped state. They are now the deletion-sweep pin on the bound, still measured against the **shipped tree** rather than a fixture — a fixture proves `render_body` can count, which nothing doubted. `test_the_staged_release_has_headroom` becomes the end-to-end guard on `_BODY_BUDGET`: raise it toward the ceiling and it goes red.

Added beside them: every staged note is present in the body it would publish; every elided note links to a path that exists in this checkout; the notice is written once and sits at the cut; a filename cannot forge a heading or break out of the link (#502, one document over); the link names the constant repository and not `CHARTER_UPSTREAM_REPO`; and a release whose notes are merely long still publishes — the refusal is for what cannot be bounded, not for what is big.

## Two couplings this had to honour

`tests/test_news_ordering.py` pins the number of ways `charter news --for` can refuse, because `release.yml`'s `::error::` enumerates them for an operator who has only the CI log. The count moves 2 → 3, the annotation gains the third cause, and the two texts are coupled on `too long` — the phrase the command prints and the phrase GitHub's own refusal uses.

And `announce` now writes `set -eu`. Actions runs `run:` under `bash -e`, so a refused render already aborted the step — but "the render refuses and the Release is therefore not created" is load-bearing now, and a property that holds because of a platform default is one nobody states and nobody notices losing. Asserted under both shells: `bash -e -c` is Actions' own invocation, and a shell handed no flags is the only one under which deleting the line changes anything.

## Not done here, deliberately

`publish` still has no `skip-existing`, so any *other* post-upload failure in `announce` is still unrecoverable by the documented retry. That is a change to the job that mints the OIDC token, it alters what a re-run of an irreversible step means, and it belongs to the release engineer rather than to this fix. Filed as a follow-up rather than folded in.

## Supersedes #665

#665 is the finding, opened red on purpose and not to be merged. Its tests are in this branch. Close it in favour of this.

No version bump, no stamp, no tag — the release is the operator's. The entry is staged as `docs/news/unreleased-a-release-body-fits-where-it-is-sent.md`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -110,7 +110,7 @@ jobs:
             echo "::error::this run names $claimed ($said) but pyproject.toml says $pkg — refusing to publish"
             exit 1
           fi
-      - name: Refuse to publish a version whose news entries cannot be rendered
+      - name: Refuse to publish a version whose notes cannot be rendered, or would be refused
         run: |
           # The version comes from pyproject, not the tag: it is defined on BOTH trigger
           # paths (a workflow_dispatch retry has no tag to read), and the step above has
@@ -125,15 +125,33 @@ jobs:
           # It exits non-zero for MORE than that now: an ordering field charter cannot
           # read, or two entries both claiming `lead: true` (#486); a frontmatter key
           # charter does not know, which is where `Security: true` lands (#503); or a file
-          # in docs/news that is not an entry at all. So the annotation names the two
-          # SHAPES — no entry, or something the entries declare that charter cannot honour
-          # — and defers to the command's own stderr for which, since that stderr is
-          # already one sentence per finding. `>/dev/null` keeps it, dropping only stdout.
-          # Prescribing `news stamp` for any of the second shape would be prescribing a
-          # no-op for a failure stamping cannot fix.
+          # in docs/news that is not an entry at all.
+          #
+          # And for a third shape, which is why this step is worth more than it was (#665).
+          # Rendering was never the failing step: this command exits 0 on a 300,000-
+          # character body, because producing the string is what it was asked to do. The
+          # claim nobody was making is that the string FITS where announce sends it —
+          # GitHub refuses a release body over 125,000 characters outright rather than
+          # trimming it, and the announce job below waits on publish — so that refusal
+          # would land after an upload PyPI will not take back, on a path the documented
+          # workflow_dispatch retry cannot re-reach. charter now bounds the body it renders and
+          # refuses one it cannot bound, so the failure arrives HERE — ahead of test, build
+          # and publish — instead of at the one point in this workflow with no way back.
+          #
+          # The assertion lives in the command rather than in a step of its own precisely
+          # because this step and announce run the same command against the same tree: a
+          # second copy of the limit in shell would be a second answer to one question, and
+          # it is the announce-side answer that decides whether the release exists.
+          #
+          # So the annotation names the three SHAPES — no entry; something the entries
+          # declare that charter cannot honour; or a body GitHub would refuse — and defers
+          # to the command's own stderr for which, since that stderr is already one
+          # sentence per finding. `>/dev/null` keeps it, dropping only stdout. Prescribing
+          # `news stamp` for any shape but the first would be prescribing a no-op for a
+          # failure stamping cannot fix.
           version=$(python -c 'import tomllib;print(tomllib.load(open("pyproject.toml","rb"))["project"]["version"])')
           if ! python -m charter news --for "$version" >/dev/null; then
-            echo "::error::cannot render the release notes for $version: either no docs/news/$version-<slug>.md (every published version ships an entry, including a patch — run: charter news stamp $version), or something those entries declare that charter cannot honour — an ordering, a frontmatter key, or a file in docs/news that is not an entry. See the error above for which."
+            echo "::error::the release notes for $version are not publishable: either no docs/news/$version-<slug>.md (every published version ships an entry, including a patch — run: charter news stamp $version), or something those entries declare that charter cannot honour — an ordering, a frontmatter key, or a file in docs/news that is not an entry — or a body GitHub would refuse as too long, which is a failure this step exists to catch before the upload rather than after it. See the error above for which."
             exit 1
           fi
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -226,6 +226,14 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
+          # Written out rather than inherited. Actions runs `run:` under `bash -e`, so this
+          # step already aborted on a failed command — but "the render refuses and the
+          # release is therefore not created" is load-bearing now (#665), and a property
+          # that holds because of a platform default is a property nobody in this file
+          # states. `guard` spells it out for the same reason. `-u` besides: every variable
+          # read below is one this workflow sets, and an empty `$tag` is a `gh release
+          # create` against whatever it decides that means.
+          set -eu
           version=$(python -c 'import tomllib;print(tomllib.load(open("pyproject.toml","rb"))["project"]["version"])')
           tag="v$version"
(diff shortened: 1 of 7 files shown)


diazoxide:
Filed the follow-up this deliberately does not do: #673 — `publish` has no `skip-existing`, so *any* post-upload failure in `announce` is unrecoverable by the documented retry, not just the one fixed here. That change touches the OIDC job and alters what a re-run of an irreversible step means, so it is the release engineer's call rather than this fix's.

diazoxide:
Deletion sweep on d00bfbf came back **4 survivors**, all in `charter/news.py`, and none of them an equivalent mutant. Closed in af1f7c9, each with the test it was asking for:

| line | what survived | why nothing caught it |
|---|---|---|
| `:644` | `_BODY_BUDGET = 100_000` | Pinned upward by the headroom case and free downward — every other assertion in the file reads a *smaller* budget as safer, so a budget of 8,000 turned every release into a table of contents with the whole suite green. Now asserts the claim the number's own docstring makes: the bound reaches for at most one stamped version. |
| `:649` | `_part`'s `.rstrip()` | Looks dead, isn't. `persona.parse` returns a stripped body, so for every entry that *has* one the strip is a no-op — what it is actually for is an entry with a headline and nothing under it, where without it the heading keeps the two newlines meant to separate it from a body that is not there. Nothing tested that shape. |
| `:723` | the elision rule's `if shown` | One half was asserted — no rule when nothing is above it — and "always absent" passed that just as well. Entry bodies write 122 `##` headings of their own across `docs/news`, so the rule is what makes the break a break. |
| `:791` | `<=` on the candidate | The direction was pinned, the boundary was not. `<` trades a note's whole text for a link to buy a character nobody needed. Reached by moving the dial to a real candidate's length rather than by sizing a fixture to the character. |

Each verified by hand to redden exactly the case named and nothing else. Full suite green at 8,665 tests; sweep re-running on the new tip.

diazoxide:
Sweep on the final tip: **1 survivor**, down from 4, with 11 of 12 mutations pinned.

```
[1/12]  pinned    news.py:663  [swap-synonym]     _part's .rstrip
[2/12]  pinned    commands.py:2593 [shift-boundary] sent > RELEASE_BODY_MAX
[3-4]   pinned    news.py:702  [collapse-ifexp]   tag-vs-branch ref, both arms
[5-6]   pinned    news.py:737  [collapse-ifexp]   the elision rule, both arms
[7]     pinned    news.py:798  [drop-if]          the whole body fits -> return it
[8]     pinned    news.py:805  [drop-if]          this candidate fits -> return it
[9]     pinned    news.py:798  [shift-boundary]   <= on the whole body
[10]    pinned    news.py:805  [shift-boundary]   <= on the candidate
[11]    pinned    news.py:618  [retune-constant]  RELEASE_BODY_MAX = 125_000
[12]    SURVIVED  news.py:658  [retune-constant]  _BODY_BUDGET = 100_000
```

The survivor is an **explained equivalent, not a missing test.** `retune-constant` rewrites `100_000` as `100_001` and the suite stays green because it must: this is a policy dial, and one character of it changes no output anyone can name. What is pinned is the *window* the dial has to sit in, from both ends and for different reasons — `test_the_staged_release_has_headroom` from above so GitHub keeps real slack, and `test_the_bound_is_an_exception_rather_than_the_normal_path` from below so a budget small enough to turn ordinary releases into a table of contents cannot pass while every length assertion still does. 100,001 is inside that window, which is why it survives and why that is correct.

The reasoning is in the constant's own docstring rather than anywhere else, because the sweep has no suppression list and this needs none. The only test that would redden on `100_001` is one asserting the literal back at itself, which is the shape this repository has watched fail more than once.

Note that `RELEASE_BODY_MAX = 125_000` right above it *is* pinned — GitHub's number is a fact and is asserted against an independently-stated copy; charter's is a choice and is asserted into a range. That difference is the whole of why one survives and the other does not.

PR #693, opened 2026-08-30
A chat switch that cannot move you stops reporting success and tearing your panels down (#684)

### LINKED ISSUE #684: A cross-session chat switch reports success and then tears down the panels of the chat you are looking at
Found by an adversarial review of #668 (Phase 5, Stage 5b). Verified by reading and by measurement against real tmux 3.7c.

## The guard measures a name; the property is a tmux session on a server

`chats.of_workspace` (`charter/frame/chats.py:142-144`) decides which chats are in a roster with one test:

```python
if is_chat(n) and state.frame_workspace(n) == workspace
```

It never asks `state.frame_server`, and it cannot ask which tmux *session* the chat's window is in. But `cmd_launch` makes the tmux **session** the workspace (`session = state.workspace_prefix(ws)`, `commands_frame.py:3515`), and `cmd_chat` sends its `select-window` to **the source's** socket with **the target's** pane id (`commands_frame.py:4940`).

Both halves of that record move at runtime:

- `chats.roster` asks `state.workspace_for(fid)`, whose rung 1 is what `charter workspace use <other>` typed at the agent writes — *"it moves the panels too"* is a documented promise;
- `switch.to_workspace` (`charter/frame/switch.py:192`) calls `state.record_workspace(fid, name)`, rewriting rung 2 as well.

So after one `F2 → workspace → beta` inside chat `api.1`, `of_workspace("beta")` returns `api.1` (a window of tmux session `api`) **and** `beta.1` (a window of tmux session `beta`) in one roster. `chats.check` says ok — its last refusal, `pane_of(chat) is None`, guards that the pane **record is well-formed**, never that the pane is somewhere this client can be moved to.

## Why nothing catches it downstream

Measured on tmux 3.7c, my own socket:

```
$ tmux -L … select-window -t %3        # %3 is a window of session B; client is in session A
rc=0
after:  A current=0   B current=1      # session B moved. The client did not.
```

`cmd_chat` checks `selected.returncode != 0` and gets **0**, so the `has no window any more` refusal never fires. It then runs:

```python
here = _relayout_target(fid)
if here is not None:
    _apply_arrangement(fid, where=here, want=[])      # tears down the chat still on screen
```

This is verbatim the failure #668's own last commit says it closed — *"charter would have reported a switch that did not happen and then torn this chat's panels down around it"* — closed for the `select-window -t ""` spelling only.

## Other reachable routes to the same shape

- `charter workspace use beta` typed inside the frame (rung 1 alone).
- A chat in the operator's own tmux and a chat on charter's private server, same workspace: pane ids are **per-server**, so `%3` on `-L charter` is somebody else's pane entirely.
- An old-shape `{ws}-{pid}` frame: `roster` folds the asker in only when `is_chat(fid)`, so an old frame is absent from its own roster and `check`'s `chat == fid` refusal cannot fire.

## Shape of a fix

The refusal wants a property, not a spelling: *is this chat reachable by the client that is asking?* That is at minimum `state.frame_server(chat) == state.frame_server(fid)`, and beyond that the session the target's window is in — which only tmux can answer, and which `cmd_chat` is already positioned to ask before it tears anything down. Teardown must be gated on the client having actually moved, not on `select-window`'s return code.

Reported alongside #668. Not fixed here — the right refusal is a design call about how much tmux `chats` is allowed to ask.

diazoxide:
Closes #684.

## Reproduced against a real server first

Driving the real `cmd_chat` on `main`, two chats whose `workspace` files both say `beta`,
their windows in two tmux sessions:

```
CROSS-SESSION (main)
  said            : []
  api current     : @0 -> @0        <- the client never moved
  api.1 panes     : ['%0']          <- its two panels are gone
  VERDICT: FAIL
```

and on this branch, on tmux 3.7c and on tmux 3.2 alike:

```
CROSS-SESSION
  said            : ["cannot switch: chat 'beta.1' is a window of another tmux session, …"]
  api current     : @0 -> @0
  beta current    : @1 -> @1        <- the other session was not moved either
  api.1 panes     : ['%0', '%2', '%3']
  VERDICT: PASS
SAME-SESSION
  api current     : @0 -> @2  target window: @2
  api.1 panes     : ['%0']          <- the ordinary switch still tears down what it leaves
  VERDICT: PASS
```

## Why the return code could not catch it

```
$ tmux -L … select-window -t %3     # %3 is a window of session B; client is in session A
rc=0
after:  A current=@0   B current=@2
```

A command that exits 0 having acted on the wrong target is indistinguishable from one that
acted on the right one. `chats.check`'s last refusal guarded that the pane **record** was
well-formed, never that the pane was somewhere this client could be moved to.

## Reachable, not hypothetical

`chats.of_workspace` matches `state.frame_workspace` — a FILE. `charter workspace use beta`
typed inside chat `api.1` rewrites it ("it moves the panels too" is a documented promise)
and `switch.to_workspace` rewrites it again. The tmux session is fixed at launch. So one
`F2 → workspace → beta` puts `api.1` and `beta.1` in one roster with their windows in two
sessions, and every check said yes.

## The two properties

- **Server** — a record, so `chats.check` asks it. Pane ids are per-server: `%3` recorded by
  a chat on charter's own socket names a real, live, unrelated pane on the operator's. No
  default is filled in for a missing marker; every chat this charter launches records one
  on both paths, and `of_workspace` keeps pre-chat `{ws}-{pid}` frames out of the roster.
- **Session** — a fact only tmux holds, which moves while the palette is open, so `cmd_chat`
  asks it of both chats before aiming anything. Refusing *before* the select is what keeps
  charter from moving somebody else's session.

And the teardown is gated on the client having **moved** — read back from tmux — rather
than on `select-window`'s status.

## The reading is the guard, not the status

Measured on 3.7c: `display-message -p` against an unresolvable target answers **rc 0, empty
stdout, no stderr**. So the placement readings are held to their shape (`$<digits>`,
`@<digits>`) rather than to their exit code — the same mistake this issue is about, one
command over. An empty `-t` is not nothing either: it resolves to the current window.

## The audit the review asked for

Every other place in `commands_frame` that targets a pane or window was walked. Findings
are in the report; two are worth naming here:

- `_say_on_screen`'s `-t <fid>` is a **name**-shaped target that works today only by
  accident of tmux's fallback (a chat id starts with its session's name, and tmux resolves
  `api.1` to session `api`). Sound in the reachable configuration; noted, not changed here.
- `layout.chat_window_argv`'s `-t <session>` genuinely breaks for a workspace whose name
  contains a dot — `new-window -t api.2` answers `can't specify pane here`, rc 1, so a
  second chat can never be opened in that workspace. Separate defect, filed separately.

## One deliberate reversal

A chat whose own harness-pane record is unreadable now refuses the switch instead of moving
the client anyway. The previous argument ("leaving the operator where they asked not to be
is the worse failure") held while the only cost of not knowing where you stand was an
untidy background window. That record is what says which session this client is in.
`TheSwitchIsFourStepsInOneOrder.test_a_chat_whose_own_pane_record_is_broken_switches_nothing`
carries the reasoning.

## Verification

- Full suite green on 3.14: **8858 tests, OK**.
- New real-tmux class `ASwitchAcrossSessionsIsRefusedRatherThanReported`, run on 3.7c and
  on 3.2. It measures `select-window`'s rc against another session's pane on *this* tmux
  first, so the guard cannot outlive the behaviour it is for.
- Hand deletion sweep, applied and restored with sha256 **and** `git diff --quiet` checked
  on every restore. All ten guards go RED; two survivors on the first pass were fixed at
  the source rather than papered over — `_session_window`'s shape check was subsumed by the
  comparison it feeds (deleted), and `_pane_place`'s pane-id check was masked by
  `chats.pane_of` on every route from the top (now asserted of the reader directly).

### DIFF
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -252,6 +252,38 @@ _WINDOW_SIZE_FORMAT = "#{window_width}:#{window_height}"
 #: the pane after the side panels have been put back where they belong.
 _PANE_WIDTH_FORMAT = "#{pane_width}"
 
+#: The format that says WHERE a pane is — the session and the window it belongs to, as
+#: tmux's own ids rather than as names (#684).
+#:
+#: Both halves, in one ask, because both are needed at the same instant and by the same
+#: caller: `cmd_chat` has to know the target is in the session it is switching WITHIN
+#: before it selects anything, and has to know which window to expect the client on
+#: afterwards. Ids and never `#{session_name}`/`#{window_name}`: a name is not an
+#: identity (`_CHAT_OPTION` records the measurement — `allow-rename on` let a pane's own
+#: output rename its window), the ids are what tmux hands back for its own targets, and
+#: `#{session_id}` is about to be one.
+#:
+#: `:` is the separator because neither id can contain it: tmux spells them `$<digits>`
+#: and `@<digits>`, which is exactly what :data:`_SESSION_ID_RE` and :data:`_WINDOW_ID_RE`
+#: hold them to on the way back.
+_PANE_PLACE_FORMAT = "#{session_id}:#{window_id}"
+
+#: The format that says which window a SESSION is on right now — what a switch has to
+#: read back before it may believe the client moved. Measured on tmux 3.7c and 3.2:
+#: `display-message -p -t $N` resolves to that session's CURRENT window, so this is the
+#: reading `select-window`'s return code is not.
+_WINDOW_ID_FORMAT = "#{window_id}"
+
+#: tmux's own shapes for a session and a window, held to for :data:`_PANE_ID_RE`'s reason
+#: and at its boundary: both come back through text tmux re-parsed, and the session id
+#: goes straight back out as a `-t` target. A format that expanded to nothing — which is
+#: what tmux answers for a target it cannot resolve, with rc **0** and no stderr (measured
+#: on 3.7c) — fails these rather than becoming an empty target, and an empty `-t` resolves
+#: to the CURRENT window, which is the exact shape #668's own last commit closed for
+#: `select-window`.
+_SESSION_ID_RE = re.compile(r"\$[0-9]+")
+_WINDOW_ID_RE = re.compile(r"@[0-9]+")
+
 #: How long `_wait_for_harness` leaves between asks. There is no `attach` to block on
 #: inside an operator's tmux — the operator is already attached, to their own server —
 #: so the launcher watches the pane instead, and this is the cost of that: one local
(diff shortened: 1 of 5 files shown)


diazoxide:
## The two `not measured` mutations, named

```
[1/11] UNRESOLVED  charter/commands_frame.py:289 [retune-string] r"@[0-9]+"    (46 module(s))
[2/11] UNRESOLVED  charter/commands_frame.py:288 [retune-string] r"\$[0-9]+"   (46 module(s))
  charter/commands_frame.py:288  in `<module>`  [retune-string]  no tests ran
  charter/commands_frame.py:289  in `<module>`  [retune-string]  no tests ran
```

`_SESSION_ID_RE` and `_WINDOW_ID_RE` — the two constants this branch adds.

**They did not hang, and they did not time out.** `retune` shifts each character one along
inside its class, which for a digit is `0 -> 1` and `9 -> 0` — and inside a character class
that **inverts the range**:

```
$ python3 -c "import sweep, re; print(repr(sweep.retune(r'\$[0-9]+')))"
'\\$[1-0]+'
$ python3 -c "import re; re.compile(r'\$[1-0]+')"
re.error: bad character range 1-0 at position 3
```

The mutant is not a program. `re.compile` raises while `charter/commands_frame.py` is being
imported, every selected test module fails to load, zero tests run, and the sweep refuses —
correctly — to call that a pin. Reproduced by hand on this branch: applying the tool's own
mutant and running one covering module ends in `re.PatternError: bad character range 1-0`
with no test result at all.

The timings agree with "no tests ran" and not with the report's UNRESOLVED prose ("it timed
out rather than failing"): the second of the two finished in **32 seconds**, against ~55s
for a `retune-string` on a plain constant with the same 46-module selection.

This is **#698**, filed against the tool. It is not specific to this branch —
`tmuxctl.PANE_ID_RE` is `r"%[0-9]+"` and retunes the same way; it has simply not been in a
diff yet. `retune`'s own docstring already makes this argument for the neighbouring case
(a character after a backslash is left alone, because shifting it produces
`re.error: bad escape`); a character range is the same hazard one step over.

## Hand-scored, with the question asked in a form that runs

Restores verified with sha256 **and** `git diff --quiet`:

| mutation | verdict |
|---|---|
| `_SESSION_ID_RE` takes the window sigil `r"@[0-9]+"` | **RED** — 6 tests |
| `_SESSION_ID_RE` -> `r"\$.+"` | **RED** |
| `_SESSION_ID_RE` -> `r".+"` | **RED** |
| `_WINDOW_ID_RE` takes the session sigil `r"\$[0-9]+"` | **RED** — 6 tests |
| `_WINDOW_ID_RE` -> `r".+"` | survives, deliberately — see below |

**The first `r".+"` run was a genuine survivor**, and the gate was right to make me look.
That id is the one `_session_window` sends straight back out as a `-t` target, so its
alphabet is #475's rule and nothing asserted it. It does now:
`test_a_session_id_that_is_not_tmuxs_own_never_becomes_a_target` — a placement whose
session field is `$0;kill-server` is refused, and no tmux command line ever carries it.

`_WINDOW_ID_RE` widened to `r".+"` still survives and that is the honest answer rather than
a gap: that value is only ever **compared**, against a reading taken from the same tmux a
moment later, so beyond telling a window id from a session id its strictness decides
nothing the comparison does not. The constant now says so at the line, so the next sweep
that asks gets an answer instead of a shrug.

Full suite green after the change: **8860 tests, OK**. Real tmux 3.7c and 3.2 re-run:
cross-session refused, same-session switch still tears down what it leaves.

diazoxide:
Merging on an explained `no verdict`, and recording why so the next reader does not have to re-derive it.

## The two unmeasured mutations are not a program

`charter/commands_frame.py:288-289`, both `[retune-string]` on `_SESSION_ID_RE` and `_WINDOW_ID_RE`. `retune` shifts each character one along its class, so `0 → 1` and `9 → 0` — which inside a character class **inverts the range**. Verified independently:

```
'\$[1-0]+'  ->  PatternError: bad character range 1-0
'@[1-0]+'   ->  PatternError: bad character range 1-0
```

`re.compile` raises while `commands_frame` is being imported, every selected module fails to load, and zero tests run. **The gate is right to refuse to call that a pin** — that is exactly what `not measured` is for, and it is doing its job.

Not a hang: the second finished in **32 seconds** against ~55s for a `retune-string` on a plain constant with the same 46-module selection. The branch's own earlier report said "it timed out"; that prose was wrong and would have sent the next reader to raise a timeout forever.

## Asking the question in a form that runs found a real survivor

```
_SESSION_ID_RE takes the window sigil   RED (6 tests)
_SESSION_ID_RE -> r"\$.+"                RED
_SESSION_ID_RE -> r".+"                  RED — after a new test; it SURVIVED first
_WINDOW_ID_RE  takes the session sigil   RED (6 tests)
_WINDOW_ID_RE  -> r".+"                  survives, deliberately
```

The `r".+"` survivor was genuine and load-bearing: that id is what `_session_window` sends back out as `-t`, so its alphabet is **#475's rule** and nothing asserted it. It does now — `$0;kill-server` in the session field is refused and reaches no command line. `_WINDOW_ID_RE`'s strictness stays unasserted because that value is only ever *compared*, and the constant now says so at the line, so the next sweep gets an answer rather than a shrug.

**That is the argument for holding a merge on `no verdict`.** Merging on the strength of the other three PRs being clean would have shipped an unasserted alphabet on a value that reaches a command line.

## Why merge rather than wait

The cause is a defect in `tools/sweep.py`, filed as **#698**, and it is not specific to this branch — `tmuxctl.PANE_ID_RE` is `r"%[0-9]+"` and retunes identically; it has simply not been in a diff yet. Holding a correct fix hostage to a tool defect would cost more than it buys, now that the two are named, reproduced, and answered by hand.

`retune`'s own docstring already makes this argument for the neighbouring case — a character after a backslash is left alone, or you get `re.error: bad escape`. A character range is the same hazard one step over.

diazoxide:
## Re-run: same two, and everything else is pinned

```
shard 1 of 2: [3/18] UNRESOLVED  charter/commands_frame.py:296 [retune-string] r"\$[0-9]+"
shard 2 of 2: [4/17] UNRESOLVED  charter/commands_frame.py:297 [retune-string] r"@[0-9]+"
merged: 0 unpinned, 0 in masked cluster(s), 0 platform-deferred, 2 unresolved, 0 not applied, 33 pinned
        SURVIVED: 0 on both shards
```

The same two constants — the line numbers moved 288/289 -> 296/297 only because the comment
above them grew. The mutation count went 11 -> 35 and the pinned count 9 -> **33**, with
**0 survived**, so the two unscoreable ones are the only thing between this and
`no survivors`.

**And it will stay that way.** This is not a flake and not a load problem: the mutant
`r"\$[1-0]+"` cannot be imported, so no re-run and no quieter machine will ever score it.
The gate goes green here only when #698 lands in `tools/sweep.py`.

So the choice is yours and I am not going to pretend it away:

1. **Merge on the hand evidence** — five mutations run in a form that executes, sha256 and
   `git diff --quiet` verified on every restore, four RED and the fifth documented at the
   line as not load-bearing. The one that survived first grew a test, which is the whole
   point of the gate having stopped you.
2. **Or hold until #698 is fixed** and let the tool answer for itself. I can take that,
   but it changes how every branch in flight is scored, so I would rather be asked than
   assume — it is a shared gate, not this branch's.

What I will not claim is that `no verdict` here means the same thing as `no survivors`. It
does not, and #644 is why you were right to stop.

Issue #690, opened 2026-08-30
The docs snippet for turning on the chat bar builds a frame with ONLY the chat bar


diazoxide:
Found by an adversarial review of #668 task 7. Verified by resolving the snippet through `instance.frame_of`.

`docs/frame.md:281-290`:

> Turn one on with a `[[frame.component]]` table, which is also how it gets a key of its own:
>
> ```toml
> [[frame.component]]
> use = "chats"
> edge = "top"
> size = 1
> key = "F9"
> ```

Copied verbatim into a plane's `charter.toml`, that resolves to:

```
slots      : ['chats']
components : ['chats']
```

No identity row, no attention strip, no repo table, no sidebar — because a
`[[frame.component]]` block **replaces the whole arrangement** (`charter/instance.py:1401-1403`),
it does not add to it. An operator following the instruction to "turn one on" turns
everything else off.

The caveat does exist — but 720 lines further down, in the toggle-keys section
(`docs/frame.md:1008-1010`), not where the reader is being told to write the table.

## Shape of a fix

Either show the whole arrangement in the snippet (the shipped four plus the bar, which is
what an operator actually has to write), or put the "this replaces your frame, list
everything you want" sentence immediately above the fence. The first is longer and cannot
be misread; the second keeps the section short and depends on the reader not stopping at
the code block.

Related: #687 (a `size` other than 1 on the same table silently drops the arrangement).

diazoxide:
Fixed in #692 — and it was **four snippets, not one**: the chat bar, `visible = false`, the repo pin, and the provider example all listed a partial arrangement.

All four now write the whole thing out, and `EveryDocumentedArrangementIsAWholeArrangement` resolves **every** `[[frame.component]]` fence in `docs/frame.md` through the real `instance.frame_of`, holding it to the frame it prints. So a documented `size`/`bg`/`pad`/`edge`/`key` that charter would refuse now goes red here instead of in an operator's terminal.

Verified on `main`: 6 tests, OK.

Worth recording why this mattered more than a typo. I was about to hand the operator the chat-bar snippet so they could turn the bars on — an arrangement is refused whole (#535), so following our own documentation would have silently dropped every other panel and cost them an evening.

Issue #687, opened 2026-08-30
A `size` on the `chats` or `workspaces` bar silently drops the whole arrangement, and the reason the guard gives is false for those two


diazoxide:
Found by an adversarial review of #660 + #668. Small, but it is the project's signature shape: a guard matching a **declaration form** where the property is *is this number read anywhere*.

## The stated rule, and where it stops holding

`instance._built_in_size` (`charter/instance.py:1520`):

> **A committed value is accepted exactly where something reads it, and refused everywhere else** […] `layout._derive` turns each component's declared size into `layout.SLOT_SIZE` once, at import, and `layout._size_of` answers `top`, `bottom` and `right` out of that table on every launch. There is no per-plane override between `charter.toml` and `split-window` for those three, so a `size` that disagreed with the declaration would be a value read, validated, stored and then ignored.

The implementation asks a different question — which **policy class** the component declared:

```python
if isinstance(c.size, Content):
    return Fixed(value)                    # repos
if isinstance(c.size, Fixed) and value == c.size.n and not isinstance(value, bool):
    return c.size                          # echo only
return None                                # refuses the whole arrangement (#535)
```

`chats` and `workspaces` are `Fixed(1)` (`charter/frame/builtins.py:465-470`), so they take the echo branch. But the argument behind that branch does not apply to them: **they are not in `SLOT_SIZE`**, because `layout._derive` builds it from *placed* components and neither is placed by default. Their size is read at launch through `_placed_here()` → `_policy_cells`, which is a live read of the committed number.

## Measured

```
[{'use': 'chats', 'size': 1}] -> slots: ['top','bottom','repos','right','chats']
[{'use': 'chats', 'size': 2}] -> slots: ['top','bottom','repos','right']    # arrangement gone
[{'use': 'workspaces','size':3}] -> slots: ['top','bottom','repos','right'] # arrangement gone

SLOT_SIZE has chats? False | SLOT_EDGE has chats? False
_size_of("chats") = 1   _edge_of = top   _is_fixed_row = True
slot_sizes([...]) = {'top': 1, 'chats': 1, 'bottom': 1, 'repos': 6, 'right': 22}
```

So a number on `chats` **is** read — `_size_of` returns it and `slot_sizes` honours it — and the boundary refuses anything but `1` anyway, on the grounds that it could only be ignored. That is exactly backwards for these two.

## What the operator sees

[...]

PR #715, opened 2026-08-30
`charter` on its own opens the frame, once the plane names a harness (§4a)

diazoxide:
§4a of `docs/superpowers/specs/2026-08-30-charter-opens-like-an-ide.md`. The operator's ask, in their words: `charter claude` → `charter`.

```toml
[harness]
default = "claude"
```

and then `charter`.

## What this is

Bare `charter` is **rewritten** into `charter claude` in `cli.main` and run. Not a second route into `cmd_launch` — the same argv, through the same three splitters, into the same handler, so the workspace picker, `--no-frame`, `--probe`, `--workspace`, the chat it opens and the exit code it carries back are the behaviours the typed form already has rather than a second set of answers about them. `TheBareCommandLaunchesTheDeclaredHarness.test_it_is_the_same_call_typing_the_harness_makes` asserts that as an equality on the two namespaces.

Every existing subcommand is untouched, `charter claude`, `charter opencode` and `charter status` included.

## There was no `[harness]` section at all

Zero references in `instance.py`, so this is a new config section plus a new dispatch mode, not three lines. It follows `[update]`'s shape rather than inventing a second one: a `HARNESS_FIELDS` table paired `(default, toml_key)`, a `harness_of` reader, a closed set matched exactly, and **the matched constant stored rather than the file's own string** — which matters more here than it does for `channel`, because this value becomes `argv[0]` of a charter invocation, out of a file that arrives from somebody else's machine.

The legal names come from `harness.KINDS` (`instance.launchable_harnesses`), not a literal — `registry.py`'s own rule, *"a harness added to KINDS is covered everywhere the day it is registered"*. A tuple of three words in `instance.py` would go stale silently, and the symptom would be charter refusing a harness charter can launch. `frame` is excluded: the escape hatch carries no command of its own, so a plane defaulting to it would get `nothing to run` on every bare `charter`.

## The three grills

**Where does an unknown `default` get refused?** At the boundary, in `harness_of`, and it is *recorded* rather than swallowed. A refused value degrades to no default, and no default renders as argparse's usage message — byte-identical to what a plane declaring nothing gets. Silently, `default = \"clyde\"` would be indistinguishable from a key nobody wrote. Two readers say so:

```
$ charter
✗ charter: [harness] default = \"clyde\" in ~/plane/charter.toml is not a harness charter
  can launch — one of: claude, opencode, codex. Nothing was started.
```

and `doctor.check_control_plane_config`, which is exactly where `config.worktrees_root_for`'s own silently-ignored key is named, for the reason that function's docstring already gives: *\"the resolver itself runs inside `derive`, where there is nobody to tell.\"* `instance` still never raises — it is imported by every command including `charter --version`.

The refused value is rendered through `contain.readable` before it reaches a report line. It is a committed identifier a sentence is telling somebody to go and fix, so a newline must not forge a second line of charter's own output and an invisible codepoint must not name nobody (#498).

**What does `charter --version` do?** Nothing new. It is a flag on the root parser, so `[\"--version\"]` is a non-empty argv and the rewrite never sees it — `_bare_launch` returns on its first line. Pinned, because `commands_update._handoff` runs the newly installed `charter --version` and reads the last word back: a bare launch there would hang an upgrade inside a harness. `--help` likewise.

**Is `[harness]` the right section name?** `[persona] default` and `[workspace] default` already spell this exact idea with this exact word, and `[update]` is the shape for a validated closed set. Both followed; nothing new invented.

## The hazard

`cmd_launch` returns `bypass(argv)` — an `os.execvp` — when stdout is not a tty. Bare `charter` is argparse's usage error today, exit 2 with no side effect, so `charter 2>&1 | head` is a free probe for \"is charter installed\". Rewrite argv unconditionally and, on a plane that sets a default, that pipeline execs Claude Code: correct against every config anyone tested, wrong the first time something automated goes looking for charter. #687 and #690's exact shape, twice over.

**The non-tty path keeps printing usage, and it is a test.** `TheNonTtyPathStartsNothing` pins that the launcher is not reached, that `os.execvp` is never called, and that the text is argparse's own — not a new message that happens to exit 2. Stdout and stdout alone, because that is the stream `cmd_launch` branches on; a second, looser condition here is how the two would come to disagree about what interactive means.

The two gates (`not default`, `not sys.stdout.isatty()`) are two statements rather than one `or`, deliberately: they are different facts with different reasons, and an arm two inputs can both satisfy is an arm neither of them ends up testing.

## `claude`, verified rather than inherited

§2.8's reasoning holds on this tree. `record_harness_session` has exactly one caller — `statusline.py:2835` — and it needs `payload[\"session_id\"]`, which only Claude Code's `statusLine` hook supplies; `charter statusline --watch`, the remedy the other two are given, carries no payload. Codex and opencode both declare a `status-bar` deficit. So Claude Code is the only harness with a context gauge and the only one a future resume (§4e) could restore.

## Verification

- **Suite: 8972 tests, green** (`python3 -m unittest discover -s tests`), stdlib `unittest`, no new dependency.
- **Hand-run on charter's own plane** (§7's rule), under a real pty with `claude` off `\$PATH` so nothing is started: bare `charter` and `charter claude` are byte-identical (exit 127, same message). Piped, bare `charter` is the usage list and exit 2.
- **Nine hand-applied mutations, all red.** Each gate deleted in turn — the tty gate, the no-default gate, the refusal report, the `_bare_launch` call, the typed-argv guard, `refused` not being recorded, the `cli_name` filter, storing the file's own string, and the doctor row. The repo's own `tools/sweep.py` runs on this diff in CI.
- `config.HARNESS` joins `_planeguard._GUARDED_SETTINGS` **on the day it arrives**, which is what that list's own comment asks for: it is opt-in, and charter's dogfood plane now declares it, so an unisolated test would assert against whatever the machine running the suite commits.

## Not in scope

`charter -w foo` (§4k), the `workspaces` bar (§4b) and `charter status` learning about open/live (§4m) are their own stages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,6 +310,8 @@ you use. The browser lane additionally shells out to `npx`. That is the whole li
   owns the rectangles; charter fills the edges and never draws in the agent's own pane.
   `charter frame -- <cmd>` does it for a command charter has never met, and `--no-frame`
   (or piping the output anywhere) skips the frame entirely and carries the real exit code.
+  Name the one you use — `[harness] default = "claude"` — and the command is just
+  `charter`.
   → [docs/frame.md](docs/frame.md)
 - **An unattended run that stops to ask a question.** Every git operation authenticates
   with that repo's own forge CLI token over HTTPS — never an SSH key, never signing —
--- a/charter.toml
+++ b/charter.toml
@@ -128,3 +128,17 @@ key = "F7"
 # Track main instead of the last release. `charter update` installs straight from git;
 # nothing is published, and PEP 610 gives `charter --version` the exact commit.
 channel = "dev"
+
+[harness]
+# Bare `charter` opens the frame with this harness — the same command as `charter
+# claude`, which still works and is what this expands to. Claude Code and not one of the
+# other two, and the reason is not preference: it is the only harness that writes its
+# session id where charter can read it (`state.record_harness_session`, whose one caller
+# is Claude Code's own `statusLine` hook), so it is the only one with a context gauge in
+# the top bar. codex and opencode both carry a `status-bar` deficit saying charter cannot
+# render into their chrome at all.
+#
+# A plane that writes no default keeps argparse's usage message, which is what charter does
+# for every plane but this one. And `charter 2>&1 | head` still prints that usage rather
+# than exec-ing the harness: a script probing for charter must not start a session.
+default = "claude"
--- a/charter/cli.py
+++ b/charter/cli.py
@@ -1723,8 +1723,13 @@ def _bare_launch(argv: list[str]) -> tuple[list[str], int | None]:
     from . import config
 
     declared = config.HARNESS
+    # Truthiness, not `is not None`, and the same test `doctor` makes of the same key.
+    # `contain.readable` never returns a blank string, so on anything `harness_of` produced
+    # the two are the same question — but `config.HARNESS` is a module attribute anything
+    # in-process can assign, and a planted `{"refused": ""}` would otherwise print a
+    # refusal naming nothing at all instead of falling through to the usage message.
     refused = declared.get("refused")
-    if refused is not None:
+    if refused:
         util.err(f'charter: [harness] default = "{refused}" in '
                  f'{util.short_path(config.ROOT / "charter.toml")} is not a harness '
                  f'charter can launch — one of: '
--- a/charter/cli.py
+++ b/charter/cli.py
@@ -20,6 +20,7 @@ from . import (
     contain,
     harness,
     hooks,
+    instance,
     statusline,
     toolgate,
     util,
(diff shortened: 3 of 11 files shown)


diazoxide:
## The deletion sweep: 19 pinned, 0 survived, and the 2 it could not measure, resolved by hand

[Run 33332143877](https://github.com/diazoxide/charter/actions/runs/33332143877) — shard 1 of 1, 44m19s:

```
pinned            : 19
SURVIVED          : 0
UNRESOLVED        : 2

UNRESOLVED — these mutations have no verdict
  charter/instance.py:2039  in `<module>`    [retune-string]  no tests ran
  charter/instance.py:2123  in `harness_of`  [retune-string]  no tests ran
```

Both are the `retune-string` operator on `"default"`, and both **timed out rather than failing** — the shard's own rule is that machine load is not evidence, so a timeout is neither a pin nor a survivor. Line 2039 is `HARNESS_FIELDS = {"default": (None, "default")}` (two literals on one line) and 2123 is `toml_key = HARNESS_FIELDS["default"][1]`.

Reproduced locally with the sweep's own retuned value (`sweep.retune("default") == "efgbvmu"`), against `tests.test_bare_charter_opens_the_frame`, with `PYTHONDONTWRITEBYTECODE=1` — the two spellings are the same length, so a same-second restore can otherwise be served from a stale `.pyc` and report a false colour in either direction:

```
RED       instance: FIELDS key retuned  [sweep #11]   -> HARNESS_DEFAULTS has no "default"
RED       instance: toml_key retuned    [sweep #11]   -> [harness] default is no longer read
RED       instance: lookup retuned      [sweep #14]   -> KeyError: 'efgbvmu'
GREEN     baseline, unmutated
```

So all 21 mutations are accounted for: 19 pinned in CI, 2 pinned by hand, none survived.

The same harness also ran nine hand-written deletions of the guards this branch adds — the tty gate, the no-default gate, the refusal report, the `_bare_launch` call, the typed-argv guard, `refused` not being recorded, the `cli_name` filter, storing the file's own string, and the doctor row — all red.

PR #722, opened 2026-08-30
A provider's `ctx.chrome` colour reaches its pane instead of printing as text (#707)

### LINKED ISSUE #707: A provider's `ctx.chrome` colour is escaped on the way out, so the channel docs/frame.md documents renders as literal text
`docs/frame.md:1169-1176` documents a provider writing `ctx.chrome['heading']` into a row. `charter/frame/registry.py:662` escapes foreign rows — `escape=cid in self._foreign` — so it never reaches the pane as colour.

**Verified round trip:** a provider's `\x1b[1mMetrics\x1b[0m` arrives in the pane as the literal text `\x1b[1mMetrics\x1b[0m`.

## Why this is a contradiction rather than a limitation

The escaping is right and should stay: a foreign component must not be able to paint outside its rectangle or forge charter's own chrome. #604 then landed `ctx.chrome` recipes *on the same day*, which hand a provider colour it cannot use.

So charter serves a colour channel and then strips it, and the documentation tells providers to use it.

**Tests pin the escaping, and tests pin that recipes are *served*. Nothing tests that a provider's colour survives to a pane** — which is why the two could disagree for two days without anything going red. That gap is the defect underneath the defect: the two halves are each asserted, and the property that spans them is not.

## What a fix has to decide

Either the recipes are not for providers — in which case `docs/frame.md` is wrong and `ctx.chrome` should not be served to a foreign component at all — or a provider's chrome must survive escaping, which means charter distinguishing *its own recipes echoed back* from *arbitrary escapes a provider invented*. The second is the harder and more useful one, and it is exactly the "spelling versus property" shape this repo keeps meeting: matching the recipe's text is not the same as knowing charter served it.

Whichever way it goes, **the test that would have caught this asserts the round trip**, not either end.

## Provenance

Found while triaging #605 (closed). It is the strongest of three findings that survived that experiment, and it got *worse* after the PR was written rather than better.

diazoxide:
Closes #707.

`docs/frame.md` shows a provider writing `ctx.chrome['heading']` into a row. `registry.py`
escaped every foreign row with `contain.one_line`, so that row reached the pane as the
literal text `\x1b[1mMetrics\x1b[0m`. Charter served a colour channel, stripped it, and
documented it. Verified round trip before the fix:

```
served heading: '\x1b[1m'
drawn:          ('  \\x1b[1mMetrics\\x1b[0m',)
```

## What changed

The containment stays; a hole is cut in it exactly the size of the vocabulary charter
already publishes. `chrome.contain_row` passes an SGR through when **every one of its
parameters is one of `chrome.served_params()`** — the numbers `chrome._role_values()` is
made of — and escapes everything else through `contain.one_line`, which is what the whole
row used to get.

That is safe on the axis the escaping was written for. An SGR costs zero columns and says
nothing about position, so it cannot reach the pane beside it; and the seven roles are
plain attributes and the sixteen names in the operator's own palette, so it cannot name a
colour charter did not pick. A cursor move, an erase, an OSC title string, a cube index, a
24-bit triple, a background colour, a newline: each still arrives as the characters of its
own escape. One parameter outside the vocabulary takes the whole escape with it, so
`\x1b[1;41m` is contained entire rather than half-kept. Under `NO_COLOR` or a pane that is
not a terminal nothing is exempt, because §3.2's rule is about what charter *emits*, not
only about what it serves.

## The spelling-versus-property trap, and why the question changed

The issue asks charter to tell *its own recipe echoed back* from *an escape a provider
invented*, and notes that matching the recipe's text is not the same as knowing charter
served it. That is right, and it is unanswerable: a row is a string, and
`ctx.chrome['heading']` and a hard-coded `"\x1b[1m"` are the same six characters by the
time `draw` sees them. There is no provenance to recover.

So the question asked is the one that is answerable and is the one that matters — *is what
this escape does inside charter's vocabulary* — and it is asked of the parameters as
numbers, the way `chrome.resets_everything` already asks its own:

```
spelling         params    equals a served recipe   every parameter served
'\x1b[1m'        '1'       True                     True
'\x1b[01m'       '01'      False   <- MISSED         True
'\x1b[1;32m'     '1;32'    False   <- MISSED         True
'\x1b[m'         ''        False   <- MISSED         True   (empty is 0)
'\x1b[38;5;236m' '38;5;2…' False                     False
'\x1b[41m'       '41'      False                     False
```

The three `MISSED` rows are a provider composing charter's own vocabulary in a legal
spelling charter does not happen to write, and escaping those while keeping `\x1b[1m` would
be the string-match defect this repo has now met six times.

## The deliverable is the spanning test

One test pinned that the recipes are *served* and another pinned that a foreign row is
*contained*; nothing spanned them, which is the defect underneath the defect.
`TheRecipeSurvivesToThePane` installs a real distribution (the same `_SitePackages` fixture
`test_component_providers` uses — a real `.dist-info` on `sys.path`, nothing stubbed) whose
renderer reads `ctx.chrome`, draws it through the registry, and asserts what arrives:

* the docs' own worked example comes out as colour, and *not* as `\\x1b`;
* every one of the seven roles survives, one case each;
* the identical renderer run as one of charter's own components produces the **same tuple**
  — §7's exit criterion, which is what "indistinguishable from a built-in" has to mean;
* a hard-coded cube index still arrives as text;
* under `NO_COLOR` the recipes are empty *and* a hard-coded `\x1b[1m` is escaped.

Plus `TheBudgetNeedsNoSecondGuard`, which proves over every row of up to three pieces of a
ten-symbol alphabet (1 111 rows × 6 limits, exhaustive rather than sampled) that the answer
never exceeds `limit + 1`, never carries half an escape, and — for a row with no recipe in
it — **is `contain.one_line`'s own string character for character**. The containment did not
become cleverer or weaker.

## What the sweep found

**A deleted guard.** An `if used >= limit: break` was written first and removed. It was
there to stop `contain.one_line` being handed a negative limit, and it could not: `_pieces`
strictly alternates and a kept piece is appended only when it fits, so an escaped piece
always starts from `used <= limit`. A 71 711-call fuzz confirmed no negative limit is
reachable without it and that the `limit + 1` bound holds. `TheBudgetNeedsNoSecondGuard`
pins the property without the guard.

**Two survivors on the guard that stayed, both now pinned.** `tools/sweep.py` reported a
masked cluster on `keep and used + len(text) > limit`:

* *drop-conjunct* — dropping the `keep` half. Caught by
  `test_a_row_with_no_recipe_in_it_is_contained_exactly_as_it_always_was`, which the sweep
  ran before it existed.
* *shift-boundary* — `>` becoming `>=`. This was a genuinely missing test rather than an
  equivalent mutant: it drops an escape whose last character lands *on* the limit, which
  is inside the budget. `test_a_recipe_that_lands_exactly_on_the_budget_is_still_inside_it`
  now pins the boundary and the row one character longer as its control.

Every remaining arm was mutated by hand as well, each caught by a test with an input that
reaches only it — including the `isprintable()` fast path, whose removal correctly changes
nothing, which is why it is written as one expression the way `tui.sanitize` writes the
same shortcut rather than as a guard no test could turn red.

## Cost, measured

This is the repaint path, so: one 24-row foreign pane through `registry._fit`, against the
`contain.one_line`-over-the-whole-row it replaces.

```
24 rows, 4 escapes each (the docs' own example)   323 us   was   73 us
24 rows, 12 escapes each (deliberately dense)     914 us   was  457 us
24 rows, no escape at all                         205 us   was   65 us
24 rows of 300 plain characters                   340 us   was  510 us
```

The last row is not a typo — `_escaped`'s `isprintable()` scan beats `one_line`'s
per-character loop, the same shortcut `tui.sanitize` takes. The worst case is half a
millisecond against `render`'s own measured 4 816 us, paid only by a pane charter did not
write. A first cut was four times that because `served_params()` was called per escape;
`is_recipe` now takes the vocabulary as a parameter and `contain_row` builds it once.

## Contract, stated

`docs/frame.md` now says what survives the trip and what does not, so the two halves of the
seam are documented in one place instead of contradicting each other in two.

### DIFF
--- a/charter.toml
+++ b/charter.toml
@@ -105,6 +105,31 @@ bg  = "brightblack"
 pad = 1
 key = "F5"
 
+# The two tab strips, between the identity row and the work area: what plane, then
+# which workspace, then which chat inside it. Split order is the geometry (#488).
+#
+# Neither ships placed (`instance.FRAME_SLOTS`) and that is deliberate — a plane with one
+# chat is the ordinary permanent state, and each costs a row, a ~24 MB process and about
+# seven of every switch's tmux calls. This plane opts in because its operator wants them.
+#
+# `workspaces` draws `switch.workspaces()` — EVERY directory under `workspaces/`, with the
+# current one marked. It is a listing, not a project switcher, and it stays one until open
+# and live are real states (the IDE spec, 4c). Placed here anyway at the operator's
+# request, with that limit stated rather than discovered.
+[[frame.component]]
+use  = "workspaces"
+edge = "top"
+size = 1
+bg   = "brightblack"
+pad  = 1
+
+[[frame.component]]
+use  = "chats"
+edge = "top"
+size = 1
+bg   = "brightblack"
+pad  = 1
+
 [[frame.component]]
 use = "attention"
 bg  = "brightblack"
--- a/charter/frame/chrome.py
+++ b/charter/frame/chrome.py
@@ -1,4 +1,15 @@
-"""The frame's paint: filling a row to the pane's edge, and highlighting one.
+"""The frame's paint: filling a row to the pane's edge, highlighting one, and deciding
+what a stranger's row may keep of it.
+
+**The vocabulary and the containment of that vocabulary are one module, deliberately.**
+:func:`recipes` is what a provider is handed so it can match the rest of the frame;
+:func:`contain_row` is what `registry.Registry.draw` runs over what the provider hands
+back. Written apart, the two were each asserted and disagreed from the day the recipes
+landed (#604) against an escaping that had shipped two days before them (#550): charter
+served ``ctx.chrome['heading']`` and then escaped it into the literal text
+``\\x1b[1mMetrics\\x1b[0m`` on the way to the pane (#707). :func:`served_params` is the one
+list both read, so a role added to :func:`_role_values` is served and admitted on the same
+commit or on neither.
 
 **Everything here takes a FINISHED row** — a string some `tui` node has already
 composed and clamped — and hands back a string that must not go back through `tui`.
@@ -37,7 +48,7 @@ import os
 import re
 from types import MappingProxyType
 
-from .. import tui
+from .. import contain, tui
 from . import pane
 
 #: Reverse video on, and everything off. `_OFF` is a FULL reset rather than SGR 27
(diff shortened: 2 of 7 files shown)


diazoxide:
Sweep response: the two survivors on `_escaped` are equivalent mutants by construction — a shortcut cannot change an output, so `collapse-ifexp` and `shift-boundary` on it can never be pinned. The line stays and the comment now carries the measurement that decides it: without the shortcut a 24-row pane of 300-character plain rows costs **750 us against 340 us**, which is *more* than the `contain.one_line` this function replaces (512 us). Removing it would buy a provider its colour by making every text-heavy pane slower than it was. `tui.sanitize` and `tui.width` carry the same shortcut against the same kind of scan.

PR #723, opened 2026-08-30
The provider seam refuses a size policy it cannot place, and records what `Component` costs (#708, #709)

### LINKED ISSUE #708: A foreign component can only ever be Fixed(n), and the change that made `size` real for built-ins left the provider branch coercing
`charter/instance.py:1855` coerces a foreign component's size to `Fixed(n)`. A provider cannot express `Content()` or `Fill()` — the two policies charter's own components use for everything that is not a fixed strip.

## Why now

#660, #662 and #702 made `size` real for **built-ins** — `_built_in_size` (`instance.py:1521-1596`) — across three PRs and two follow-up defects. **The provider branch was left coercing by the very change that looked like it would fix it.**

That matters more than the line count suggests: `repos` is the component whose whole design is `Content()`, and it is charter's own worked example of what a component looks like. A provider copying the shipped example cannot express the shipped example's size.

## The property, not the spelling

The refusal is presumably about trust — a `Fill()` from a stranger's distribution competes with charter's own arithmetic for rows, and `layout` supports exactly one variable-row slot by construction (`VARIABLE_ROW_SLOTS`). That is a real constraint and may well be the right answer.

**But then the refusal should say so.** Today it silently coerces, so a provider declaring `Content()` gets a frame that draws — just not the one it asked for, with nothing said. That is the shape #687 was: a value read and then quietly overridden, where the honest options are honour it or refuse it.

## Related

- #660 / #662 / #702 — `size` for built-ins, and the two defects that followed.
- #687 — a `size` charter read and then refused, dropping the whole arrangement in silence.
- Found while triaging #605 (closed).

### LINKED ISSUE #709: `charter-cp` is a hard install-time dependency of every provider, and the entry-point group does not make that coupling loose
`charter/frame/registry.py:387` is `isinstance(obj, Component)`. So every provider distribution must import `charter` at build and run time to construct the object charter will accept — there is no Protocol or duck-typed shape anywhere in the seam.

## Why the entry point does not soften it

The component seam is advertised as loose: a distribution declares an entry point, charter reads metadata without importing anything (`§4b`, and `instance.py`'s provider branch is careful to ask metadata rather than import). That looseness is real for *discovery* and absent for *construction*.

The consequence is a version coupling nothing states: a provider pinned against `charter-cp` 0.5x gets whatever `Component`'s dataclass fields are that week. This tree has changed that class three times this month — `events`, `on_event`, `children` — and each change is a silent compatibility edge for every provider built against the previous shape.

## Possibly a won't-fix, and worth recording either way

A structural protocol (`typing.Protocol` plus a runtime shape check) would let a provider construct something charter accepts without importing charter — at the cost of losing the validation `Component.__post_init__` performs, which is where four of this month's provider-facing guards live (`events` without `on_event`, `on_event` without `events`, the edge names, the size policies).

That trade may not be worth making. **But today it is not a trade anybody took — it is an accident of `isinstance`**, and it should be a decision recorded rather than a coupling rediscovered.

## Provenance

Found while triaging #605 (closed), which built a real provider distribution and hit this from the outside.

diazoxide:
Closes #708. Closes #709 — as a decision recorded, not a defect fixed; the reasoning is
below and in the code.

Stacked on #722 (`fix/707-provider-chrome-round-trip`), which is its base. Merge #722
first and this retargets to `main` on its own.

## #708 — the refusal is right, so charter now refuses

A component declares `Fixed(n)`, `Content()` or `Fill()`. For a component charter did not
write, only the first was ever real: the pane's height is the whole number of cells its
`[[frame.component]]` table gives it, resolved by `instance.component_tables` without
importing anything. The declared policy was validated by `Component.__post_init__`, kept on
the object, and **read by nothing** — `layout._derive` runs over charter's own built-ins
only, and `panel.py` measures the pane it was given. So a provider that copied charter's own
worked example (`repos` is `Content()`) drew at a height it never asked for, with nothing
said. That is #687's shape, and #687's verdict was honour it or refuse it.

Honouring it is not available, and the two reasons are different:

* **`Content()`** would mean charter importing the provider's module and calling its
  `render` to measure its content — on every command that resolves a config, `charter
  --version` included. That is precisely what binding a component by *name* exists to
  prevent (§4b), and `docs/frame.md` already says charter will not run a stranger's code to
  answer a geometry question.
* **`Fill()`** would mean a second pane taking the remainder, and the frame has exactly one
  by construction: `layout.slot_sizes` answers every member of `VARIABLE_ROW_SLOTS` with
  `repos_rows`, and `commands_frame._reassert_sizes` leaves that set unasserted so tmux's
  `resize-pane -y` — which moves one boundary — has one remainder to give the rows to.
  `frame/builtins.py` records the measurement of a second claimant: registered as a placed
  `Content()`, the sidebar's `changes` section was handed the repo table's height.

So `registry._placeable` refuses, and the standin pane names the component, the policy it
declared, and why there is nowhere for it to be read. The rest of the frame draws around it,
in the rectangle config asked for, so a machine with the wrong provider installed draws the
geometry a machine without it draws (#535).

**PLACED and not registered**, and that distinction is load-bearing: `Fill()` is the
*required* policy for exactly one child of a composite, and a part is drawn inside its
parent's pane rather than split for, so it never comes through `place`. There is a case for
that.

Charter's own components are unaffected — `builtins.build()` registers them, and `place`
returns early for an id already registered, so `repos`' `Content()` and `personas`' `Fill()`
never reach the guard.

## #709 — a won't-fix, with the trade stated where the line is

`isinstance(obj, Component)` means every provider distribution depends on `charter-cp` at
build time and at run time. That is true, it is a real coupling, and the issue is right that
it should be a decision rather than an accident of one line. It is now written at the line,
and in `docs/frame.md`:

Discovery is genuinely loose — `_scan` reads entry point metadata and imports nothing, so an
installed-but-unplaced provider costs a frame nothing. Construction is genuinely tight, and
that is the trade taken: `Component.__post_init__` is where this contract's provider-facing
refusals live — the closed id alphabet that reaches tmux, the edge, the size policy, `events`
without `on_event` and `on_event` without `events`. A structural protocol cannot perform any
of them. It would move each refusal from construction, where the message names a fixable
thing before a pane exists, to the moment the pane draws — which is §4g's own argument for
refusing a version mismatch at load rather than hoping. It buys a provider one line in a
`pyproject.toml` and costs every provider its validation.

`API_VERSION` is what makes the coupling safe rather than sharp: one integer declared on both
sides, refused rather than negotiated, so a provider built against a `Component` shape
charter has since moved does not load rather than half-working inside somebody's frame. That
is the answer to the version coupling the issue names, and it is why the class may change
without a shim band.

## Tests

`APlacedProvidersSizePolicyIsRefusedRatherThanCoerced` — nine cases, each asserting the
reason and not only the refusal: `Content()` and `Fill()` are separate cases so a message
that named the wrong one cannot pass; the pane says `[[frame.component]]` and "without
importing"; a `Fixed(12)` provider is the control that stops every other assertion passing
on a seam that refuses everybody; the rectangle and split order are unchanged and the rest of
the frame still draws; a composite's `Fill()` part is still registered; and the *declaration*
is what is checked, not what the committed table replaced it with — a guard placed after
`_rectangle` passes every other case and fails that one.

Three hand mutations: no refusal (4 errors + 1 failure), refuse everything (14 failures,
mostly pre-existing cases — the control is real), and check-after-`_rectangle` (3 errors).

### DIFF
--- a/charter.toml
+++ b/charter.toml
@@ -105,6 +105,31 @@ bg  = "brightblack"
 pad = 1
 key = "F5"
 
+# The two tab strips, between the identity row and the work area: what plane, then
+# which workspace, then which chat inside it. Split order is the geometry (#488).
+#
+# Neither ships placed (`instance.FRAME_SLOTS`) and that is deliberate — a plane with one
+# chat is the ordinary permanent state, and each costs a row, a ~24 MB process and about
+# seven of every switch's tmux calls. This plane opts in because its operator wants them.
+#
+# `workspaces` draws `switch.workspaces()` — EVERY directory under `workspaces/`, with the
+# current one marked. It is a listing, not a project switcher, and it stays one until open
+# and live are real states (the IDE spec, 4c). Placed here anyway at the operator's
+# request, with that limit stated rather than discovered.
+[[frame.component]]
+use  = "workspaces"
+edge = "top"
+size = 1
+bg   = "brightblack"
+pad  = 1
+
+[[frame.component]]
+use  = "chats"
+edge = "top"
+size = 1
+bg   = "brightblack"
+pad  = 1
+
 [[frame.component]]
 use = "attention"
 bg  = "brightblack"
--- a/charter/frame/chrome.py
+++ b/charter/frame/chrome.py
@@ -1,4 +1,15 @@
-"""The frame's paint: filling a row to the pane's edge, and highlighting one.
+"""The frame's paint: filling a row to the pane's edge, highlighting one, and deciding
+what a stranger's row may keep of it.
+
+**The vocabulary and the containment of that vocabulary are one module, deliberately.**
+:func:`recipes` is what a provider is handed so it can match the rest of the frame;
+:func:`contain_row` is what `registry.Registry.draw` runs over what the provider hands
+back. Written apart, the two were each asserted and disagreed from the day the recipes
+landed (#604) against an escaping that had shipped two days before them (#550): charter
+served ``ctx.chrome['heading']`` and then escaped it into the literal text
+``\\x1b[1mMetrics\\x1b[0m`` on the way to the pane (#707). :func:`served_params` is the one
+list both read, so a role added to :func:`_role_values` is served and admitted on the same
+commit or on neither.
 
 **Everything here takes a FINISHED row** — a string some `tui` node has already
 composed and clamped — and hands back a string that must not go back through `tui`.
@@ -37,7 +48,7 @@ import os
 import re
 from types import MappingProxyType
 
-from .. import tui
+from .. import contain, tui
 from . import pane
 
 #: Reverse video on, and everything off. `_OFF` is a FULL reset rather than SGR 27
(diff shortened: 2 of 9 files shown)


diazoxide:
Sweep response: the survivor was `contain.one_line(c.id)` in `_placeable`'s message, and the sweep was right — it is deleted. `_placeable` is handed a **constructed** `Component`, so `component.usable_id` has already held the id to `_ID_RE` (`[a-z][a-z0-9_]{0,31}` with at most one dot), which contains no character `one_line` would rewrite. Every other refusal in that module quotes a `cid` that arrived from a committed file or an entry point name and has been through nothing, which is why they keep theirs. Verified by construction: a newline, an escape, an uppercase letter, a space and a non-ASCII letter in an id are each refused by `Component.__post_init__` before this function can be reached.

Issue #735, opened 2026-08-30
A corrupt `gather.json` leaves the repo panel saying `gathering this workspace's repos…` forever, with no route to a re-gather


diazoxide:
The repo panel tells "not gathered yet" and "the cache is unreadable" apart nowhere, so a
corrupt `gather.json` reads as a gather that is still running — permanently.

## Reproduce

```
$ charter claude --workspace alpha    # 8 clones, table draws fine
$ echo 'not json {{{' > .charter/frame/alpha.1/gather.json
# next repaint (any hook, or `charter workspace todo …` from inside the frame)
```

```
────────────────────────────────────────
    ⋯ gathering this workspace's repos…
                                          <- eight blank rows
────────────────────────────────────────
7 todos · F2 palette
```

That message never changes. No gather is running and none will start: `docs/frame.md` is
explicit that "A panel never gathers on its own — it reads the cache or says it has none."

The same wording appears when the workspace directory disappears under a running frame
(`mv workspaces/alpha /tmp/`), where the accurate answer is "alpha is gone", not "still
looking".

## No route out, either

There is no palette row for a re-gather, and the command wants two flags an operator has
no reason to know:

```
$ charter frame-gather
usage: charter frame-gather [-h] --session SESSION --workspace WORKSPACE
charter frame-gather: error: the following arguments are required: --session, --workspace
$ charter frame-gather --session alpha.1 --workspace alpha    # fixes it instantly
```

Two things worth separating:

1. **The message.** `⋯ gathering…` is right for "the cache file does not exist yet". A file
   that exists and will not parse is a third state and wants its own line — one that names
   the remedy, the way `no clones in <ws> · charter clone <repo> -w <ws>` already does.
2. **The route.** `refresh the repo table` is a one-line palette action over a command that
   already exists, and the two flags can both be defaulted from the frame the palette is
   running in.

Issue #733, opened 2026-08-30
Switching one chat's workspace strands it: the palette refuses it, `frame-chat` refuses it, and tmux's own key reaches a frame-less pane


diazoxide:
Chat membership is read off `state.frame_workspace`, never off the id's prefix — which is
right for a renamed workspace (spec §3.2) and wrong for `F2 → workspace`, because that
moves one chat's workspace and leaves the other chats in the same tmux session behind.
The moved chat then has no route back and the chats it left behind cannot see it.

## Reproduce

```
$ charter claude --workspace alpha          # chat alpha.1, tmux session `alpha`
F2 → workspace → gamma                      # alpha.1 now belongs to gamma
<second terminal>
$ charter claude --workspace alpha          # chat alpha.2, SAME tmux session `alpha`
```

Both terminals are now looking at `alpha.2` (documented — one session, one current
window). `F2` in either says:

```
chat: alpha.2 — pick another   this workspace has one chat — open another with `charter <harness>` in this workspace, …
```

There are two chats. The window list says so:

```
$ tmux -L charter list-windows -t alpha -F '#{window_index} #{window_name}'
0 alpha.1
1 alpha.2
```

The doorway is refused, so there is no palette route. By hand:

```
$ charter frame-chat alpha.1
charter: no chat 'alpha.1' here — have: alpha.2
```

tmux's own window key does reach it, and lands on a pane with **no frame at all** — the
switch away tore alpha.1's panels down (correct, #668) and nothing will ever put them
back, because the only thing that splits them is the `frame-chat` that refuses.

So one `F2 → workspace` on a two-chat session strands a chat: unreachable by the palette,
refused by the command, and frame-less if you reach it with tmux.

## Two things tangled here

1. `alpha.1` is still the tmux window name, still `$CHARTER_SESSION_ID`, and still what the
   palette prints — while it belongs to `gamma`. The id says one thing and membership says
   another, and both are on screen at once.
2. Whatever the answer for (1), a chat that is a live window of the session you are in
   should not be unreachable from inside it. `chats.of_workspace` is the reading half; the
   palette's chat doorway and `cmd_chat` both ask it, so both inherit the gap.

Worth deciding explicitly whether `F2 → workspace` moves *the chat* or *the frame*. If a
chat is a window of a workspace's session, moving its workspace and leaving its window
where it is produces exactly the state above.

diazoxide:
[...]

Issue #731, opened 2026-08-30
A recycled chat id inherits the previous frame's workspace pointer and lock, so `--workspace alpha` can come up labelled `gamma`


diazoxide:
A chat id is `<workspace>.<n>` and is reallocated when a frame exits. The session pointer
`charter workspace use` writes is keyed by that same id and is **not** reaped with the
frame — so the next frame that gets the id inherits the previous one's workspace and its
lock, and outranks its own launcher.

## Reproduce

```
$ charter claude --workspace alpha      # top row: ⬢ alpha        (chat id alpha.1)
F2 → workspace → gamma                  # top row: ⬢ gamma
<exit the harness>

$ ls .charter/sessions/
alpha.1.lock  alpha.1.workspace         # both say 'gamma'; the frame dir is gone

$ charter claude --workspace alpha      # asked for alpha, explicitly
```

What comes up:

```
 ⬢ gamma  ◆ steward                                          charter 0.54.0
 …
▪ repos 8
  ├─ api                               main*
  ├─ auth                              main↑1
  ├─ billing                           main*
  ├─ gateway                           main
  ├─ ledger                            main
  …(+3 more, all clean)
────────────────────────────────────────
0 todos · F2 palette
```

The identity row says `gamma`, the todo count is `gamma`'s (0), and the repo table is
`alpha`'s eight clones — because the launcher gathered for `alpha` (its own resolution,
recorded in `.charter/frame/alpha.1/workspace` = `alpha`) while every panel that asks
`state.workspace_for` gets the stale session pointer instead. The repo pane's height is
`gamma`'s seven rows, which is why alpha's eight repos come out with `…(+3 more)`.

The lock is inherited too: the new session is locked to a workspace nobody in it selected.

## Cause

`docs/frame.md` states the resolution order — pin, then "what you chose **inside** this
frame", then "what the launch resolved". The second rung is `.charter/sessions/<fid>.*`,
and its whole meaning depends on `<fid>` naming *this* frame. `state.new_chat_id`
allocates the lowest free ordinal from `.charter/frame/`, so an id is free the moment its
frame directory is reaped — while `sessions/<id>.workspace` and `sessions/<id>.lock` are
left behind by `state.reap`.

Either reap the session pointer with the frame, or make the launcher's explicit
`--workspace` (which is the most explicit thing on the command line) win over a pointer it
did not write.

diazoxide:
[...]

PR #757, opened 2026-08-30
A workspace you already have open is focused rather than added to, and a chat keeps the session id a resume will need (§4e, §4k)

diazoxide:
Two stages of `docs/superpowers/specs/2026-08-30-charter-opens-like-an-ide.md`, both from
§6's "Ready now" list: **the durable session id (§4e, reduced form)** and **open-or-focus
(§4k)**.

**One PR, and here is the case for it.** They share no code, no file and no test: §4e is
`charter/frame/state.py` and a new test module; §4k is `charter/commands_frame.py`, `docs/frame.md`
and a second new test module. Neither reads the other. Two PRs would have been two
five-minute reviews of two things nobody has to hold in their head at once — but they are
also the two halves of one delivery decision, and the ordering hazard §4e states (below) is
the kind of note that gets lost when it is filed on its own. Read either half alone; the
diff does not interleave them.

---

## Stage A — the durable session id (§4e, reduced form)

Charter records the harness's own session id at `.charter/frame/<fid>/session`, written by
Claude Code's `statusLine` hook. `state.clear_shape` deletes it, with its reason at the
line: *"a gauge reading somebody else's 78% is worse than either."*

**Both requirements are right and one file was serving two purposes.** The usage history
the gauge reads is keyed by that id and lives outside the frame directory, so deleting the
mapping is the only thing suppressing a stale gauge today — that argument is about a number
drawn on screen and it stands. It is not an argument about the identifier, which is the only
thing charter records that could ever ask a harness for its conversation back.

So `record_harness_session` writes a durable sibling, `session.durable`, from the same value
on the same branch, and `clear_shape` does not list it. The two can only ever disagree by
the sibling being **absent** — never by holding a different id. `clear_shape` and the gauge
are untouched.

**Nothing reads it, and that is the stage.** Reopen (§6 stage 5) is what reads it, and the
id cannot be written retroactively for a chat whose `session` a launch has already cleared —
so it has to land first, alone.

### ⚠ The ordering hazard, restated so it does not get lost

Keeping the id is "no behaviour change" **only while nothing relaunches into an existing
chat directory**. The moment a reopen does, `session` comes back from this file and the
gauge starts reading a history belonging to a run that is over.

**The gauge's gate must ship before reopen, not with it.** §4e settled what that gate is:
`state.exit_code(fid) is None` — one `stat`, already written and already correct — and
*not* a liveness check, because `state.is_live` for a chat has no launcher pid and falls to
`pane_matches`, which is `False` for every panel, always. It is deliberately not in this PR:
a gate is a behaviour change and this is not.

One migration case, pinned rather than discovered: a chat whose harness was already running
when this charter arrives keeps a `session` written by the previous version, so the no-op
guard returns before the sibling is reached and the sibling first appears when that harness's
session id next changes. There is no back-fill and there should not be — reading `session`
as a stand-in would be reading exactly the file whose deletion this pair exists to survive.

---

## Stage B — open-or-focus (§4k)

`charter -w foo` meant "add a chat to `foo`" whatever `foo` was doing. Where somebody was
attached, that was destructive: both clients of a tmux session share one current window, so
the `select-window` that put the new launch on its new chat pulled the existing client onto
it, and `_drop_panels` then tore down the panels they were reading.

§2.10 says the drag has no small fix, and I re-measured it: on 3.7c and at the 3.2 floor
alike, tmux has no per-client current window inside one session, and the only non-dragging
mechanism is a session group, which would stop a workspace being a tmux session.

**So the reason to drag is removed rather than the drag fixed.** A workspace whose session
already has an attached client is **focused** — one more client on the session they are on,
looking at the window they are on. Nothing is started, no ordinal is claimed, no directory is
made, no window is selected. A workspace nobody is attached to opens a chat exactly as
before: there is nobody there to drag.

### §3.3, which is the whole reason this is not ten lines

One tmux server serves **every plane on this machine**, session names are bare workspace
names, and `default` is a name every plane has. A focus decided on a name would have made an
existing cross-plane collision into charter's *advertised* behaviour.

`_workspace_to_focus` therefore **starts on disk** — `chats.of_workspace(ws)` under this
plane's own `config.STATE_DIR` — takes the `%<pane>` id this plane's own launcher wrote
down, and asks tmux only for `#{session_id}` and `#{pane_id}`. #693's rule: ids the server
minted, never names from a namespace every plane shares. A plane that has never opened a
workspace has no directory for it, **makes no tmux call at all**, and can never focus
anybody.

The residual is stated in the code rather than hidden: pane ids restart at `%0` when a tmux
server does. Within one plane that is already closed by the reap `cmd_launch` runs
immediately above this. Across two planes it is not, because that liveness list is by chat
id and chat ids collide exactly as session names do (`new_chat_id` counts from 1 on each
plane's own disk, so `shared.1` is the id both planes mint first). Closing it needs a plane
marker on the window — a stage, not a line.

### Two things it deliberately does not do

* **A launch that names something to run still runs it.** Attaching answers "put me in
  `foo`"; it cannot answer "run *this* in `foo`". `charter frame -- <cmd>` and
  `charter claude --resume <id>` both open a chat and run what they named, attached client
  or not — a launcher that swallowed an argv the operator typed and attached instead would
  be wrong in the one direction this module refuses everywhere else, silently.
* **Adding a chat from inside the frame is unchanged.** A charter started in a frame's own
  pane has `$TMUX` set and takes `_launch_in_operator_tmux`, so `+ charter <harness> opens
  another` — what the chat bar tells you — still does exactly that.

§2.4's geometry race between two clients of one session is untouched, and §5 already says
so: this removes the reason to have two clients on different windows, not the cost of having
two.

---

## Verification

**None of this runs in CI** — there is no tmux there, so a green gate says nothing about
either stage (§2.12).

* **The two-plane acceptance test (§7).** `TwoPlanesOnOneMachine` stands up two plane roots
  against one real tmux server with one workspace name they share — *and* the same chat id,
  `shared.1`, which both planes mint first. The plane that opened it focuses it; the plane
  that did not focuses nothing, though its workspace name, its chat id and the live session's
  name all match. **Both wrong implementations were run against that test and both go red**:
  matching on the session name, and matching on the chat id (disk ∩ live). A single-plane
  test cannot fail for §3.3's reason at all.
* **A whole `cmd_launch`, by hand, on a real server with a real attached client**, on this
  branch and on `main`, on both tmux versions and identical on each:

  | | second launch, plane A | third launch, plane B |
  |---|---|---|
  | `main` | adds `shared.2`, **moves the attached client onto it** | adds a window to plane A's session |
  | this branch | `attach -t $0`, no window, client stays on `shared.1` | adds a window to plane A's session |

  (Plane B's column is the pre-existing cross-plane collision §3.3 names. It is unchanged —
  this PR neither fixes it nor makes it the advertised path.)
* **The tmux facts, measured on 3.7c and at the 3.2 floor, identically on both:**
  `list-panes -a` lists every pane on the server with its own session id; `list-clients -t $N`
  takes a session *id*, prints one line per attached client, exits 0 with nothing on stdout
  when none is attached and rc 1 for a session the server does not have; `attach -t $N`
  attaches; and **a second client attaching to a session does not move its current window** —
  which is the fact that makes focusing safe where adding was not. The drag itself is
  asserted as a control in the same class, so "the client did not move" cannot pass for a
  client that was never movable.
* **Both new modules on both versions**: 29/29 and 18/18 on tmux 3.7c, and the same at the
  3.2 floor with it first on `$PATH`.
* **Full suite**: 9048 tests, `OK (skipped=1)`. The first full run caught something worth naming:
  `test_plane_spawn_guard.NoCharterEscapesThroughTheExecFamily` refuses any `os.exec*` in
  the tree that is not written down with a reason, and the new module's `pty.fork` +
  `os.execvp("tmux", …)` was one. It is registered with its reason now, beside the five
  other real-tmux modules that fork a client the same way — it execs tmux, never charter,
  and `os._exit`s in its `finally`.
* **Deletion sweep, decomposed by hand and never a bare count.** Every line this branch adds
  that refuses, clamps, defers or falls back was deleted one at a time — 23 mutations, each
  run on its own, **zero survivors**. Each alternative arm has an input that reaches only it:
  the `list-clients` return code and the empty client list are two separate tests, and the
  field-count check is two (a short row raises where a long row merely lies, and only the
  second can be mistaken for a match). Two of the 22 are not deletions but the wrong DESIGN,
  because that is what these two stages are actually at risk of: putting `session.durable`
  in `clear_shape`'s list, and keeping it outside the frame directory (#731's shape). Both
  go red.

  Run with `python3 -B` and `PYTHONDONTWRITEBYTECODE=1` over a purged `__pycache__`, so no
  colour here comes from a stale `.pyc` served to a same-second restore.

  **The sweep found one real defect in my own gate and it is fixed rather than documented.**
  The "only a launch that asked for nothing but the workspace" condition was first written
  `if h is not None and not rest:` — and `h is not None` is unreachable there: an
  unregistered harness means `argv = rest`, and `if not argv:` above has already ended the
  launch with "nothing to run". It is `if not rest:` now, with the reason written at the
  line.

  **And CI's own sweep found two more, which are fixed rather than argued with.**

  * `clients.stdout.strip()` was the `strip`/`lstrip` equivalent mutant `_live_chats`'
    own docstring names — truthy for exactly the same strings, so no test could tell the
    two apart. The question here was only ever "is there a client at all", so it is
    `clients.stdout.split()` now: one question, asked once on tmux's own newline-separated
    output, with no half to mutate — and deleting it IS pinnable, because a server
    answering a bare newline then reads as one client with a blank name.
  * `_KEPT_SESSION_FILE = "session.durable"` survived because every test reads through the
    constant, so a rename moved writer and reader together. That is the right code and the
    wrong test coverage: the name is **durable state in a directory a much newer charter
    reopens**, so renaming it after this ships silently loses the id for every chat already
    recorded. It is pinned as bytes now, together with the fact that the write leaves those
    two files and no `.tmp` beside them.

## The three findings routed to me tonight — #731, #733, #735

**I am taking none of them into this PR**, and the reasoning is more useful than the verdict.

### #731 — a recycled chat id inherits the previous frame's workspace pointer and lock

**My design does not make this impossible, and saying it did would be conflating two
different session ids.** #731 is about the CHAT id (`<workspace>.<n>`, `$CHARTER_SESSION_ID`)
being recycled while `.charter/sessions/<fid>.workspace` and `.lock` outlive the frame
directory whose removal freed the ordinal. Stage A is about the HARNESS's own session id
(Claude Code's UUID), and it lives *inside* `.charter/frame/<fid>/`.

That difference is the whole answer, and it is now **asserted rather than left as a property
the design happens to have** — `ItDoesNotOUTLIVETheChatItself` reaps a chat, allocates the
ordinal again, and requires the durable id to be gone. `state.new_chat_id` frees an ordinal
by the absence of that directory and `reap` removes it whole, so "the id is free" and "the
durable id is gone" are one event, not two. There is no window in which a new chat is handed
an old one's harness session. The deletion sweep pins it from the other side: **moving the
durable file out of the frame directory — #731's exact shape — goes red on that test.**

So #731 is not made impossible by Stage A; it is made *contrastable* by it, and the contrast
says which of the two candidate fixes in the issue is right. **Reap the session pointer with
the frame.** The alternative offered there — "make an explicit `--workspace` outrank a
pointer it did not write" — fixes only launches that carry the flag; a bare `charter` (which
#715 has just made the ordinary way to open the frame) still inherits, and the identity row
still lies. It is a small, well-specified job in `state.reap` + `workspace.py`, and it wants
its own reproduction and its own tests — `reap`'s keep-rules carry ~60 tests and §6 already
calls that file "not a stage — a redesign".

**One disclosure: this PR adds one new caller of the file #731 is about.** `_focus_workspace`
calls `_pin_workspace`, which calls `workspace.set_active(ws, session_id=<chat>)` — so a
*picked* focus writes `sessions/<chat>.workspace` and `.lock`, exactly as every picked launch
already does. It cannot introduce a mismatch (it writes the workspace that chat is already
in) but it is one more occasion for the leak, and it should be counted when #731 is fixed.

### #733 — `F2 → workspace` strands a chat

**Different job, and I recommend refusing it here.** It is about `F2 → workspace` breaking
the invariant that a chat's workspace is its tmux session's name — `frame/switch.py`, the
palette's chat doorway and `cmd_chat`, none of which this PR touches.

**But it interacts with open-or-focus and that is worth having on the issue as evidence.**
`_workspace_to_focus` reads `chats.of_workspace`, so it inherits #733's membership gap. In
the stranded state the issue describes — `alpha.1` moved to `gamma`, still a window of tmux
session `alpha` — `charter -w gamma` now finds a live chat for `gamma` (that is `alpha.1`)
and attaches to the session holding it, whose current window is `alpha.2`. So the operator
asking for `gamma` lands looking at `alpha`'s chat. Before this PR they would have got a new
tmux session named `gamma` instead.

I deliberately did **not** guard that at my seam. The only guard available is "refuse unless
the chat's session is the one named for `ws`", which is a session-NAME comparison — the one
thing §3.3 forbids — and it would paper over #733 rather than fix it. It is a fourth
consequence of the same broken invariant and belongs on #733.

### #735 — a corrupt `gather.json` says "gathering…" forever

**Not mine.** It is a repo-panel message (`frame/slots.py`, `frame/gather.py`) plus a palette
row over an existing command (`frame/builtin_actions.py`). No overlap with either stage.

## Conflicts

Rebased onto `origin/main` at 67d9121 (#715, bare `charter` — §4a) with **no conflicts**;
bare `charter` rewrites argv to `charter claude` with an empty `rest`, so it focuses, which
is the IDE behaviour that stage was for. **#724** (`fix/714-reconcile-panels`) was read
before pushing: it works in `_draw_panels`/`_relayout`/`_split_panels` and adds
`_window_panels`/`_reconcile_panels`, all of it below `cmd_launch`'s early branching; a trial
merge is clean. Its test changes are concentrated in `test_frame_launcher.py`'s `_FakeTmux`,
which is why both new test modules here are new files rather than additions to it.

Nothing here touches a version file or `docs/news/0.54.0-*`. Two news entries, both
`version: unreleased`.

### DIFF
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -1547,6 +1547,119 @@ def _chat_being_left(socket: str, *, beside: str) -> str:
     return chat if chat != beside and _FRAME_ID_RE.fullmatch(chat) else ""
 
 
+#: What :func:`_workspace_to_focus` asks of every pane on a server, in one call.
+#:
+#: Two fields, both of them tmux's OWN ids and neither of them a name —
+#: :data:`_WINDOW_SEAT_FORMAT`'s rule one noun over, and here it is the whole correctness
+#: argument rather than a target-grammar convenience. One tmux server serves every plane
+#: on this machine (:data:`SOCKET`), session names are bare workspace names, and `default`
+#: is a name EVERY plane has, so `#{session_name}` cannot say whose session it is.
+#: `#{pane_id}` can: a pane id is minted by the server, and the only plane holding one is
+#: the plane whose launcher wrote it down (`state.record_harness_pane`).
+#:
+#: `\t` because neither can contain one: `#{session_id}` is `$<digits>` and `#{pane_id}`
+#: is `%<digits>`, both tmux's own and neither settable by anyone.
+_PANE_SEAT_FORMAT = "#{session_id}\t#{pane_id}"
+
+
+def _workspace_to_focus(socket: str, *, ws: str) -> tuple[str, str] | None:
+    """``(tmux session id, chat id)`` for THIS PLANE's live *ws* when somebody is looking
+    at it — otherwise ``None``, meaning "open a chat as you always did".
+
+    **§4k, and §3.3 is why it is spelled this way.** `charter -w foo` used to mean "add a
+    chat to `foo`" unconditionally, and §2.3 measured what that does to a workspace
+    somebody already has open: both clients of a tmux session share one current window, so
+    the `select-window` that puts THIS launch on its new chat drags the other client off
+    the chat they were reading, and `_drop_panels` then tears that chat's panels down.
+    §2.10 measured the alternative and there is not one — tmux has no per-client current
+    window inside a session on 3.7c or at the 3.2 floor, and the only mechanism that does
+    not drag is a session GROUP, which would stop a workspace being a tmux session.
+    So the drag is not fixed; the reason to drag is removed. A workspace somebody is
+    already in is FOCUSED — one more client on the session they are on, looking at the
+    window they are on — and a workspace nobody is in opens a chat exactly as before.
+
+    **Matched on this plane's own chat directories, never on a live session name, and
+    that inversion is §3.3's.** `cmd_launch` already asks `if session in live_sessions:`
+    and that question is unanswerable across planes: `_live_sessions` returns every
+    session on the machine, `default` is `DEFAULT_WORKSPACE_FALLBACK` and therefore a name
+    every plane has, and `config.STATE_DIR` is the only thing that is per-plane. Reading
+    the answer off a name would make an existing collision into the ADVERTISED behaviour —
+    `charter -w default` in one plane attaching to another plane's frame. So the question
+    starts on disk: `chats.of_workspace` is `.charter/frame/` under THIS plane's state
+    directory, and `state.harness_pane` is a `%N` this plane's own launcher wrote down.
+
+    **The residual, stated rather than hidden, and it is narrower than it first looks.**
+    A pane id is unique on a running server but restarts at `%0` when that server does, so
+    a `%3` recorded for a chat that is over can later name a different live pane. Within
+    ONE plane that is already closed by the caller: `cmd_launch` reaps immediately above
+    this, and a chat directory whose id tmux no longer reports live is removed — so the
+    only directories reaching here are ones the server still lists. Across two planes it
+    is not, because that liveness list is by CHAT ID and chat ids collide exactly as
+    session names do: `new_chat_id` counts from 1 on each plane's own disk, so `shared.1`
+    is the id both planes mint first. What it costs is one focus of a frame that is not
+    yours, on a machine running two planes, across a tmux server restart, for a workspace
+    both planes have, on a pane id that came back round. Closing it needs a plane marker
+    on the window — a second option written at launch and a schema both halves agree on —
+    which is a stage, not a line. What is closed today is the case §3.3 names: a plane
+    that has never opened `ws` has no directory for it, reaches no tmux call at all, and
+    can never focus anybody.
+
+    **Three answers, and the two tmux calls are asked in the order that makes most
+    launches pay for neither.** No chat directory for *ws* on this plane — the ordinary
+    first launch — returns before `list-panes`. No live pane among the ones this plane
+    recorded — every chat of *ws* is cold — returns before `list-clients`. And a live
+    session nobody is attached to returns ``None`` as well, deliberately: with no client
+    on it there is nobody to drag, so a launch there SHOULD add its chat and select it,
+    which is the behaviour that shipped and the one an operator reopening a detached
+    workspace wants.
+
+    Measured on tmux 3.7c and at the 3.2 floor, identically on both: `list-panes -a` lists
+    every pane on the server with its own session id; `list-clients -t $N` takes a session
+    ID as its target, prints one line per attached client, and exits 0 with nothing on
+    stdout when none is attached (rc 1 only for a session id the server does not have).
+
+    **Nothing off tmux is re-checked on the way out and that is deliberate**, unlike
+    `_chat_being_left`, which holds its answer to `_FRAME_ID_RE` because `@charter_chat`
+    is a user OPTION and its value is whatever somebody set. Both fields here are tmux's
+    own built-in ids; a shape check over them would be a guard no input can reach, which
+    is the shape this repository's deletion sweep exists to delete rather than document.
+    The chat id returned is this plane's own directory name and never came from tmux at
+    all. The recorded pane ids are not held to `tmuxctl.PANE_ID_RE` for the mirror-image
+    reason: they are only ever COMPARED against what tmux just said, never sent to it, so
+    a value that is not a pane id matches nothing and is refused by the comparison itself.
+    """
+    mine: dict[str, str] = {}
+    for chat in chats.of_workspace(ws):
+        pane = state.harness_pane(chat)
+        if pane is not None:
+            mine[pane] = chat
+    if not mine:
+        return None
+    out = tmuxctl.run("finding this plane's live chats in this workspace",
+                      tmuxctl.server_argv(socket, "list-panes", "-a", "-F",
+                                          _PANE_SEAT_FORMAT),
+                      timeout=5, report=False)
+    # No branch on the return code, for `_chat_being_left`'s reason: a listing that failed
+    # has an empty stdout and falls out below as "no seats and therefore no answer", which
+    # is the same sentence said once. Exactly two fields per row, so a server that answers
+    # something other than this format cannot have half a row read as a pane id.
+    seats = [line.split("\t") for line in out.stdout.splitlines()]
+    seat = next((s for s in seats if len(s) == 2 and s[1] in mine), None)
+    if seat is None:
+        return None
+    clients = tmuxctl.run("asking whether anybody is looking at that workspace",
+                          tmuxctl.server_argv(socket, "list-clients", "-t", seat[0],
+                                              "-F", "#{client_name}"),
+                          timeout=5, report=False)
+    # Both halves, and they are two different facts: a non-zero return is a server that
+    # would not answer (or a session it no longer has, between the two calls), and an
+    # empty answer is a session with no client on it. Either way nobody is being dragged,
+    # so either way this launch opens its own chat.
+    if clients.returncode != 0 or not clients.stdout.strip():
+        return None
+    return seat[0], mine[seat[1]]
+
+
 def _frame_is_live(socket: str, fid: str) -> bool:
     """Is the frame *fid* still running on *socket*? One question, one place — #408.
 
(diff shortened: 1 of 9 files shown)


diazoxide:
## Both sweep survivors: pinned, not explained away

Neither was an equivalent mutant. You were right about the second one, and the first turned
out to be a question the code should not have been asking at all. Fixed in 49ff53a.

### `state.py:613` — `"session.durable"`, the persisted key

**Your reading is exactly right and the fix is a test, not an argument.** Every test in the
module read the name through `state._KEPT_SESSION_FILE`, so a rename moved writer and reader
together and nothing went red. Reading through the constant is the *right code* — one name,
one place — and it is the *wrong coverage* for this particular constant, because the value is
not an implementation detail: it is a filename in `.charter/frame/<fid>/`, in directories a
much newer charter reopens (§5's "survive a schema change across a long-lived chat
directory"). Rename it after this ships and every already-recorded id is silently lost, with
a resume that quietly does not resume.

So it is pinned as bytes, with the reason written at the test:

```python
def test_the_name_on_disk_is_the_contract_and_not_an_implementation_detail(self):
    state.record_harness_session(FID, SID)
    self.assertEqual(sorted(p.name for p in _dir().iterdir()),
                     ["session", "session.durable"])
```

The exact-set assertion is deliberate and does a second job: it pins that the atomic write
leaves those two files and **no `session.durable.tmp` beside them**, so a half-finished write
can never be mistaken for state either.

Verified by hand, `python3 -B` with `PYTHONDONTWRITEBYTECODE=1` over a purged `__pycache__`:
respelling the constant to `"session_kept"` now goes **red** on that test alone.

### `commands_frame.py:1658` — `clients.stdout.strip()`

**Smaller than you feared, and the honest fix was to stop asking the question.** Nothing
downstream consumed that value — it was `if ... or not clients.stdout.strip():`, a pure
truthiness test on "did tmux name any client". No name, no key, no `\n` reaching a target. So
the trailing-newline failure mode you describe could not occur here.

But that is exactly what makes it unpinnable, and `_live_chats`' own docstring already names
this shape in this file: `s.strip()` and `s.lstrip()` are truthy for **exactly** the same
strings, so no input can tell the two apart. The repo's rule for that shape is not "document
the equivalent mutant" — it is *say it once, in a form both halves can be pinned in*.

It is now:

```python
if clients.returncode != 0 or not clients.stdout.split():
```

One question, asked once, on tmux's own newline-separated output: no non-whitespace token,
no client. There is no `strip`/`lstrip` pair left to mutate, and **deleting `.split()` is
pinnable and pinned** —
`test_a_client_list_that_is_only_whitespace_is_no_client` feeds a server that answers blank
lines, which without it reads as one client with a blank name and focuses a workspace nobody
is looking at. That is your failure mode, arriving through the door it could actually come
through.

### Sweep, re-run

23 decomposed mutations by hand, each run on its own, **zero survivors** — including the two
above and two that are the wrong *design* rather than a deletion: putting `session.durable`
into `clear_shape`'s list, and keeping it outside the frame directory (#731's shape). All run
with `python3 -B` / `PYTHONDONTWRITEBYTECODE=1` over a purged `__pycache__`, so no colour
here came from a stale `.pyc`.

### #733, #731, #735

Answered in full in the PR description, under **"The three findings routed to me tonight"** —
posted there before this comment, so it is where the next reader finds it. In short: **none of
the three is taken into this PR.**

* **#731** — my design does **not** make it impossible, and saying it did would conflate two
  different session ids. It is now *contrastable* instead, and asserted:
  `ItDoesNotOUTLIVETheChatItself` reaps a chat, allocates the ordinal again, and requires the
  durable id to be gone — because it lives inside the directory whose removal frees the
  ordinal, while `.charter/sessions/<fid>.*` does not. That contrast says which of #731's two
  candidate fixes is right: **reap the session pointer with the frame.** The other one ("let
  an explicit `--workspace` outrank the pointer") only fixes launches carrying the flag, and
  bare `charter` — which #715 has just made the ordinary way in — still inherits.
  One disclosure: `_focus_workspace` adds one more caller of `workspace.set_active`, so a
  *picked* focus writes that pointer too. It cannot create a mismatch, but count it when
  #731 is fixed.
* **#733** — a different job: `F2 → workspace` breaking the chat↔session invariant, in
  `frame/switch.py` / `cmd_chat` / the palette doorway, none of which this touches. **But it
  interacts**, and that belongs on the issue as evidence: in the stranded state, `charter -w
  gamma` now attaches to the session holding `alpha.1` — whose current window is `alpha.2` —
  so the operator asking for `gamma` lands looking at `alpha`. I deliberately did not guard
  it, because the only available guard is a session-*name* comparison, which is the one thing
  §3.3 forbids, and it would paper over #733 rather than fix it.
* **#735** — not mine: a repo-panel message plus a palette row over an existing command.

PR #759, opened 2026-08-30
The tab bar draws the tabs that fit, not only the one you are already on (#758)

### LINKED ISSUE #758: The `workspaces` bar is clickable and reaches nothing: 15 names need 274 columns, and every narrower width draws only the tab you are already on
#725 made both bars clickable. On the plane the report came from, the `workspaces` bar
still switches to nothing at any width an operator runs at.

**Measured** (`slots._bar` against this plane's fifteen workspaces, `here =
harness-wrapper`):

| width | row |
|---|---|
| 80 | `  workspaces  *harness-wrapper  +14` |
| 120 | `  workspaces  *harness-wrapper  +14` |
| 160 | `  workspaces  *harness-wrapper  +14` |
| 200 | `  workspaces  *harness-wrapper  +14` |
| 240 | `  workspaces  *harness-wrapper  +14` |
| **274** | every name, `*harness-wrapper` marked |

Rung 1 needs exactly **274 columns**. Below it the ladder falls to the rung that keeps
the marked name and counts the rest — whose only drawn tab is the one the frame is
already on, and `slots._Tabs.switch_to` correctly answers `None` for that. So the bar is
clickable, correct, and inert: the operator's original report ("trying to click on
workspace or chats to switch — seems not working?") survives the change that answered it.

`chats` is not affected on this plane — two chats called `harness-wrapper.1` and
`harness-wrapper.2` fit on one row from 47 columns — but it is the same ladder, and four
chats of a workspace called `relations-and-delegations` need 127 columns, which an
80-column terminal does not have.

**What is wrong is the rung, not the click.** The rung should draw the names that FIT
rather than collapsing to the marked one, with the names it leaves out counted.

diazoxide:
Closes #758.

#725 made both bars clickable and the report that produced it survived. Fifteen
workspaces need **274 columns** for rung 1, so every real width fell to the rung that
draws `*harness-wrapper  +14` — one tab, the one the frame is already on, which
`_Tabs.switch_to` correctly refuses. Clickable, correct, and reaching nothing.

That rung now draws the **page** the marked name falls on, with a count at each end.

## Measured, on this project's own fifteen workspaces

Rows painted by a real tmux, captured off the pane with `capture-pane`:

| width | before | after | tabs a click can reach |
|---|---|---|---|
| 120 | `workspaces  *harness-wrapper  +14` | `workspaces   authority-audit   autonomy   charter-update-skill   default   fleet  *harness-wrapper  +9` | 0 → **5** |
| 160 | `workspaces  *harness-wrapper  +14` | `… +  news-dispatch-guard   opencode-integration  +7` | 0 → **7** |
| 200 | `workspaces  *harness-wrapper  +14` | `… +  plane-shape   relations-and-delegations  +5` | 0 → **9** |

Rung 1 still needs 274 columns; nothing above the windowed rung changed.

## A page, not a window centred on the marked name

The cut is a pure function of the **names and the width** — which name is marked never
enters it — so switching to a tab that is on the row redraws the same row with only the
`*` moved.

A centred window slides one step per switch, which puts a **different name under the cell
the operator just pressed**. Measured on the same fifteen names at 160 columns: six of the
nine tabs a centred strip draws answer a second press at the identical column with a
second, different workspace. `_Tabs.switch_to`'s "the tab you are on is not news" cannot
refuse that, because the name at that column really did change — so a double-click would
switch twice, which is the one thing that rule exists to stop.

An **anchored** window (only move when the marked name would leave it) has the right
behaviour and has to be remembered. A panel is one long-lived process per slot, so an
anchor survives an ordinary repaint — but not `cmd_respawn`, not the respawn hook a dead
pane fires, and not a density change re-splitting the panels. The row would then depend on
history the operator cannot see. Pages get the same steadiness with nothing remembered.

## Two counts, not one

`+5  *harness-wrapper  …  +9`. A lone trailing `+14` beside a page that begins in the
middle of the list says the names on the row are the *first* fourteen, which is false. The
leading count is what makes the row say where in the list it sits — the readout the `n/N`
rung one step narrower has always given.

Neither count is a tab, at either end.

## The map still describes exactly what was drawn

`row()` takes the leading `+N` as its own argument rather than having the caller glue it
onto the body, because a map measured from the heading would hand every tab's cells to its
**left neighbour** — a real name and a plausible switch. Every rung still publishes,
including the rungs that draw no tab: a windowed row makes `_Viewport.blank`'s rule
sharper, since a stale map now answers with a name that is *elsewhere on the row* rather
than merely absent from it.

## Scope

`charter/frame/slots.py` only. `builtins.py`, `switch.py`, `chats.py` and
`commands_frame.py` are untouched — `TABS.publish` and `_Tabs.switch_to` keep their
signatures and their rules, so `_bar_events` needs to know nothing about pages.

`chats` gets this because there is one `_bar` for both bars, by that function's own stated
design ("so the two cannot degrade differently"). There is no second code path to add. It
reaches the rung later — two chats named `harness-wrapper.N` fit from 47 columns — but it
does reach it: four chats of `relations-and-delegations` need 127.

## Verification

- Full suite green (9082 tests).
- New real-tmux class `ARealClickOnAWINDOWEDWorkspaceBarMovesTheFrame`: fifteen real
  workspace names, a 120-column window, a real client on a real pty, a real
  `charter panel workspaces` holding the bar's pane, and a real detached
  `charter frame-switch` started from its handler. Clicking a neighbour's tab switches the
  frame; the page does not move under the pointer; the same click a second time switches
  nothing; neither count is a tab. Green on **tmux 3.7c and the 3.2 floor**.
- Every other case in that file clicks rung 1, which needs 274 columns — i.e. a bar no
  operator can produce. That is why the new class exists.

### DIFF
--- a/charter/frame/slots.py
+++ b/charter/frame/slots.py
@@ -2592,6 +2592,63 @@ def _tab_columns(start: int, drawn) -> dict:
     return cols
 
 
+def _page(fields: list[str], at: int, room: int) -> tuple[int, int]:
+    """The half-open run of *fields* the tab at index *at* is drawn with, inside *room*.
+
+    :func:`_bar`'s windowed rung. *fields* are the already-marked, already-contained tab
+    texts; the answer is a slice of them wide enough to draw, with room left for the two
+    counts that stand for what it leaves out.
+
+    **The pages do not depend on *at*, and that is the whole design.** The list is cut
+    into consecutive pages left to right — greedily, as many whole names as fit, then the
+    next page from where that one stopped — and this returns whichever page the marked tab
+    falls in. So the page is a pure function of the NAMES and the WIDTH, and switching to a
+    tab that is on the page redraws that page unchanged with only the `*` moved.
+
+    That property is what makes a windowed strip safe to click, and the alternative is
+    where it was measured. A window CENTRED on the marked tab moves every column each time
+    the operator switches, so the cell they just pressed holds a different name a moment
+    later: on this project's own fifteen workspaces at 160 columns, six of the nine drawn
+    tabs answer a second press at the identical column with a SECOND, different workspace.
+    `_Tabs.switch_to`'s "the tab you are on is not news" cannot catch that — the name at
+    that column really did change — so a double-click would switch twice, which is the one
+    thing that rule exists to stop. Pages cannot: a drawn tab is by construction on the
+    current page, so switching to it cannot turn the page.
+
+    **Anchoring to the previous window would buy the same steadiness and would have to be
+    remembered.** A panel is one long-lived process per slot, so state does survive an
+    ordinary repaint — but not `cmd_respawn`, not the respawn hook a dead pane fires, and
+    not a density change that re-splits the panels. The row would then depend on history
+    the operator cannot see, and two frames showing the same plane at the same width could
+    disagree. Pages need nothing remembered.
+
+    **Both counts are paid for before a name is, and the trailing one is sized for the
+    worst case.** The leading `+N` is exactly ``+start``, which is known before the page is
+    built. The trailing one is not: whether it is drawn at all depends on where the page
+    ends, which is what is being computed. Reserving the widest count the list can produce
+    breaks that circle for the price of at most a cell or two on the final page, where the
+    reserve is spent on a count that is never drawn. The alternative is an iteration whose
+    fixed point is not obvious, to save a column.
+
+    At least one field per page, always, so a name wider than the whole row cannot spin
+    this. Such a page overflows, and :func:`_bar` measures the row it composed and gives
+    the rung up rather than drawing it — the ladder's own rule, one rung down.
+    """
+    # The widest count either end can carry: every name but one left out. Measured with
+    # `tui.width` for this module's own reason, even though charter mints the digits.
+    tail = _BAR_GAP + tui.width(f"+{len(fields) - 1}")
+    start = 0
+    while True:
+        budget = room - tail - (_BAR_GAP + tui.width(f"+{start}") if start else 0)
+        used, stop = tui.width(fields[start]), start + 1
+        while stop < len(fields) and used + _BAR_GAP + tui.width(fields[stop]) <= budget:
+            used += _BAR_GAP + tui.width(fields[stop])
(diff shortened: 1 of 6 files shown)


diazoxide:
**Sweep verdict: `no survivors`** — and it took two rounds, which is the useful part of this note.

The first CI sweep on this branch reported **1 survivor**:

```
[17/17] SURVIVED  charter/frame/slots.py:2642 [collapse-ifexp]
        _BAR_GAP + tui.width(f"+{start}") if start else 0
        — is the `if` branch pinned, or is `0` always the answer?
```

Not an equivalent mutant, and the shape of the difference is what the tests were missing.
Without that reserve a wider room cuts a page one name longer; the leading count that page
then has to carry was never budgeted for; the composed row overflows; `_bar` measures it and
gives the entire rung up. So the ladder stops being **monotone** — with a 15-name list the row
is drawn at 31 columns, **gone at 42 through 45**, and back at 46. Every case here asked about
one width at a time, and none of them could see a gap between widths.

Pinned as the property rather than as the arithmetic: every name of the list, at every width
from 0 to 200, is drawn at every width above the one it first appears at. Randomised probe
over 3,230 (list, name) pairs found no non-monotone case in the fixed tree; the mutant dies on
the first one.

Re-swept on CI after the fix: **`no survivors`**.

I had hand-mutated eight other lines of this change (the trailing reserve, the page-walk
comparison, the composed-row fit check, both count conditionals, the map's starting column)
and all eight went red — so the survivor was the one edge my by-hand pass missed, in the one
direction I did not try. I mutated `if start else 0` to *always subtract* and it went red; the
sweep mutated it to *never subtract*, which is the direction that breaks monotonicity.

**The local sweep gave no verdict**, twice, and `no verdict` is not `no survivors`: on this
machine `tools/sweep.py` aborts at its own baseline — `Ran 0 tests — no tests ran in 122s`,
before any mutation — so it correctly refused to report. Not this branch: the same tree's full
suite passes directly (9082 tests), passes under the sweep's exact detached, piped,
`start_new_session=True` invocation, and passes on four Python versions on CI. It is a
sandbox-local harness failure and the CI sweep is the verdict of record.

Issue #728, opened 2026-08-30
A launch shows a blank terminal for ~1.8s and an empty skeleton for another 0.4s, with nothing saying it is starting


diazoxide:
`charter claude` shows a completely blank terminal for about 1.8 seconds, then an empty
skeleton of bordered boxes for another 0.4, before anything readable is on screen. Nothing
says a launch is in progress. #719 found `_screenshot` catching that state; an operator
looks at it on every launch.

## Measured

Three consecutive launches, 120x40, tmux 3.7c, warm caches, time from the launcher's first
byte:

```
run 1: first rule=1.76s  everything else=2.28s
run 2: first rule=1.87s  everything else=2.30s
run 3: first rule=1.77s  everything else=2.13s
```

("first rule" is the first horizontal pane border on screen — the empty-box state.
"everything else" is the identity row, attention row, repo table and sidebar, which all
land together.)

Between 0.09s and ~1.8s the terminal holds **one blank line**. A slower machine, a cold
gather, or a plane with more clones stretches that, and there is no upper bound the
operator can see.

## What is missing

Nothing at all is printed, and nothing can be: `docs/frame.md` already establishes that a
line written to the terminal microseconds before tmux takes the alternate screen is
unreadable (measured at 86 bytes ahead of the switch). So the place for a "starting…" is
inside the frame — the panes exist from the first `split-window`, and the frame draws its
rules a full 400ms before any panel has content in it. An empty box is charter's own paint
surface going unused.

At minimum: the harness pane could carry one line, or the attention strip could be the
first pane split and drawn rather than the last.

diazoxide:
Measured this while fixing #729 and #727, and the two remedies proposed above are both
aimed at a window in which charter has nothing on screen to write on. Posting the numbers
so the next attempt starts from them.

## Where the 1.8 s actually goes

The blank is not "the panes exist and are empty". **No client is attached during it.**
`cmd_launch` creates the session with `new-session -d` (detached, `commands_frame.py:3777`)
and does not `attach` until `commands_frame.py:4000` — after every `split-window`, every
`set-option` and every hook. So for the whole blank period there is no terminal showing the
frame at all, and a pane charter has drawn into is not visible because nothing is looking
at it.

Decomposed on this machine (macOS, warm caches, 8-clone plane), median of 5:

```
bare python3 start                      31 ms
+ import charter.cli                   497 ms   (charter's own import: 466 ms)
+ build_parser()                       632 ms   (parser build: 135 ms)
[...]

Issue #517, opened 2026-08-25
frame: switch workspace and persona from inside the TUI


diazoxide:
Requested by the operator: *"We need to have workspace and persona switcher options in charter TUI, some hotkeys."*

Today the frame has one binding — `F2`, opening `frame-menu` via tmux `display-menu`. Switching workspace or persona means leaving the frame and typing `charter ws use <name>` / `charter persona use <name>`.

## What this needs to settle

**Where the list comes from.** `charter workspace list` and `charter persona list` already answer it. A menu built from them is data charter already has — but both are read from the plane, so the same rule as everywhere else applies: **a workspace or persona name is a committed value** and must be contained before it reaches a tmux `display-menu` item. This is the exact class as `[frame] hotkey` (a newline in a committed value reached tmux config text and achieved code execution at launch) and #453 (a committed key injected YAML). `menu._ACTION_ID_RE` and `contain.one_line` exist for this.

**What a switch actually does to a running frame.** Changing the workspace changes what every panel should draw. `notify.plane_changed` already exists to tell panels the plane moved, and #411 established that panels follow `charter ws use` only through the `$CHARTER_SESSION_ID` collision — which #412 then made explicit by carrying identity on `-e`. A switcher must move the frame's *own* identity, not just write a pointer some panels may or may not read.

**Persona switching has a lock.** `workspace.set_active` locks the session to a workspace, and `charter persona use` has its own rules. A menu that silently fails against a lock is worse than no menu — if a switch is refused, the frame must say so.

**Which keys.** `[frame] hotkey` is configurable and validated by `instance._HOTKEY_RE` — that constant is the single guard between a committed config value and tmux config text, so any new binding must go through the same validation, and new config keys must be added to `FRAME_FIELDS` with the same type-check-and-degrade discipline.

## Suggested shape

Extend the existing `F2` menu rather than adding two more top-level bindings: a submenu for workspaces and one for personas, built from the live lists, each item an `frame-action` that performs the switch and then bumps the frame so every panel repaints against the new plane. That reuses the containment, the action-id validation and the repaint path that already exist, instead of opening three new doors.

PR #763, opened 2026-08-30
The frame stops freezing for four seconds to say it moved, and stops spinning after the work ends (#729, #727)

### LINKED ISSUE #727: The attention row keeps a frozen spinner and a stale `N running` after the work finishes, until something else bumps the frame
Found by driving a real frame. When in-flight work ends, the attention row keeps claiming
it is running and keeps the spinner frozen on whichever frame it last drew — which is
exactly what a hung frame looks like. It clears only when something unrelated bumps the
frame version.

## Reproduce

Inside a running frame, with the frame's own `CHARTER_ROOT`:

```python
from charter import inflight
tok = inflight.start("a-dispatch-that-will-be-killed")   # the row starts spinning
# ... then remove the record, which is what a finished or killed dispatch does:
for f in os.listdir(".charter/dispatch-inflight"): os.unlink(...)
```

Measured, reading the frame's bottom row back with `capture-pane`:

```
with a live record   : 7 todos · ⠏ 1 running · F2 palette
record removed at 01:35:11
  +2s  after removal : 7 todos · ⠏ 1 running · F2 palette  | dir: []
  +4s  after removal : 7 todos · ⠏ 1 running · F2 palette  | dir: []
  +6s  after removal : 7 todos · ⠏ 1 running · F2 palette  | dir: []
  +8s  after removal : 7 todos · ⠏ 1 running · F2 palette  | dir: []
  +10s after removal : 7 todos · ⠏ 1 running · F2 palette  | dir: []
  +12s after removal : 7 todos · ⠏ 1 running · F2 palette  | dir: []
```

Same spinner frame (`⠏`) throughout, an empty tracker directory, and a row saying one thing
is running. A `state.bump(fid)` from anywhere clears it instantly:

```
$ python -c "from charter.frame import state; state.bump('alpha.1')"
7 todos · F2 palette
```

## Cause

`frame/panel.py::_tick`:

```python
if resized["flag"] or now != seen or animating or handled:
    ...
    (paint or _paint)(slot, fid)
```

with `animating = animates and bool(_running(inflight_cache))` from `_watch`.

While a record exists the panel repaints each tick and draws `⠙ 1 running`. On the tick
after the last record clears, `_running` answers 0, `animating` goes false, the version has
not moved, nothing was resized and no event was handled — so **no paint happens at all**
and the screen keeps the last frame drawn, which is the one with the spinner on it. The
loop is missing the falling edge: it needs one more repaint when `animating` goes
true → false. `_watch` already has the previous value in hand
(`inflight_cache["running"]`).

## How often

Not every launch — the launch gather's own record can be short enough that the bottom panel
never draws a spinner at all (measured: 0.36s → 0.48s on a warm 8-clone plane, 6/6 clean
launches). It is every time the panel *does* catch it, which is every dispatch, every
`charter clone` and every `gl-refresh` that runs long enough to be seen — i.e. all the ones
worth animating. Observed unprompted on four separate launches during this audit, including
one that held `⠸ 1 running` for over fifteen minutes and survived a detach/reattach.

In a Claude Code session the row self-corrects at the next `posttooluse*` hook, so the lie
lasts until the agent's next tool call; in a `charter frame -- <shell>`, or while the agent
is waiting on a sub-agent (which is precisely when nothing bumps the version — see
`panel.py`'s own docstring on why the animation exists at all), it lasts indefinitely.

`docs/frame.md` says "The moment the last of it finishes, the frame goes completely still
again." It does go still — still showing a spinner and a count of work that has finished.

### LINKED ISSUE #729: Every switch freezes the whole frame for four seconds: `display-message -d 4000` suspends the client's pane redraw
Every workspace, persona and chat switch puts a `display-message -d 4000` on screen, and a
tmux client does not redraw its panes while a message is up. The panels have already
repainted into their panes; the operator cannot see it for four seconds.

## Measured

`F2` → `workspace` → `gamma`, tmux 3.7c, one attached client. Left column is the outer
terminal's screen, right column is `capture-pane -t <the identity pane>` at the same
instant:

```
 0.04  screen: (entirely blank)                 pane %1: '⬢ alpha  ◆ steward'
 0.23  screen: '⬢ alpha  ◆ steward'             pane %1: '⬢ gamma  ◆ steward'
       screen bottom: "charter: workspace → gamma  (lock moved from 'alpha')"
 4.21  screen: '⬢ gamma  ◆ steward'             pane %1: '⬢ gamma  ◆ steward'
```

So at 0.23s the frame is correct in every pane and the operator is looking at the previous
workspace, under a message that says they have moved. Nothing changes on screen until the
message expires at 4.2s.

A chat switch is worse, because the switch tears panels down and splits them fresh:

```
 0.05  screen: '⬢ alpha …'   bottom: 'charter: chat → alpha.1'
 0.70  screen: 'chatA$'      (frame torn down, harness holding the whole window)
 1.39  screen: (blank)
 3.96  screen: '⬢ gamma …'   bottom: '0 todos · F2 palette'
```

Four seconds of a blank or wrong screen for a switch `docs/frame.md` measures at "about a
third of a second".

## Cause

`commands_frame._say_on_screen` picks `-d 4000` deliberately ("long enough to read a
sentence"). The cost it does not account for is that the message suspends the client's
pane redraw for its whole duration. Any keypress dismisses it early, so an operator who
keeps typing never sees this; one who looks at the result of what they just chose always
does.

## What a fix wants

The outcome line and the frame's own repaint are competing for the same screen. Options
worth weighing: a much shorter `-d` for the success case (the frame itself is the
confirmation — the top row says `⬢ gamma`), keeping 4000ms only for refusals, which have
no other surface; or putting the outcome on a panel charter already draws rather than in
tmux's message area.

Related: the message is drawn over the terminal's last row, which is the attention strip —
the one row `docs/frame.md` promises is never dropped.

diazoxide:
Closes #729. Closes #727.

Two defects, one missing repaint edge. Both are the frame having entirely correct state and
failing to put it on screen at the moment it changed, so they share a fix rather than
merely a theme.

## #729 — a switch no longer costs four seconds of a screen you cannot see

Every workspace, persona and chat switch put its outcome up as `display-message -d 4000`.
**A tmux client does not redraw its panes while a message is up**, so the sentence
announcing the switch was also what hid it.

Measured with an outer terminal mirroring a real panel, `capture-pane` on both at the same
instant — the auditor's own method:

|                                   | origin/main | this branch |
|-----------------------------------|-------------|-------------|
| the pane had repainted at         | 0.52 s      | 0.33 s      |
| the operator's screen showed it at| 4.28 s      | 0.33 s      |
| **stale screen for**              | **3.76 s**  | **0.00 s**  |

At the tmux **3.2 floor**, same rig: **3.85 s** before, **0.00 s** after. The freeze tracks
`-d` linearly — `-d 200` froze for 0.20 s, `-d 750` for 0.74 s, `-d 4000` for 4.03 s — so
the four seconds bought nothing at all.

**Where the confirmation goes instead: the attention row.** `instance.FRAME_DENSITY` puts
`bottom` in every level charter ships (`minimal`, `normal`, `full`), so it is the one
surface besides the identity row that is always there — and it is the row `docs/frame.md`
already promises is never dropped, which is exactly the promise an outcome line needs. It
is a new highest-priority field on that row with a dwell: 4 s for a success (the frame
itself is the confirmation — the identity row already reads the new workspace), 10 s for a
refusal (nothing moved, so no panel repaints into the answer, and the text carries the
fix). "Just delete the message" was not the answer; #517 and #684 are still right that a
silent refusal is worse than a slow one.

### A second, independent reason, which the issue does not name

**`display-message -t <pane>` does not choose the screen.** `-t` is the target for FORMAT
evaluation; the client is `-c`, and with no `-c` tmux picks its own current client.
Measured on tmux 3.7c and at the 3.2 floor, two sessions on one server with a terminal
attached to each: a message aimed at a pane of session `sa` was drawn on the terminal
attached to `sb`, and not on `sa`'s at all.

charter had already been bitten by this by hand and read it as a property of an *empty*
target — `cmd_chat` carries a guard and a comment saying so. It is not. A well-formed `%N`
naming the right frame's own harness pane leaks identically, because the pane was never
what selected the screen. On a control plane with eleven frames on one socket, that is a
refusal about one frame drawn across another operator's.

A panel reads its own frame's state, so there is no direction for that to leak in — and
every client attached to the frame sees the row, which is strictly more than the one client
`-c` could name. Consequently the hotkey bind stops carrying `#{client_name}`: it was
threaded through a tmux bind, a CLI positional and a subprocess relaunch for that one
consumer. `charter frame-palette` still *accepts* the positional, because a bind installed
by an older charter is still sitting in a running server's key table across the upgrade.

## #727 — the spinner stops when the work does

`panel._watch` repainted while work was in flight and had no reason to repaint on the tick
after the last record cleared: nothing resized, the version had not moved, no event
arrived. So the pane kept the last thing drawn into it — a spinner frozen mid-turn over a
count of work that had finished, which is what a hung frame looks like.

Reproduced on a real client with the tracker directory empty:

|                              | origin/main      | this branch |
|------------------------------|------------------|-------------|
| row stopped claiming it runs | never (15 s+)    | **0.21 s**  |

on both tmux versions. `was_ticking` spends exactly one more paint on the falling edge —
and one more only, so `docs/frame.md`'s promise that the frame goes completely still stays
true.

## Why one PR

A notice expiring and a spinner stopping are the same question: *what this row would draw
now differs from what is on it, because the clock moved, and nothing will say so.* That is
why `_tick`'s third repaint reason is now `ticking` rather than `animating`, and why the
switch outcome could move onto that row at all — without the falling edge it would have
become a sentence that never went away, i.e. #727 again through #729's surface.

**#728 is deliberately not here.** Its 1.8 s blank happens before any client is attached,
so during it the frame has no state to show and no surface to show it on — a different
defect from these two. Measurements and two refutations of its proposed remedies are posted
on the issue.

## Also fixed, found while writing the tests

- A whitespace-only message wrote a live notice that drew nothing and still held the row's
  top priority (and `_watch` repainting) for its whole dwell.
- `float("nan")` parses, and every comparison against NaN is False — so a `now >= expiry`
  test read False and the notice became permanent. The comparison asks for the live case
  instead, which is the direction every other degenerate value here already falls.

## Verification

Full suite green. Every new guard mutation-checked by hand (`python3 -B`): reverting the
falling edge reddens 3 tests, the destination rule 1, and each of the two guards above 1.
Real-tmux numbers above measured on tmux 3.7c and on the 3.2 floor at
`~/.local/share/charter-testing/tmux-3.2`, each on its own socket, servers reaped.

### DIFF
--- a/charter/cli.py
+++ b/charter/cli.py
@@ -792,12 +792,15 @@ def _add_frame_parsers(sub) -> None:
     # `conf_text` — and, with `--pane`, by tmux's own `split-window`; never typed by an
     # operator.
     #
-    # Its `client` argument is `#{client_name}`, expanded by tmux INSIDE the bind's
-    # `run-shell` text before this process starts — never queried after the fact (see
-    # `cmd_palette`'s own docstring for why: `list-clients` cannot tell WHO pressed the
-    # key, only who is attached, and picking among several guessed wrong). `nargs="?"`
-    # because the pane half is started by charter itself with whatever the bind carried,
-    # and an empty client is a `display-message -t <session>` rather than a refusal.
+    # Its `client` positional is ACCEPTED AND IGNORED, and is kept for one reason only:
+    # a bind installed by a charter from before #729 is still sitting in a running
+    # server's key table across the upgrade, and it fires this command with a
+    # `#{client_name}` in it. Dropping the positional would make F2 on every such frame
+    # fail to parse. Nothing reads it: an outcome is now written to the frame's own state
+    # and drawn by its attention panel (`commands_frame._say_on_screen`), which every
+    # client attached to that frame sees, so there is no longer a per-client screen to
+    # choose between. `nargs="?"` is what lets the new bind, which carries no client at
+    # all, parse through the same parser.
     #
     # `--pane` says "you ARE the palette" rather than "open one". One subcommand and not
     # two, because they are two halves of one keypress and two spellings would be two
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -644,33 +644,35 @@ def conf_text(*, hotkey: str, mouse: bool, history_limit: int, session: str,
     a window created `-e CHARTER_SESSION_ID=chat-A` in it: the window's own PANE reported
     `chat-A` and a `run-shell` fired against that session reported `session-wide`. This
     bind's action IS a `run-shell`, so the variable would hand every chat's keypress the
-    same frame. The window option does not: like `#{client_name}` beside it, it is
-    expanded in the context of whoever's keypress is firing the bind, so `F2` in chat two
-    opens chat two's palette. `cmd_palette` falls back to `$CHARTER_SESSION_ID` when the
+    same frame. The window option does not: it is expanded in the context of whoever's
+    keypress is firing the bind — the same property `#{client_name}` was carried here for
+    until #729 removed the need for it — so `F2` in chat two opens chat two's palette. `cmd_palette` falls back to `$CHARTER_SESSION_ID` when the
     option is empty, which is what a frame launched by an older charter — its window
     carrying no such option, its bind text replaced by this one the moment a newer frame
     launches on the shared server — still resolves through.
 
-    `"#{client_name}"` is a THIRD thing this same bind carries, for a DIFFERENT
-    reason: which of possibly several clients attached to one frame should be TOLD when
-    an action refuses. Format expansion resolves `#{client_name}` in the context of
-    whoever's keypress is firing the bind — verified by hand with two real ptys attached
-    to one session, pressing the hotkey from each in turn: each press's own `run-shell`
-    resolved its OWN presser's client name, never the other one's, regardless of which
-    was attached first. `charter frame-palette` receives it as a plain argv value (`args
-    .client` in `cmd_palette`), carries it to the palette's own pane, and hands it to
-    `display-message -c`. Earlier, this module queried `list-clients` and guessed the
-    first one reported when several clients were attached — confirmed wrong, on the menu
-    this replaced: pressing the hotkey on the SECOND-attached client drew on the FIRST
-    client's screen, worse than tmux's own unscoped default single-client guess this
-    module was replacing. Carrying the presser by name removes the guess entirely rather
-    than making it a better one.
+    **This bind used to carry `"#{client_name}"` as a third value, and does not any
+    more** (#729). It named which of several clients attached to one frame should be TOLD
+    when an action refuses, and it was threaded from here through `args.client` and the
+    palette pane's own relaunch argv for exactly one consumer: `display-message -c`. That
+    consumer is gone. An outcome is now written to the frame's own state and drawn by its
+    attention panel (`_say_on_screen`), which is a PANE — so every client attached to the
+    frame sees it, and "which client pressed the key" stopped being a question that has
+    to be answered rather than being answered better. The value is not merely unused, it
+    is unnecessary: the surface it was selecting between no longer has variants.
 
-    **What the palette does NOT carry it for, and this is a real difference from the
-    menu.** A `display-menu` was drawn per client; the palette is a PANE (§4k), and a
+    `charter frame-palette` still ACCEPTS the positional (`cli.py`, `nargs="?"`), and that
+    is not vestigial — a bind installed by a charter that predates this change is still
+    sitting in a running server's key table across the upgrade and fires the command with
+    a client name in it, exactly as `--chat`'s own optionality is there for. Accepting and
+    ignoring it is what keeps `F2` working on those frames; emitting it is what stopped.
+
+    **The palette was never per-client anyway, and that is the difference from the menu it
+    replaced.** A `display-menu` was drawn per client; the palette is a PANE (§4k), and a
     pane belongs to the window, so with two clients attached to one frame both of them
     see it. That is the price of the surface being something charter draws rather than
-    something tmux draws, and it is the same price every other panel already pays.
+    something tmux draws, it is the same price every other panel already pays, and the
+    outcome line has now been moved onto the same footing as the surface that produces it.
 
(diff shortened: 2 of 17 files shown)


diazoxide:
## Verification

**Sweep: `no survivors`** (the verdict job's name). `gate: 0 unpinned, 0 in masked
cluster(s), 1 platform-deferred, 0 unresolved, 0 not applied, 0 withheld, 69 pinned`.

It took two rounds, and the first round's report is worth recording because one of the six
survivors was a test of mine that passed for the wrong reason: `_say_on_screen`'s dwell was
checked by saying a success, reading the expiry, saying a refusal and asserting the second
expiry was later — which **the clock advancing between the two calls satisfies on its
own**, so it passed against a version handing both outcomes the same dwell. That is why the
sweep flagged both branches of the `if ok else` at once. It now asserts the two durations
against the two constants. The other five: two unpinned `<`/`<=` boundaries, a `.strip()`
that read as `lstrip()` because the reader strips too and hid the difference,
`notice_expiry`'s containment guard having no test where `notice`'s twin did, and an
`except (OSError, ValueError, OverflowError)` whose last two clauses no caller can reach —
answered by deleting them rather than by testing them, which also puts `say` on the same
`except OSError` shape as `bump` and `record_exit`.

**Tests:** `test (3.11)` … `test (3.14)` all pass. Locally, where tmux is present, the
full suite is **9125 tests, OK, 1 skipped** — CI skips 9, so the real-tmux modules
(`test_frame_tmux_integration`, `test_frame_palette_integration`, `test_frame_surface_live`
— 151 tests) were run here rather than on the runner.

**Real tmux:** every number in the description was measured on tmux **3.7c** and on the
**3.2 floor** at `~/.local/share/charter-testing/tmux-3.2`, each trial on its own socket,
all servers reaped afterwards. The operator's live `charter` socket was never a target: the
before-rig asserts `state.frame_server(fid)` equals its own socket name before it lets the
old `display-message` path run at all.

Issue #752, opened 2026-08-30
A workspace removed or renamed under a running frame is drawn as an empty workspace, on the only surface still claiming it exists


diazoxide:
Remove a workspace's directory while a frame is drawing it and the frame keeps drawing it —
name, personas, an empty todo list, and the repo panel's `⋯ gathering this workspace's
repos…`. Nothing anywhere says the workspace is gone.

## Reproduce

```
$ charter claude --workspace alpha       # 8 clones, 7 todos, table drawn
$ mv workspaces/alpha /tmp/              # from another terminal
# next repaint
```

```
 ⬢ alpha  ◆ steward                                          charter 0.54.0
────────────────────────────────────────────────────────────┬────────────────
h$                                                          │▪ personas 6
                                                            │▸ steward
 …
────────────────────────────────────────────────────────────┴────────────────
    ⋯ gathering this workspace's repos…
────────────────────────────────────────────────────────────────────────────
0 todos · F2 palette
```

Meanwhile:

```
$ charter workspace list
  WORKSPACE  MODE   CLONES  REPOS
  beta       local  2       api, ledger
  delta      local  0       —
  gamma      local  1       search
```

The frame is the only surface on the machine still claiming `alpha` exists. The todo count
silently drops to 0, the sidebar's todo section disappears, and the repo panel reports work
in progress — three different ways of saying "empty" for a thing that is absent.

Putting the directory back does not recover it either: nothing re-gathers, so the panel
holds `⋯ gathering…` until `charter frame-gather --session <id> --workspace <name>` is
typed by hand.

## Why this case is worth naming separately

It is not exotic. `charter workspace remove`, `charter workspace rename`, a `git clean` on
the plane, and a teammate's `charter save`/pull all produce it, and a frame is by
definition long-lived enough to be running while one of them happens. `charter workspace
rename` is the interesting one: it "moves its clones + memory", so a renamed workspace
leaves every running frame on it pointing at a path that is gone, drawing the old name.

What is wanted is one honest line where the empty ones are now — the shape
`no clones in <ws> · charter clone <repo> -w <ws>` already has: name the workspace, say it
is not there, and name the command that fixes it (`F2 → workspace`).

Issue #371, opened 2026-08-22
The clone-commit nudge is 100% of this plane's ask volume and was approved 97 times out of 98


diazoxide:
`_clone_commit_reason` (`charter/hooks.py:951-967`) asks before a git write inside a workspace clone. It is not a safety rule and does not claim to be — the clone is its own repository, the plane's git is untouched either way, and `docs/hooks.md` describes it as a preference. This issue is about what it costs.

**It is the whole of charter's ask volume.** Across 5,765 traced events in this plane (`.charter/persona-state/trace/`, 2026-08-08 → 2026-08-22, test-bucket sessions excluded), **471 of 471 `ask` rows are this one rule.** No other nudge has ever fired here. `routing-ask` is 0, and `pretooluse_dispatch` cannot ask because no persona declares `dispatch-isolation: worktree` — so every `ask-approved` row is unambiguously an approval of this nudge.

**It is approved essentially every time.** `_ask_mark` / `posttooluse_bash` (`:123`, `:2145`) landed 2026-08-19 and the plugin caught up on 08-20, making 08-22 the first clean day:

| date | asks | approved |
| --- | --- | --- |
| 08-17 | 70 | (counter not live) |
| 08-18 | 116 | — |
| 08-19 | 87 | — |
| 08-20 | 37 | 5 |
| 08-21 | 47 | 24 |
| **08-22** | **98** | **97** |

Ninety-eight prompts in one day; ninety-seven approved. Against that, the persona tool-gate — the mechanism whose job is to *remove* prompts — has fired 16 times in two weeks and not since 08-18. charter's net effect on prompt volume in this plane is roughly −16 / +471.

**Nothing in the record shows it changing anyone's behaviour.** The nudge's message recommends moving to a repo-rooted session. Sessions that receive it keep receiving it, 98 times in a day. There is no `ask-declined` event — `posttooluse_bash`'s docstring notes that a declined ask leaves its marker behind and produces no row — so a decline is only ever inferred from an absence, and an absence is also what an interrupted turn or an ended session produces.

**It nudges against the workflow charter's own skill prescribes.** `skills/working-in-a-clone` says *"Commit to the repo you are in"*, and its own boundary section blesses the plane-rooted session for cross-repo work and light edits. The hook then asks whether that was intentional, on every git write, with no memory across a session.

---

### Supporting note: it also fires on commands that are not commits

Demonstrated in a lab fixture, not measured in the field — see the caveat below.

[...]

PR #764, opened 2026-08-30
Charter says which key it refused, and a frame-* command that did nothing stops reporting success (#738) (#734) (#747) (#744)

### LINKED ISSUE #734: Four `frame-*` commands exit 0 and print nothing outside a frame — including the two that are the only route inside your own tmux
`charter frame-toggle`, `frame-density`, `frame-chrome` and `frame-chat` all return 0 and
print nothing when `$CHARTER_SESSION_ID` is unset. There is no way to tell "it worked"
from "you are not in a frame".

## Reproduce

From any ordinary shell:

```
$ charter frame-chat alpha.1  ; echo $?
0
$ charter frame-density minimal ; echo $?
0
$ charter frame-toggle repos  ; echo $?
0
$ charter frame-chrome dark   ; echo $?
0
```

Four commands, zero bytes, four successes, nothing done.

## Why it matters more than it looks

`docs/frame.md` makes these commands the *only* route in two situations:

* **Inside a tmux you already have**, charter binds no key at all — "`charter frame-toggle
  <name>` typed in the frame's own window is the route there", and the same for
  `frame-density`. An operator who types it in the wrong window (the one they started
  from, say, rather than the one charter opened) gets silence and rc 0.
* **`charter frame-chat <chat id>` is the same switch by hand** — the documented fallback
  when the palette's doorway is refused (see the sibling issue about orphaned chats).

Inside a frame these commands do report: `charter frame-chat` for a chat that is not here
prints `charter: no chat 'alpha.1' here — have: alpha.2` through `_say_on_screen`. So the
machinery for saying something exists; the "no frame" branch is the one that returns
early, before it.

`cmd_chat`/`cmd_density`/`cmd_chrome`/`cmd_toggle` each open with

```python
fid = os.environ.get("CHARTER_SESSION_ID", "")
if not fid:
    return 0
```

One line on stderr — "not inside a frame; run this in the window `charter <harness>`
opened" — and a non-zero exit would close it. The zero is load-bearing for the tmux
`run-shell` path (a non-zero status makes tmux print into the harness pane), but that path
always has the variable set, so the guard can be honest.

### LINKED ISSUE #738: A refused `[[frame.component]]` arrangement is silent on every surface — launch, `frame-probe` and `doctor` all say nothing
`docs/frame.md`:

> **An arrangement charter cannot draw is refused whole, and your frame falls back to
> `slots`.** … You see your whole arrangement not take effect, which is something you can
> act on, rather than one pane's worth of quiet fiction.

There is nothing to see. A refused arrangement produces a frame identical to the default
one, and every surface that could say so says nothing.

## Reproduce

`charter.toml`:

```toml
[[frame.component]]
use = "identity"

[[frame.component]]
use = "attention"

[[frame.component]]
use = "repos"
bg  = "midnight"      # not one of the seventeen words

[[frame.component]]
use = "sidebar"
```

Then:

```
$ charter frame-probe
✓ charter frame: tmux 3.7 — a frame can run on this machine
$ echo $?
0

$ charter doctor | grep frame
  ✓  frame             tmux 3.7

$ charter claude --workspace alpha
# frame comes up, zero bytes printed, four panes, no bg anywhere
$ tmux -L charter show -p -t %3 -v window-style
                                    # empty — the whole arrangement was dropped
```

The same holds for every other refusal on that list: an unknown `use`, a duplicate, a
provider with no `edge`/`size`, a `pad` outside 0–5, a `size` charter cannot give a
built-in, a `key` it will not bind.

## Where it belongs

`charter frame-probe` is the command whose stated job is "can a frame run here, and what
will it not be able to do", and it already prints one line per standing limit — including
"any `[frame] slots` entry charter sizes but has no renderer for". A refused arrangement is
the same kind of fact and wants the same line, naming the key that did it:

```
! charter frame: [[frame.component]] ignored — `bg = "midnight"` on `repos` is not one of
  the seventeen colour words; the frame is drawing `slots` instead
```

`charter doctor`'s `frame` row wants it too. The launch path deliberately prints nothing
(the alternate-screen argument, which is right), so these two are the whole surface.

### LINKED ISSUE #744: The 3.2 resize limit says "until the frame is relaunched"; `charter frame-resize` fixes it in place and neither surface says so
Two surfaces tell an operator on tmux 3.2 that a resized frame stays broken until it is
relaunched. `charter frame-resize`, typed by hand inside the frame, fixes it completely.
Neither surface mentions the command.

## The claim

`charter frame-probe` on 3.2:

```
! charter frame: tmux 3.2 — a frame can run on this machine
  ↳ tmux 3.2 predates the `window-resized` hook (tmux 3.3+), which is what restores each
    panel's fixed size after the terminal is resized. Everything else in the frame works;
    resize this terminal and the panels stretch, and stay stretched until the frame is
    relaunched.
```

`docs/frame.md`, twice: "below that a resize still works, panels can just drift out of
shape until the frame is relaunched", and "what it costs is exactly one thing: resize your
terminal and the panels stretch, and stay stretched until the frame is relaunched".

## Measured, on `~/.local/share/charter-testing/tmux-3.2`

Launch at 120x40, drag to 80x24, drag back to 120x40:

```
$ tmux -L … list-panes -a -F '#{pane_id} #{pane_height}x#{pane_width}'
%1 5x120     <- identity strip, one row of text, five rows of pane
%0 22x97
%4 22x22
%3 5x120
%2 5x120     <- attention strip, same
```

Then, in the frame's own window:

```
$ charter frame-resize
$ tmux -L … list-panes -a -F '#{pane_id} #{pane_height}x#{pane_width}'
%1 1x120
%0 26x97
%4 26x22
%3 9x120
%2 1x120
```

Fully restored, instantly. So the standing limit is not "until relaunch", it is "until you
ask" — and asking is one command that already exists and is already documented as the
thing the hook calls.

## The second half: "everything else in the frame works" oversells it

The same drag to 80x24 on 3.2 gives:

```
────────────────────────────────────────────────────────────────┬──
h32$                                                            │▪…
                                                                │▸…
                                                                │▫…
────────────────────────────────────────────────────────────────┴──
  ⋯ too narrow for the repo table — 95 columns needed
────────────────────────────────────────────────────────────────
7 todos · F2 palette
```

The sidebar is squeezed to **two columns** of glyph stubs (and still costs three), and the
repo pane is a permanent `⋯ too narrow for the repo table` line — a message written to be
transient ("until that settles", 400ms) that on 3.2 never settles. "Panels stretch" reads
as cosmetic; this is not.

## Fix

Smallest version is documentation: both surfaces should name `charter frame-resize` as the
remedy, and the probe's line should say what a resize actually does to the panels rather
than "stretch". Worth considering beside it: a palette row for the same command, since on
3.2 that is the only recovery an operator has and `F2` is bound there like anywhere else.

### LINKED ISSUE #747: Nothing outside the frame documents F2, F12 or copy-mode scrollback — `charter claude --help` says nothing about driving the frame
Nothing charter prints outside the frame mentions `F2`, `F12`, the palette, or that
scrollback is now tmux's copy-mode. The only place any of it appears is the attention
strip's `F2 palette`, which is a two-word field on the bottom row that `density = minimal`
drops.

## What an operator gets

```
$ charter claude --help
usage: charter claude [-h] [--no-frame] [--probe] [--workspace WORKSPACE] [--pick] ...

positional arguments:
  rest                  Passed to the harness verbatim.

options:
  --no-frame            Run the harness bare, with no charter frame.
  --probe               Read-only: can a frame run here? …
  --workspace WORKSPACE Run in this workspace (skips the picker).
  --pick                Choose the workspace before the harness starts …
```

`charter frame --help` is the same text. Neither says what the frame *is* to drive.

`charter frame-probe` answers "can a frame run here" and names the standing limits, which
is its job and it does it well — but it is also the closest thing to "tell me about the
frame" and it says nothing about the keys.

`charter docs show frame` has all of it and nothing points there.

## The three facts a first launch needs

From driving it: `F2` opens everything; `F12` gets the keyboard back from a pane that has
stopped answering; scrollback is tmux's copy-mode, not your terminal's. The third is the
one `docs/frame.md` itself calls "the difference people notice first", and an operator who
hits it has no reason to connect it to charter at all.

## Cheapest fixes

* One epilog line on `charter <harness> --help` and `charter frame --help`:
  `Inside the frame: F2 opens the palette, F12 returns to the harness. charter docs show frame.`
* The same line on `charter frame-probe`, which already prints a paragraph per limit.

diazoxide:
Five issues filed tonight by a UX auditor driving a real frame, four of them fixed here.
They are one defect wearing four hats: **charter established a fact, acted on it, and
reported health.**

Closes #738, closes #734, closes #747, closes #744.

---

## #738 — a refused `[[frame.component]]` arrangement was silent on every surface

Reproduced as filed. Four tables, one `bg = "midnight"`:

```
$ charter frame-probe
✓ charter frame: tmux 3.7 — a frame can run on this machine     rc=0
$ charter doctor | grep -E 'frame|charter.toml'
  ✓  charter.toml   parsed cleanly (…)
  ✓  frame          tmux 3.7
```

The arrangement is refused whole — #535's rule, and it stays, for its own good reason —
so the frame that comes up is the `slots` one, byte for byte identical to the frame the
operator would have had if they had never written the tables. `docs/frame.md` promised
they would "see your whole arrangement not take effect". There was nothing to see.

`instance.component_arrangement` now answers `(placements, reason)` and each of the
fifteen refusals carries a sentence naming the table, the key and the value.
`component_tables` stays as the projection that drops the reason, so its twenty-odd
callers are untouched. `frame_of` records the reason as `components_refused`, which is
`harness_of`'s `refused` under a name that says which of `[frame]`'s keys it is about.

```
$ charter frame-probe
! charter frame: tmux 3.7 — a frame can run on this machine
  ↳ `[[frame.component]]` is not in force — `bg = "midnight"` on `repos` is not one of
    charter's 17 pane background words (`default`, the 8 ANSI colour names, and each of
    those with a `bright` prefix). An arrangement charter cannot draw is refused whole
    rather than one table at a time, so the frame you get is the one `[frame] slots`
    describes and every other table's `edge`, `size`, `bg`, `pad` and `key` goes with it.
    Fix that one key and the arrangement comes back.

$ charter doctor
  !  charter.toml   [[frame.component]] is refused; the frame is drawing `[frame] slots`
```

**`charter.toml` and not the `frame` row**, per the brief's steering and on its merits: it
is the same fact as the two silently-ignored settings already in that row —
`config.worktrees_root_for`'s `[plane] worktrees` and #715's `[harness] default`. A
committed key that is not in force and reads exactly like one nobody wrote. `check_frame`
answers "can a frame run on this MACHINE"; this is about the file.

### What is genuinely refused

Every path that already threw the arrangement away, and nothing else. Enumerated in
`ARefusedArrangementNamesTheKeyThatDidIt.test_every_refusal_on_the_documented_list_has_a_reason`:
an unknown `use`, a `use` that is not a string, a duplicate, a table that is not a table, a
misspelt key, a non-bool `visible`, a `pad` outside 0–5, a `size` charter cannot give that
component, an `edge` it cannot move a built-in to, a `key` it will not bind, a `key`
something already has, a provider with no `edge`, a provider with no `size`, a provider
this machine has no distribution for — plus one the issue did not list and TOML makes easy:
**single-bracket `[frame.component]`**, which parses as a table rather than an array of
them and produced the default frame in silence.

### And what is deliberately not

`_BRANCH_MOVERS` and #371 are this repo's record of a guard that fired too often being
deleted outright, so the reporting branch is gated on the KEY's presence and not on the
value being falsy: `component` absent → `None`, always. That covers every plane charter
ships. `RefusedIsNotTheSameAsUndeclared` pins it, including
`test_this_repositorys_own_plane_says_nothing`, which parses this repository's real
`charter.toml` — the test that would have caught a fix that warned on a working file.

## #734 — six commands, not four

Reproduced, and **the class is wider than the issue names**:

| command | typeable by hand? | before |
|---|---|---|
| `frame-chat` | yes — the documented palette fallback | rc 0, 0 bytes |
| `frame-density` | yes — the only route inside your own tmux | rc 0, 0 bytes |
| `frame-toggle` | yes — the only route inside your own tmux | rc 0, 0 bytes |
| `frame-chrome` | yes | rc 0, 0 bytes |
| **`frame-switch`** | **yes — and the only one of the six with a `help=`, so the only one `charter --help` shows you** | rc 0, 0 bytes |
| **`frame-resize`** | **yes, and #744 is what makes it so** | rc 0, 0 bytes |
| `frame-palette` / `frame-respawn` / `frame-gather` | no — a bind, a `pane-died` hook, a detached child | rc 0, 0 bytes — **unchanged** |

`commands_frame.outside_a_frame` gives the six one sentence on stderr and a non-zero
status. The three tmux drives are left exactly as they were.

**The exit code was measured before it was changed.** On tmux 3.7c, `run-shell` shows a
child's **stdout** and discards its **stderr** entirely, and prints `'<the whole command>'
returned 1` for any non-zero status — into the harness pane, the one rectangle ADR 0018
says charter never draws in. That is why the other refusals in these functions still
return 0 and always will. It does not reach this branch: a bind carries `--chat` expanded
from the presser's own window, the resize hook carries `--frame`, and a palette row's child
is handed `CHARTER_SESSION_ID` explicitly by `builtin_actions._spawn`. On every path tmux
drives, `fid` is set.

`cmd_toggle` is the interesting one. Its `if not fid` was **deleted** on purpose, by the
deletion sweep, as a guard nothing could pin — with it gone every observable was identical.
That argument was right about the mechanism and wrong about the operator: "every observable
was identical" was the bug. The guard is back with a consequence a mutation can reach, and
`test_outside_a_frame_it_says_so_and_moves_nothing` now goes red without it.

## #747 — nothing outside the frame said how to drive it

One epilog, set in `cli._wire`, so every registered harness plus the escape hatch gets it
and the two pages stay identical **by construction**:

```
Inside the frame:
  F2 opens the palette — every action the frame has is in there
  F12 takes the keyboard back from a pane that has stopped answering
  scrollback is tmux's copy-mode now, not your terminal's own
  charter docs show frame  —  all of it, and what `[frame]` configures.
```

The same three close every `charter frame-probe`, ceiling or no ceiling — it is the closest
thing charter has to "tell me about the frame", and it named every limit and no key at all.
The palette key is read from `config.FRAME`, so a plane that moved `[frame] hotkey` gets its
own key rather than a page telling it `F2`. That also answers the auditor's related note:
`charter docs show frame` now has something pointing at it, from both surfaces.

## #744 — measured at the floor, and the issue is right

`~/.local/share/charter-testing/tmux-3.2`, a real frame under a real pty, launched at
120x40, dragged to 80x24 and back:

```
%1 5x120   %0 22x97   %4 22x22   %3 5x120   %2 5x120     <- stretched, and staying
$ charter frame-resize
%1 1x120   %0 34x97   %4 34x22   %3 1x120   %2 1x120     <- the launch geometry, exactly
```

Identical to the issue's numbers. `cmd_resize` is not version-dependent; only the hook that
fires it is. So `tmuxctl.below_resize_hook_message` stops saying "until the frame is
relaunched" and names the command, and `docs/frame.md` does the same in both places.

The second half is right too. At 80x24 on 3.2 the sidebar is `%4 18x2` — **two columns**,
one truncated glyph per row — and the repo pane holds `⋯ too narrow for the repo table — 95
columns needed`, a line `frame/slots.py` writes to be transient and which here never settles
because nothing measures again. "Everything else in the frame works" is gone.

## #752 — not in this PR

Verified and reproduced exactly as filed: `mv workspaces/alpha` under a running frame
leaves the frame drawing `⬢ alpha`, `no clones in alpha · charter clone <repo> -w alpha`,
`no personas` and `0 todos` while `charter workspace list` lists only `beta` — three ways of
saying "empty" for a thing that is absent, on the only surface still claiming it exists.

The fix belongs in `charter/frame/slots.py`, which another agent holds for the windowed tab
strip. Left out rather than collided with; see the scope note below.

---

## Scope

`charter/commands_frame.py` and `charter/instance.py` are held by other clusters. The
changes here are confined to functions those clusters do not name — `frame_ready`,
`cmd_probe`, and the six commands' first guard in `commands_frame.py`;
`component_tables`/`frame_of` in `instance.py`. `doctor.py`, `cli.py`, `frame/tmuxctl.py`
and `docs/` were free.

## Verification

- **Full suite green locally**: `Ran 9344 tests … OK (skipped=1)`. CI green on 3.11, 3.12,
  3.13 and 3.14.
- **Deletion sweep on `d2ff872`: 45 mutations, 44 pinned, 0 survivors, 1 unresolved.**
  Shard 1 `no survivors` (23/23 pinned); shard 2 0 survived, 21 pinned, 1 unresolved. The
  unresolved one is `instance.py:1420 [retune-string] "hotkey"` — selected against 269
  modules and timed out rather than failing, which is #773's signature and is not a
  survivor. Measured locally instead: it is a `KeyError` inside `config.derive`, so it
  takes down `import charter.config` and every test that imports charter. Pinned many
  times over. An earlier run reported 2 survivors; both were real (not equivalent), and
  the comment thread below carries the measurement and the test that kills them.
- Every test in `tests/test_charter_says_what_it_would_not_do.py` fails without its fix,
  except the handful that are explicitly noise-guards and say so in their own docstrings.
- Real tmux 3.7c and the 3.2 floor, each on its own socket directory via `$TMUX_TMPDIR`.
  The operator's live `charter` socket was never touched.

### DIFF
--- a/charter/cli.py
+++ b/charter/cli.py
@@ -668,6 +668,37 @@ def _add_frame_parsers(sub) -> None:
     cases this docstring just said get different responses.
     """
     def _wire(parser, name):
+        # **How to drive the thing this command opens, on the command that opens it**
+        # (#747). `charter claude --help` and `charter frame --help` described four flags
+        # and never mentioned that the window they are about to hand the operator has a
+        # palette, an escape hatch, and somebody else's scrollback. The only place any of
+        # it was written outside the frame is `charter docs show frame`, and nothing
+        # pointed there; the only place inside it is the attention strip's two-word `F2
+        # palette` field, which `density = minimal` drops.
+        #
+        # Set here rather than at each `add_parser` for the reason this whole function is
+        # generated from `harness.all()`: one launcher per registered harness plus the
+        # escape hatch, and a per-parser epilog is one more thing to remember for a
+        # harness registered next year. It is also what makes `charter claude --help` and
+        # `charter frame --help` identical BY CONSTRUCTION, which is the property #747
+        # observed and wanted kept — they should say the same thing, and what was wrong
+        # was that the thing they both said was silent on the frame.
+        #
+        # Three facts and no more. They are the three an operator hits in the first
+        # minute: the palette is how you reach everything, the hatch is how you get the
+        # keyboard back from a pane that stopped answering, and scrollback is tmux's
+        # copy-mode now — which `docs/frame.md` itself calls "the difference people notice
+        # first", and the one an operator has no reason to connect to charter at all.
+        # `RawDescriptionHelpFormatter` because argparse reflows an epilog otherwise and
+        # a wrapped key name reads as prose. Called rather than read off a module constant:
+        # the sentence names this plane's own palette key, and a value frozen at import is
+        # a second answer to a question `config.FRAME` already answers — the masked-default
+        # shape the sweep is written to catch.
+        parser.formatter_class = argparse.RawDescriptionHelpFormatter
+        _keys = "\n".join(f"  {k}" for k in commands_frame.driving_keys())
+        parser.epilog = (f"Inside the frame:\n{_keys}\n"
+                         f"  charter docs show frame  —  all of it, and what `[frame]` "
+                         f"configures.")
         parser.add_argument("rest", nargs=argparse.REMAINDER,
                             help="Passed to the harness verbatim.")
         parser.add_argument("--no-frame", action="store_true",
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -373,6 +373,46 @@ def no_renderer_message(missing: list[str]) -> str:
             f"nothing there, so the harness pane keeps that space")
 
 
+def driving_keys() -> list[str]:
+    """The three facts a first launch needs, one per entry. Read by `cli._wire`'s epilog —
+    every `charter <harness> --help` and `charter frame --help` — and by
+    :func:`frame_ready`, so the launcher and the probe cannot come to disagree.
+
+    **Nothing charter printed outside the frame said any of this** (#747). `charter claude
+    --help` listed four flags; `charter frame --help` was the same text; `frame-probe`
+    named every standing limit and no key. Inside the frame it is one two-word field on
+    the attention strip — `F2 palette` — which `density = minimal` drops, and
+    `charter docs show frame` had the whole answer with nothing pointing at it.
+
+    **The keys are resolved, not spelled.** `[frame] hotkey` moves the palette off `F2`,
+    and a help line that says `F2` on a plane that moved it is worse than one that says
+    nothing: it is a fact an operator will act on. `config.FRAME` has already resolved the
+    committed value (or the shipped default, for a key charter refused) by the time any
+    parser is built, and `frame/overlay.py`'s `HATCH_KEY` is charter's own constant for
+    the hatch — the same object `conf_text` binds in both cases.
+
+    Scrollback is the third and it is not a key at all. `docs/frame.md` calls it "the
+    difference people notice first", and it is the one an operator has no reason to
+    connect to charter: their terminal's own scrollbar stops moving, and the frame is
+    simply what was on screen when it happened.
+
+    **A list rather than a sentence, and the two readers join it differently on purpose.**
+    A `--help` epilog is not reflowed (`RawDescriptionHelpFormatter`, which is what keeps
+    a key name off the end of a line) and wants one fact per row; `frame_ready` prints a
+    paragraph per limit and wants a sentence. What must not be duplicated is the FACTS,
+    and they are here once — a joiner at each surface is presentation, not a second
+    answer.
+    """
+    return [
+        f"{config.FRAME['hotkey']} opens the palette — every action the frame has is in "
+        f"there",
+        f"{overlay.HATCH_KEY} takes the keyboard back from a pane that has stopped "
+        f"answering",
+        "scrollback is tmux's copy-mode now, not your terminal's own",
+    ]
+
+
+
 def frame_ready() -> tuple[int, str, str]:
     """Can a frame run on this machine right now, and what will it not be able to do?
     ``(exit code, util.* level, text)`` — read-only: `tmuxctl.version()` and
@@ -385,7 +425,8 @@ def frame_ready() -> tuple[int, str, str]:
     launcher goes on to draw regardless — a probe that lies about `cmd_launch`'s own
     behaviour is worse than one that runs nothing at all.
 
-    **The three STANDING conditions are reported here and nowhere else.** All three used
+    **The four STANDING conditions are reported here and nowhere else.** The first three
+    used
     to be `util.warn` calls inside `cmd_launch` (or, for the resize hook, inside
     `_draw_panels`), and all three were measured to be unreadable there: `util.warn` for
     an unimplemented slot lands 86 bytes before tmux's own `\\x1b[?1049h`, so the
@@ -405,6 +446,17 @@ def frame_ready() -> tuple[int, str, str]:
     recovery at all — the gap being closed here. It does NOT change this function's exit
     code, for the same reason the other two do not: `cmd_launch` draws the frame
     regardless, and a probe stricter than the launcher lies about the launcher.
+
+    **A refused `[[frame.component]]` arrangement is the fourth (#738), and it is the one
+    that is about the plane rather than the machine.** It joins the list rather than
+    getting a surface of its own because it is the same KIND of fact as the unimplemented
+    slot beside it — a standing property of the committed file, true on every launch until
+    somebody edits it — and because it fails in the same way: the frame comes up, at rc 0,
+    looking exactly like the frame the operator would have had if they had never written
+    the tables. It does not change the exit code either. `doctor.check_control_plane_config`
+    is its second reader, where `[plane] worktrees` and `[harness] default` are already
+    named for being silently ignored; this is the surface an operator asks BEFORE they
+    launch, and the one `docs/frame.md` sends them to.
 
     Two callers share this, both read-only for the same reason `charter/news.py`
     requires of a `check:` (reads, never acts; and this module's own tmux calls all go
@@ -426,9 +478,29 @@ def frame_ready() -> tuple[int, str, str]:
     missing = frame_slots.unimplemented(config.FRAME["slots"])
     if missing:
         ceilings.append(no_renderer_message(missing))
+    # The fourth, and the only one that is about this PLANE rather than this machine —
+    # which is `no_renderer_message`'s position too, and the reason it belongs on the same
+    # list rather than in a report of its own. A `[[frame.component]]` arrangement charter
+    # will not draw is refused whole and falls back to `slots`, and the frame that comes
+    # up is indistinguishable from the one the operator would have got without the tables
+    # (#738). Read off `config.FRAME`, which resolved it once at import, so this stays what
+    # this function promises to be: `tmuxctl.version()` and `config.FRAME`, nothing
+    # started, nothing written.
+    refused = config.FRAME.get("components_refused")
+    if refused:
+        ceilings.append(instance.refused_arrangement_message(refused))
+    # **Said on every probe, ceiling or none** (#747). This is not a limit and it is not
+    # news; it is the answer to the question an operator is actually asking when they run
+    # the closest thing charter has to "tell me about the frame", and it was the one thing
+    # this command could not tell them. It rides BELOW the ceilings rather than joining
+    # them, for `_statusline_suppressed_note`'s reason one surface over: a ceiling is a
+    # capability this machine does not have, and a key that works is the opposite of one.
+    drive = "\n".join([f"  \u00b7 {k}" for k in driving_keys()]
+                      + ["  · charter docs show frame — all of it, and what `[frame]` "
+                         "configures."])
     if not ceilings:
-        return 0, "ok", head
-    return 0, "warn", "\n".join([head, *(f"  ↳ {c}" for c in ceilings)])
+        return 0, "ok", "\n".join([head, drive])
+    return 0, "warn", "\n".join([head, *(f"  ↳ {c}" for c in ceilings), drive])
 
 
 def _report_probe(code: int, level: str, line: str) -> int:
@@ -5188,7 +5260,14 @@ def cmd_resize(args) -> int:
     """
     fid = getattr(args, "frame", None) or os.environ.get("CHARTER_SESSION_ID", "")
     if not fid:
-        return 0
+        # **This command stopped being hook-only the day the 3.2 limit named it** (#744).
+        # Below `tmuxctl.RESIZE_HOOK_FLOOR` there is no `window-resized` hook to fire it,
+        # and typing it by hand is the operator's ONLY way to get the panels back — so
+        # `frame-probe` and `docs/frame.md` now tell them to, and a hand-typed recovery
+        # that answers rc 0 and nothing else is the worst possible reply to "did that
+        # work?". The hook path cannot reach here: `_resize_hook_argv` writes
+        # `--frame <fid>` into the action or arms no hook at all.
+        return outside_a_frame("charter frame-resize")
     harness_pane = state.harness_pane(fid) or ""
     if not _PANE_ID_RE.fullmatch(harness_pane):
         return 0
@@ -5309,8 +5388,14 @@ def cmd_density(args) -> int:
     settled would repaint into the old shape.
     """
     fid = os.environ.get("CHARTER_SESSION_ID", "")
+    if not fid:
+        return outside_a_frame("charter frame-density")
+    # Split off the empty *fid* above rather than sharing its `or`, and left a quiet
+    # no-op: a level outside the closed set is a DIFFERENT silence from #734's, on the
+    # `run-shell` side of the line `outside_a_frame`'s docstring draws, and it is not this
+    # change's to make. See that docstring, and the note in `cmd_toggle`.
     level = instance.density_level(getattr(args, "level", None))
-    if not fid or level is None:
+    if level is None:
         return 0
     where = _relayout_target(fid)
     if where is None:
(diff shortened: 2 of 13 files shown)


diazoxide:
## Sweep verdict on `9c1c9d1`, and what the three findings were

`2 survivors, 1 not measured` — 45 mutations, 42 pinned. All three are lines this branch
added; none is #776 cross-charging from `main`.

### Both survivors were real. Neither was equivalent.

`charter/commands_frame.py:5342` in `cmd_density` and `:5636` in `cmd_chrome`, both
`[uncontain] contain.readable(getattr(args, 'level', None))`.

The reason they survived is worth stating, because it is the shape that makes an
"equivalent mutant" claim tempting: **`state.say` already runs `contain.one_line` over the
assembled row**, so the newline half of both messages was covered before either function
called `contain.readable` at all. Every case I had written used an ordinary word.

Measured rather than reasoned about — the two functions differ on **invisible codepoints**,
and `contain.readable`'s own docstring says why: `one_line` decides on `_INVISIBLE`, five
categories, and U+3164 HANGUL FILLER and U+2800 BRAILLE PATTERN BLANK are on none of them,
are not `isspace`, and survive `strip`. Through this exact path, `charter frame-density`
with three U+3164:

```
with    contain.readable ->  charter: no density level ㅤㅤㅤ — have: minimal, normal, full
without it               ->  charter: no density level ㅤㅤㅤ — have: minimal, normal, full
```

The second draws as `no density level    — have: …` — a refusal naming **no level at all**,
which is #498's defect one surface over and the exact opposite of what this message exists
to do. `test_a_level_that_renders_as_nothing_is_still_named` covers both; each mutation
re-applied by hand goes red with it in place, and the unmutated tree stays green.

### The one not measured, and why

**Shard 2 of 2, mutation `[11/22]`**: `charter/instance.py:1420 [retune-string] "hotkey"` —
the `hotkey=out["hotkey"]` keyword on the `component_arrangement` call. The selection map
put **267 modules** behind it, and the run timed out rather than failing. That is #773's
signature and it is not a survivor: a timeout is neither a red nor a pin.

Measured locally instead, `python3 -B` with the mutation applied:

```
File "charter/config.py", line …, in derive
    d["FRAME"] = _instance.frame_of(cfg)
File "charter/instance.py", line 1420, in frame_of
    placed, refused = component_arrangement(section, hotkey=out["hotkeyX"])
KeyError: 'hotkeyX'
```

It is a `KeyError` inside `config.derive`, so it takes down `import charter.config` and with
it every test that imports charter at all. Pinned many times over; there is no test to
write. The 267-module selection is the sweep being *right* about reach and then running out
of budget to prove it.

diazoxide:
## The two questions this PR was asked

### 1. Was the silent-success class wider than the four commands #734 names?

**Yes, twice over — and the second widening is the one that would have been missed.**

I enumerated every `frame-*` command rather than trusting the issue's list. Nine exist. All
nine returned rc 0 with zero bytes when `$CHARTER_SESSION_ID` was unset:

| command | route | before | now |
|---|---|---|---|
| `frame-chat` | typed — the documented palette fallback | rc 0, 0 bytes | stderr + rc 1 |
| `frame-density` | typed — the only route inside your own tmux | rc 0, 0 bytes | stderr + rc 1 |
| `frame-toggle` | typed — the only route inside your own tmux | rc 0, 0 bytes | stderr + rc 1 |
| `frame-chrome` | typed | rc 0, 0 bytes | stderr + rc 1 |
| **`frame-switch`** | **typed — and the only one of them `charter --help` gives a described row to** | rc 0, 0 bytes | stderr + rc 1 |
| **`frame-resize`** | **typed — #744 is what makes it so** | rc 0, 0 bytes | stderr + rc 1 |
| `frame-palette` | a `bind -n` | rc 0, 0 bytes | unchanged |
| `frame-respawn` | a `pane-died` hook | rc 0, 0 bytes | unchanged |
| `frame-gather` | a detached child of the launcher | rc 0, 0 bytes | unchanged |

`frame-switch` is the fifth. It is the one an operator is *most* likely to type having never
been in a frame, because it is the one `charter --help` describes. `frame-resize` is the
sixth, and it only became typeable because #744 promotes it: below tmux 3.3 it is the
operator's only resize recovery, and both surfaces now tell them to run it. A fix covering
#734's four would have left the two most discoverable commands silent.

**The second widening is a different branch of the same defect: being in the right window
and typing a word charter has no meaning for.** `charter frame-density enormous`,
`frame-chrome solarized`, `frame-toggle reposs` — rc 0, zero bytes, nothing done. I left
these alone in the first pass on two reasons, and #729 removed both: the surface was
`display-message`, so saying anything put a typed argument through a tmux FORMAT evaluator
(and `cmd_toggle`'s guard is the one thing between an argv word and a `split-window`
target), and the message landed on whichever client tmux picked. The attention row is
charter's own pane, contained at the write and read by that frame's own panel, so both are
gone. Those three now answer on the row, at **rc 0** — the non-zero belongs only to the
branch that has no frame and no `run-shell` above it.

`frame-toggle`'s is the one with a second use: a component's `bind -n` outlives an edit to
`charter.toml` that drops the component, so this is the only thing that says why a key you
configured stopped doing anything.

The three tmux drives are deliberately untouched, and there is a regression test saying so:
a non-zero status inside `run-shell` makes tmux print `'<the whole command>' returned 1`
into the harness pane, measured on 3.7c.

### 2. What is *genuinely* refused for #738?

**Exactly the set that already threw the arrangement away, and the report is gated on the
KEY's presence rather than on the value.** `component` absent → `None`, always. That is the
whole `_BRANCH_MOVERS`/#371 exposure, and it is where the design decision sits: every plane
charter ships takes that branch.

Pinned three ways, because "it does not warn on correct configurations" is the claim most
worth distrusting:

- `test_a_plane_that_declared_no_arrangement_says_nothing` — no `[frame]` section, a
  `[frame]` with other keys, a density preset.
- `test_an_arrangement_charter_accepts_says_nothing` — a *good* three-table arrangement
  carrying `bg`, `key` and `pad`.
- `test_this_repositorys_own_plane_says_nothing` — parses this repository's real committed
  `charter.toml`. That is the one that would have caught a fix which warned on a working
  file.

What *is* refused is the fifteen paths that already returned a bare `None`, each now
carrying the table, the key and the value:
an unknown `use`; a `use` that is not a string; a duplicate; a table that is not a table; a
misspelt key; a non-bool `visible`; a `pad` outside 0–5; a `size` charter cannot give that
component; an `edge` it cannot move a built-in to; a `key` it will not bind; a `key`
something already has; a provider with no `edge`; a provider with no `size`; a provider this
machine has no distribution for — plus one the issue did not list and TOML makes easy:
**single-bracket `[frame.component]`**, which parses as a table rather than an array of
them and produced the default frame in silence.

Two judgement calls inside that set, both deliberate:

- **`component = []` is reported.** It is a key an operator wrote that decides nothing,
  which is the same class as every value refusal. Zero false-positive risk: it cannot arise
  from `[[frame.component]]` syntax at all.
- **The reason is `None` whenever the arrangement resolves**, even if it resolves to
  something odd. Charter does not warn about rules that do not exist; the trigger is
  "charter threw your tables away", never "your tables look unusual".

diazoxide:
## Final sweep verdict on `d2ff872`

Both survivors from the previous run are dead. 45 mutations, **44 pinned, 0 survivors, 1 unresolved**.

| shard | applied | pinned | survived | unresolved |
|---|---|---|---|---|
| 1 of 2 | 23 | 23 | 0 | 0 |
| 2 of 2 | 22 | 21 | 0 | 1 |

The one unresolved is the same `charter/instance.py:1420 [retune-string] "hotkey"` as before — now at **269 modules**, so it times out deterministically rather than intermittently. That is the sweep being right about reach and running out of budget to prove it, not a guard with no test behind it; the local measurement in the comment above shows it is a `KeyError` at `charter.config` import.

Checks: `test (3.11)`, `(3.12)`, `(3.13)`, `(3.14)`, `Size the sweep`, both shards, and `Add up what the shards found` all pass. Local full suite `Ran 9344 tests … OK (skipped=1)`.

Not merging — that is not mine to do.

Issue #742, opened 2026-08-30
The sidebar handles no events: its `…(+N more)` todos cannot be reached by mouse, keyboard or palette


diazoxide:
The sidebar draws two lists that look like lists you can act on, and it handles no events
at all. Its own `…(+N more)` line is the sharpest case: it advertises hidden content and
there is no route to it — not by mouse, not by keyboard, not through the palette.

## Reproduce

120x24, `[frame] mouse = true` (so pointer reports definitely arrive — with the shipped
`mouse = false` they only arrive when the harness happens to be asking, which makes this
worse, not better).

```
                                          │▪ personas 6
                                          │▸ steward
                                          │▫ docs ◦             ⚑
                                          │▫ forge ◦            ⚑
                                          │…
                                          │▪ todos 7
                                          │- Audit auth token ro…
                                          │- Cut 0.55.0
                                          │  …(+5 more)
```

* Click `forge` → nothing. The active persona does not change.
* Wheel over the todo list → nothing. `…(+5 more)` stays `…(+5 more)`.
* `F2` has no todo action of any kind.

The five hidden todos are reachable only by typing `charter workspace todo` into the
harness, which is the surface the frame exists to replace.

## Why the persona half matters too

`F2 → persona` is a real route, so persona switching is reachable — but the sidebar draws
the same roster as an inverted-row list with a cursor-looking marker on the active one,
which is the visual vocabulary of something you pick from. A click on it doing nothing is
the "tabs that do not respond to a click" shape.

## What a fix wants

The repo table already has both halves of the answer — it declares `click`/`scroll`, it
has `repo: select the next row` / `previous` palette actions for the keyboard-only plane,
and `docs/frame.md` states the rule that a pointer affordance always needs a key or a
palette route beside it. The sidebar wants the same treatment:

* `scroll` on the todo list, plus a palette action for the same movement;
* `click` on a persona row selecting it (selecting, not switching — the same
  "a click only ever selects" rule the repo table follows), or the roster drawn in a way
  that does not read as pickable.

diazoxide:
PR #765 fixes the **persona half** of this and deliberately leaves the **todo half** open, so this issue should stay open on that point.

[...]

Issue #634, opened 2026-08-28
With `[frame] mouse = true`, tmux moves the keyboard on a click and charter delivers the pointer anyway


diazoxide:
Charter now delivers `click` and `scroll` to components (#607). The delivery is point-to-act:
it acts where the pointer is and never moves focus, because the frame exists to keep the
harness the thing you type into.

That property holds with `[frame] mouse` off. **It does not hold with it on, and the reason
is tmux's, not charter's.**

## Measured

tmux 3.7c and the 3.2 floor, identical on both — real server, real client on a real pty, two
panes split `-h -l 40` in a 120x30 window, SGR reports injected as a reporting terminal sends
them:

```
tmux mouse off
  wheel over the panel   ->  panel receives it        active pane: the harness  (unchanged)
  click over the panel   ->  panel receives it        active pane: the harness  (unchanged)
tmux mouse ON
  wheel over the panel   ->  panel receives it        active pane: the harness  (unchanged)
  click over the panel   ->  panel receives it        active pane: THE PANEL    (moved)
```

tmux's default root binding is `MouseDown1Pane  select-pane -t = \; send-keys -M`. With its
own mouse on, tmux selects the pane under the pointer before forwarding the click.

## Why it matters more now than it did

Before this release nobody had a reason to click a panel, so the focus steal was a curiosity.
Now `[frame] mouse = true` is the setting an operator turns on *precisely because* they want
to click panels — and every click takes their keyboard off the harness until they put it
back (`F12`, or their own prefix keys). The setting that guarantees the feature works is the
setting that makes it hostile.

## What was deliberately not done in #607's pointer branch

Rebinding `MouseDown1Pane` to forward without selecting. A blanket
`bind -n MouseDown1Pane send -M` is wrong: it would also take away clicking **back** to a
pane, including the harness, which is worse than what it fixes.

The shape that looks right is conditional on the harness pane, using the same per-window
option pattern `overlay.HATCH_OPTION` already uses (a root key table is server-wide, so one
frame's pane ids must never be baked into the bind — `commands_frame.conf_text` records what
that cost the last time):

```
bind -n MouseDown1Pane if-shell -F -t= '#{==:#{pane_id},#{@charter_harness_pane}}' \
    'select-pane -t=; send -M' \
    'send -M'
```

It was left out of that branch deliberately: it changes tmux's documented `mouse on`
semantics for operators who already set it, it touches `conf_text`'s line-order invariants
[...]

Issue #530, opened 2026-08-25
frame: the top bar repeats the persona roster the sidebar already shows, except when the sidebar is dropped


diazoxide:
Reported by the operator: *"we still have personas in top-bar, why to have it if we already have personas list in right sidebar?"*

They are right that it is duplicated — but only sometimes, and the fix is a condition rather than a deletion.

## What is duplicated

`slots._top` renders, via `statusline._persona_line()`:

```
◆ steward · ◇ personas forge · reddit · release · statusline
```

Since #516, `slots._right` renders the same roster with strictly more information — a heading, memory badges, vault dots, health marks, in-flight badges, in an aligned column:

```
▪ personas 5
▸ steward          ✎47
▫ forge            ✎9
…
```

When both are on screen, the top bar's `◇ personas …` half says less about the same thing.

## Why it cannot simply be removed

Two callers, two contexts:

1. **`statusline.py:2325`** — the plain status line, used by every session NOT in a frame. There is no sidebar there. The roster is the only place personas appear.
2. **`frame/slots.py:264`** — `_top`, inside the frame.

And inside the frame it is still load-bearing some of the time: `layout.visible_slots` drops **`right` first, on ANY shortage** — `cols < min_cols or rows < min_rows` — before it drops anything else. So on a narrow or short terminal the sidebar is gone and the top bar is again the only roster.

The active persona (`◆ steward`) is a different thing from the roster: that is **identity**, and it belongs on the top bar next to the workspace regardless — the sidebar marks it with `▸`, but the top bar is where "who am I being" is read.

## The fix

`_top` should render the roster **only when `right` is not live**, and always keep `◆ <active>`.

`_top` receives `fid`, and the frame already records which panes exist (`state.panes` / the slot map `_draw_panels` returns), so "is `right` on screen right now" is answerable without guessing — and it must be answered live, because #387's density hotkey and `cmd_density` re-layout can add or drop `right` while the frame is running. A value read once at launch would be wrong the moment the operator changes density.

## The same shape as #516's badge work

`_persona_line()` returns **one flat string** with both halves already joined, so `_top` cannot keep one and drop the other. It needs the same treatment `_persona_chips` got in #516: expose the parts, and let each surface compose what it needs — rather than `_top` reassembling the facts, which is the drift `_right`'s docstring warns about.

Note `_persona_line`'s two branches already disagree about vocabulary in a related way (#513 fixed `agents` → `personas` in one of them), so this function is worth reading whole before changing it.

[...]

PR #765, opened 2026-08-30
The `F2 palette` hint is a button, a persona row is a row you pick, and its badges say what they mean (#742, #751, #753)

### LINKED ISSUE #751: The identity and attention strips are inert to a click, including the `F2 palette` hint, which is the only affordance the frame advertises
Of the four panels charter ships, only `repos` declares `click`/`scroll`. The two one-row
strips are inert, and the attention strip's last field is the word `palette` next to a key
name — which is a button everywhere else an operator has seen one.

## Reproduce

`[frame] mouse = true`, 120x24, so pointer reports definitely arrive.

```
7 todos · F2 palette · ▪ ledger · main · clean
```

* Click `F2 palette` → nothing. (`F2` itself works.)
* Click `▪ ledger · main · clean`, which is the readout of the repo row you just selected
  → nothing.
* Click `⬢ alpha` or `◆ steward` on the identity row → nothing, and both are doorways the
  palette opens by name.

The active pane does not move either, which is correct and documented — the complaint is
only that nothing at all happens.

## Why it is worth an issue rather than a shrug

The frame's stated goal is to read as an application. Three of the four things on the
identity and attention rows are nouns the palette already has an action for
(`workspace:`, `persona:`, and the palette itself), so the routes exist; what is missing is
the shortest one to them. And `docs/frame.md`'s own rule — anything a pointer can do needs
a key or a palette action beside it — is already satisfied for all three, so adding the
click adds no new obligation.

`F2 palette` is the sharpest case: it is the only affordance charter advertises on screen,
and it is the one thing on the row that cannot be operated the way it is drawn.

### LINKED ISSUE #753: The sidebar's persona badges (`◦`, `⚑`) are explained nowhere reachable from the frame, and an ordinary plane shows five flags
Every persona row in the sidebar draws two glyphs I could not find an explanation for
anywhere reachable from the frame:

```
▪ personas 6
▸ steward
▫ docs ◦             ⚑
▫ forge ◦            ⚑
▫ reddit ◦           ⚑
▫ release ◦          ⚑
▫ statusline ◦       ⚑
```

Five glyphs on that block — `▪`, `▸`, `▫`, `◦`, `⚑` — and the frame explains none of them.
`docs/frame.md` describes the column as "the memory, health and in-flight badges lined up
in a column of their own" and defines exactly one glyph in passing, in a paragraph about
colour: "`⚑`/`✗` on a persona whose charter is a draft or broken". `◦` is not mentioned
anywhere in the documentation set. `charter docs list` has no topic for it, `charter
doctor` does not decode it, and `F2` has no row that does.

## What I actually did

I ran a frame on a plane with six personas, saw five of them flagged, and had to read
`frame/slots.py` to learn what was wrong with them — where it turned out to be the
expected state for a persona created by `charter persona create` and never given a vault.
So the frame's default reading of a perfectly ordinary plane is five warning flags with no
way to find out what they warn about.

## Why the doc line is not the fix on its own

The glyphs are doing real work: `docs/frame.md`'s own rule is that "Every status in the
frame carries a glyph or a word that says the same thing as its colour", which is right —
but a glyph only says something to someone who has been told what it says. The status line
outside the frame had a shape for this (a `charter status` you could read in words); inside
the frame, the column *is* the whole answer.

Cheapest useful version: a palette row — `what the badges mean` — that opens the same
surface the picker uses over a short list. Second cheapest: name every glyph in
`docs/frame.md`'s sidebar paragraph, so `charter docs show frame` at least answers it.

diazoxide:
*"I clicked the one affordance the frame advertises and nothing happened. Then I read `frame/slots.py` to find out what `◦` meant."*

Three reports from one evening of driving a real frame with real SGR bytes through a real outer tmux, and they were one defect wearing three hats. Charter drew nouns the palette opens by name, drew a roster with a cursor-looking marker on the active row, drew the words `F2 palette` — **the only affordance it advertises on screen** — and handled no pointer event on any of them. Of the four panels charter ships, `repos` was the only one that had ever declared an event at all.

Closes #751, closes #753. **#742 is deliberately NOT auto-closed** — its persona half is
fixed here and its todo-overflow half is not; see *Known limits* below. Leave it open.

## What answers now

`identity` and `attention` declare `click` and publish a set of door **columns**; `sidebar` declares `click` and publishes a **row→persona** map with the active name beside it. Both new objects are `slots._Tabs` on another axis and keep its rules word for word: published by the pass that composed the row rather than re-derived when the click arrives; published on **every rung, including the rungs that draw nothing**; and the "you are already there" refusal living in the map object rather than in the handler, so it is a property of one class with one test.

| gesture | what happens |
|---|---|
| click `F2 palette` on the attention strip | opens this frame's palette — `charter frame-palette`, the same thing the hotkey bind runs |
| click `⬢ <workspace>` on the identity strip | the same palette |
| click `◆ <persona>` on the identity strip | the same palette |
| click a persona's **name** in the sidebar | `charter frame-switch --persona <name>` — the front door the `workspaces` bar already uses, one option over |
| click a persona's **badge column** | the glyph legend lands on the attention row for ten seconds, through #729's dwell |

## All three doors go to the same place, and that is the design

§4i's rule is that the irreversible half never rides on a pointer event. `_bar_events` got out from under it by acting on the **press** and switching to a tab that is still one click away — a gesture that completes on the pointer. **A strip has no such gesture available.** `⬢ alpha` names the workspace you are *on* and `◆ steward` the persona you *are*, so a click on either can only mean *let me pick another one*, and picking needs a chooser.

`key` is in `component.EVENT_KINDS` and deliberately **not** in `events.DELIVERED` — the harness owns the keyboard — so charter cannot grow a second, pointer-driven chooser inside a one-row pane. The palette *is* the chooser, both nouns are already reachable through it, and opening it completes on the pointer alone: `overlay.modal_argvs` makes it the active pane, so the keyboard reaches the rows it just drew.

Because the destination does not vary, `_Doors` holds a **set of columns** rather than a column→name mapping. A mapping whose values nothing branches on is data no input could make observable — the survivor the sweep reports.

The persona column is the other side: a click **switches**, for `_bar_events`' three reasons unchanged (press not release; the row you left is one click away; nothing could finish a two-step gesture). That is where this departs from #742's own suggestion of "selecting, not switching" — the table's rule is not *a click never switches*, it is *the irreversible half is never driven by an event that can arrive unpaired*, and what makes selection right on the table is that it has a real intermediate state and `Enter` to choose from it. A persona column has neither; the row in reverse video **is** the selection.

## What deliberately answers nothing, each with a case

- **The charter version and the context gauge** on `top`; **the todo count, the alert, the in-flight spinner and the news** on `bottom` — readouts. The frame reporting, not offering.
- **The selected repo's `▪ ledger · main · clean`** — the sharpest of them: it is the readout of a row selected on *another pane*, so a click could only mean "select what is already selected", which `_Tabs.switch_to` and `_repos_events` both already refuse.
- **The roster half of the identity row** (`◇ personas 6`) — drawn only when the sidebar is not (#530), so a door there would exist on some frames and not others for a reason no operator can see.
- **The ` · ` between two fields, and the space past the last one.** Charter will not pick the nearest field.
- **The `▪ personas N` heading, the `…(+N more)` row** (it stands for personas *not* drawn — `_Tabs` refuses its `+14` identically), **the `no personas` line, the blank rows between sections, every todo and change row**, and **the persona you already are** (also what stops a double-click switching twice).

## #753: every glyph has a name

**#753 asked for a legend reachable *from the frame*, and it gets one — because #763 built the surface three days after that issue was filed.** Clicking the badge column puts `slots.BADGE_LEGEND` on the attention row through `state.say`: the same dwell a workspace switch uses to say what it did, so the answer costs no extra row, clears itself after ten seconds, and needed no second surface. It works on the row you are already on, which is where the question usually comes from — the *name* refuses that click (re-adopting yourself is a no-op) and the badges do not, and that asymmetry is argued at `_Chips.hit`.

It is **one legend for every row rather than a reading of the row clicked**, and that is a stated limit: `PersonaChip.badges` is a rendered string by the time `slots.py` sees it, and a handler is handed no `ctx` (§4f) — so a per-row answer would need either decoding glyphs back out of display text or a second reading of the plane, and the question an operator actually has is *what is that glyph*.

`docs/frame.md` also names all ten glyphs in a table, so `charter docs show frame` is the long version. The two that sent the auditor into the source:

- **`◦`** — a declared vault charter **cannot use here** (no such vault on this machine, or its file does not exist yet). `charter persona create` writes the declaration and nothing writes the vault, so this is the ordinary state of a persona nobody has given credentials to. **Not a warning.**
- **`⚑`** — the charter is a **draft**, so no sub-agent is generated and it cannot be dispatched.

Six personas, five flags, nothing wrong — written down, because that reading is what cost an operator a trip through `slots.py`.

## The hover measurement, which the brief asked for either way

The operator asked for hover ("on hover somehow highlight text… tooltip on the footer"). **Measured before designing anything**, on real tmux 3.7c **and** the 3.2 floor, real client on a real pty, two panes, SGR reports injected as a reporting terminal sends them:

```
mouse=off:  outer terminal told 1000h 1006h   (the ACTIVE pane's request only)
            panel asking 1003 -> tmux does NOT propagate it. no motion is ever sent.

mouse=on:   outer terminal told ... 1003h     (BECAUSE the non-active panel asked)
            motion over the PANEL   -> panel READ \x1b[<35;11;5M  (pane-relative, 1-based)
            motion over the HARNESS -> nobody
```

Byte-identical on both versions. Three findings:

1. **Hover IS deliverable** to a non-active `charter panel` process, conditional on `[frame] mouse = true`. With mouse off the panel's 1003 request is never propagated — only the active pane's is — so no motion is ever sent.
2. **It cannot poison the harness.** tmux filters per-pane by what each pane asked for: with 1003 on for the whole terminal, motion over the harness (which asks 1000/1002) reaches **nobody**.
3. **The brief's stated cost is wrong in one respect and right in another.** It is *not* "a wake-up for every panel process" — only the pane under the pointer that asked for 1003 wakes. What is genuinely expensive is the operator's own preferred *shape*: a tooltip on the **footer** means the sidebar telling the attention pane what the pointer is over, and the only cross-pane channel is a state write plus `state.bump` — a file write and a repaint of every other panel **per motion event**, ~40 for one sweep of a 40-column pane.

The measurement, and what building it would take, are filed as **#766** so they outlive this PR.

**#766 now carries the full cost measurement too** — every cell is a wire event on both versions (40 in, 40 resolved), but the pty coalesces those into 18 `read()`s, and a panel that did not declare hover sees **zero**. The expensive shape is the footer tooltip, not 1003: a cross-pane answer is a `state.bump` per motion, which is ~160 panel repaints for one sweep, against 6 local repaints for an in-pane highlight. My judgement, on the issue: not yet, and not first — this PR's badge click removes the reason hover was urgent.

**So hover is not built here, and that is a scope decision rather than a refusal.** It needs a new kind in `component.EVENT_KINDS` (a closed vocabulary, §4f), a place in `events.DELIVERED`, a motion branch in `overlay.decode`, a `MOUSE_ON` variant, per-component opt-in and a debounce on the resolved target rather than on the event. That is its own change. The affordable shape, when it is written, is **highlight in the same pane** (repaint only when the resolved row changes), not a footer tooltip.

## Known limits, stated rather than implied

- **The sidebar's todo overflow is still unreachable** — #742's second half. Every cheap route was worse than the honest `…(+N more)`: a wheel with no keyboard twin; a palette noun that is not switchable; a state-backed offset that bumps the whole frame on every notch (which `_Viewport.move` refuses in as many words). It wants a read-only list surface, which is its own change. **#742 should stay open on that point.**
- **`personas` placed on its own is drawn and inert.** `registry.Registry` refuses a composite *part* that declares events (a part is never placed, so the declaration would build a handler that received nothing, ever) — so the `sidebar` composite declares, and a plane that places the bare component through `[[frame.component]]` gets no click. Lifting that means letting a composite delegate a kind to one part, which is a change to the registry's contract.
- **No client name on the palette argv.** `conf_text` expands `#{client_name}` because tmux knows who pressed; a panel has no presser, so `_say_on_screen` falls back to the most recently attached client. Same screen on a one-client frame; on a two-client one a refusal can land on the other operator's status line.

## Verification

- `python3 -m unittest discover -s tests` green.
- **Real tmux, 3.7c and the 3.2 floor, run twice each**: `tests/test_a_real_click_opens_the_real_palette.py` — a real client on a real pty, `mouse = true` sourced from `conf_text` (so #634's `MouseDown1Pane` rebind is under test), real `charter panel bottom` / `charter panel right` processes, and a real detached `charter frame-palette` / `charter frame-switch --persona` started by those panels out of their own handlers. The palette pane appears and takes the keyboard; `▸` moves to the persona that was clicked; the keyboard never leaves the harness for the switch; and a click on a readout, a separator, a heading, the row you are already on, or an unpaired release does nothing at all.
- Red-first: with `slots.py` and `builtins.py` reverted, the unit module fails 33 of its 36.
- `tools/sweep.py` — verdict below.

## Rebased onto `c183fe0`

One conflict, in `slots._bottom`, and it was #763's: that change added a top-priority `notice` field at the head of the attention row. Resolved by keeping #763's field and its reading order and running the door walk over the merged `order` tuple — the walk reads `order`, so a field added at the head moves the hint's columns by its own width plus a separator and the door follows with no case for it. `test_a_field_added_in_front_of_the_hint_moves_its_door_with_it` is that merge, pinned. #729's notice is itself correctly inert: it is the outcome of something the operator already chose.

`#774`'s `_sidebar_live` change is inside a function `_top` calls and did not touch my hunks. `switch.py`, `chats.py` and `commands_frame.py` are untouched.

### DIFF
--- a/charter/frame/builtins.py
+++ b/charter/frame/builtins.py
@@ -422,6 +422,180 @@ def _bar_events(fid: str, command: tuple[str, ...]):
     return on_event
 
 
+#: What a click on the `personas` column starts — the same front door
+#: :data:`_WORKSPACE_SWITCH` opens for the other noun, one option over
+#: (`commands_frame.cmd_switch` takes both). Not `choose.switch_to` called in process, for
+#: :data:`_CHAT_SWITCH`'s two reasons: every refusal `switch.to_persona` can raise reaches
+#: the operator through `_say_on_screen`, which is a `display-message` on the frame's own
+#: client and a surface no panel process has; and a switch re-gathers the plane, which is
+#: not a paint loop's work to sit through.
+_PERSONA_SWITCH = ("frame-switch", "--persona")
+
+#: What a click on a door opens: `F2`'s own action, spelled the way `commands_frame
+#: .conf_text` binds it minus the client name.
+#:
+#: **The client is not passed, and that is a real difference from the bind.** `conf_text`
+#: expands `#{client_name}` because tmux knows which client pressed the key; a panel
+#: process is not a `run-shell` child of a keypress and has no presser to name. What the
+#: client buys is which screen `cmd_palette`'s refusals are drawn on, and without it
+#: `_say_on_screen` falls back to `-t <session>` — the most recently attached client,
+#: measured in that function. On the one-client frame this is drawn for they are the same
+#: screen; on a two-client one a refusal can land on the other operator's status line,
+#: which is the honest cost of a doorway that is a rectangle rather than a keystroke.
+#:
+#: **And no `--chat` either, for `_bar_events`' reason**: `builtin_actions._spawn` sets
+#: `$CHARTER_SESSION_ID` to this panel's own *fid*, which is the chat this pane was
+#: launched for, and `commands_frame._pressers_chat` falls back to exactly that. The
+#: option exists because one `bind` text is shared by every frame on the socket; a panel
+#: was TOLD which frame it draws and has no such ambiguity to resolve.
+_PALETTE = ("frame-palette",)
+
+
+def _strip_events(fid: str):
+    """The identity and attention strips' handler: a click on a door opens the palette.
+
+    **The complaint this answers is that `F2 palette` is the only affordance charter
+    advertises on screen and the one thing on the row that could not be operated the way
+    it is drawn** (#751). A key name beside a noun is a button everywhere else an operator
+    has seen one. It is a button here now, and so are the workspace chip and the persona
+    head beside it — the other two nouns on those rows that the palette already opens by
+    name.
+
+    **All three go to the same place, and that is the design rather than a shortcut.**
+    §4i's rule is that the irreversible half of an interaction never rides on a pointer
+    event, because one can arrive unpaired. `_bar_events` gets out from under it by acting
+    on the PRESS and switching to a tab that is still one click away — a gesture that
+    completes on the pointer. A strip has no such gesture available: `⬢ alpha` names the
+    workspace you are ON and `◆ steward` the persona you ARE, so a click on either can
+    only mean *let me pick another one*, and picking needs a chooser. `key` is in
+    `component.EVENT_KINDS` and deliberately NOT in `events.DELIVERED` — the harness owns
+    the keyboard — so charter cannot grow a second, pointer-driven chooser inside a one-row
+    pane. The palette IS the chooser, it is the surface both nouns are reachable through
+    already, and opening it is a complete gesture: the overlay makes itself the active
+    pane (`overlay.modal_argvs`), so the keyboard reaches the list it just drew.
+
+    **The PRESS is acted on and the release is dropped**, kept word for word from
+    :func:`_repos_events` and :func:`_bar_events`: a drag begun on a pane border delivers
+    exactly one release (`frame/overlay.py` measured it), and a drag that began elsewhere
+    and happens to end over this row never pointed here.
+
+    **It answers falsy even when it opened the palette**, which is :func:`_bar_events`'
+    reading and not the table's: nothing this process draws has changed, and the pane the
+    palette carves comes off the harness rather than out of this rectangle. Truthy would
+    buy one repaint of a byte-identical row.
+
+    **No `ev.kind` test and no `scroll`**, for :func:`_bar_events`' reasons — these two
+    components declare `click` and nothing else, `events.Dispatcher._deliver` drops
+    anything they did not declare before it reaches here, and a one-row strip has nothing
+    a wheel could move.
+
+    *fid* is closed over rather than read back out of `$CHARTER_SESSION_ID`: one tmux
+    server is shared by every frame on the machine, and this process was told which frame
+    it draws.
+    """
+    def on_event(ev):
+        from .. import util
+        from . import slots as _slots
+        from .builtin_actions import _spawn
+        if not ev.pressed or ev.name != _ACT_BUTTON:
+            return False
+        if not _slots.DOORS.opens_palette(ev.col):
+            return False
+        _spawn(util.self_relaunch_argv(*_PALETTE), fid=fid)
+        return False
+
+    return on_event
+
+
+def _persona_events(fid: str):
+    """The sidebar's handler: a persona's NAME switches to it, its BADGES say what they mean.
+
+    **Two meanings on one row, resolved by which CELL the pointer landed in** — which is
+    not a second gesture bolted on, but the layout `slots._persona_rows` already draws
+    being read back. A row is a name cell and a badge cell, and they are asking different
+    questions: the name is *be this persona*, the badges are *what is that glyph*.
+    `slots._Chips.hit` owns the resolution and this branches on what it answers.
+
+    **The complaint is that the sidebar draws the roster as an inverted-row list with a
+    cursor-looking marker on the active one — the visual vocabulary of something you pick
+    from — and handles no events at all** (#742). It picks from now on.
+
+    **A click SWITCHES, exactly as it does on the two tab bars, and the three reasons
+    :func:`_bar_events` states hold here unchanged**: this acts on the PRESS, so the one
+    event that can arrive unpaired (a release from a drag begun elsewhere) switches
+    nothing; a persona switch is reversible by the identical gesture, because the row you
+    left is still in the column one click away and nothing is created, destroyed or
+    started; and there is no chooser a select-then-confirm column could be confirmed with,
+    since `key` is in `component.EVENT_KINDS` and deliberately not in `events.DELIVERED`.
+    A column that merely *selected* would draw a second mark beside `▸` that nothing on
+    the machine could ever act on.
+
+    That is where this deliberately departs from #742's own suggestion, which asked for
+    "selecting, not switching — the same 'a click only ever selects' rule the repo table
+    follows". The table's rule is not "a click never switches", it is `frame/overlay.py`'s
+    "the irreversible half is never driven by an event that can arrive unpaired", and what
+    makes selection the right answer THERE is that the table has a real intermediate state
+    (a highlighted row, a detail on the attention strip) and `Enter` in the palette to
+    choose from it. A persona column has neither: the row in reverse video IS the
+    selection, and a second one would be dead code wearing a feature's name.
+
+    **The rows that answer nothing are `slots._Chips`' to refuse, not this handler's** —
+    the `▪ personas 6` heading, the `…(+N more)` row, the `no personas` sentence, the
+    blank rows between the sidebar's sections, every todo and change row below them, and
+    the persona you are already being. See that class; the last of those is
+    `_Tabs.switch_to`'s rule and is also what stops a double-click switching twice.
+
+    **The badge half deliberately does NOT refuse the persona you are being.** *"What does
+    the flag on my own row mean"* is the commonest form of the question #753 is about, and
+    answering it is not the no-op that re-adopting yourself would be. That asymmetry is
+    `_Chips.hit`'s and is argued there.
+
+    **Falsy even when it started a switch**, for :func:`_bar_events`' reason:
+    `switch.to_persona` ends in a `state.bump` of its own, which is the version this
+    panel's poll is already watching, so truthy would buy one repaint of a row that has
+    not changed yet.
+
+    **It is declared on the `sidebar` COMPOSITE and nowhere else, because
+    `frame/registry.py` refuses a part that declares events** — a part is never placed
+    (`on_edge`), and `panel._dispatcher` asks `wanted()` of the component that WAS placed,
+    so a declaration on `personas` would pass every check, build a handler and receive
+    nothing, ever. That refusal says the composite "is also the only thing that knows
+    which of its parts a pointer would have been over"; here it does not have to work that
+    out, because `slots.persona_section` is the one pass that composes those rows and they
+    are the first thing the pane draws, so `slots.CHIPS` needs no base and no offset.
+
+    The cost is stated at the `personas` registration: a plane that places that component
+    on its own gets the column drawn and inert.
+    """
+    def on_event(ev):
+        from .. import util
+        from . import slots as _slots
+        from . import state
+        from .builtin_actions import _spawn
+        if not ev.pressed or ev.name != _ACT_BUTTON:
+            return False
+        hit = _slots.CHIPS.hit(ev.row, ev.col)
+        if hit is None:
+            return False
+        if hit.explain:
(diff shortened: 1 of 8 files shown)


diazoxide:
## All three sweep survivors settled — and all three were genuinely mine

I checked #776 first, since the sweep charges the merge commit against the merge-base and a branch behind `main` gets billed for `main`'s lines. **Not the case here:** all three sit in `charter/frame/slots.py` inside functions this branch adds (`_badge_width`, and the `_Hit` sentinels), and the branch is now rebased onto `13b9a2b` anyway. Head is `dda1c43`.

| # | `path:line:operator` | verdict |
|---|---|---|
| 1 | `charter/frame/slots.py:1681` `[unclamp]` `max(0, width - _NAME_MIN_W)` | **equivalent where it sat — measured.** Clamp relocated so it is observable; now pinned by a direct test |
| 2 | `charter/frame/slots.py:1797` `[retune-string]` `"switch"` | **equivalent — the value was unobservable.** Constant deleted; the sentinel is a bool now |
| 3 | `charter/frame/slots.py:1800` `[retune-string]` `"explain"` | same as 2, same fix |

---

### 1. `max(0, width - _NAME_MIN_W)` in `_badge_width` — equivalent, and I can show it

This is the one line in the three that I did not *write*: it came over verbatim from `_persona_rows` when I extracted `_badge_width` so the renderer and `persona_section` could not disagree about where the badge column starts. Extraction makes it an added line, so the sweep charges it here — correctly, because on this branch it now has a second caller.

**Two measurements rather than a reading.**

*Is the clamp ever exercised?* I instrumented `_badge_width` to log every `(width, width - _NAME_MIN_W)` pair and ran the whole suite:

```
total _badge_width calls across the suite : 197
calls where width - _NAME_MIN_W < 0       : 0
narrowest width ever passed               : 20      (_NAME_MIN_W is 12)
distinct widths                           : 33
```

*Would it matter if it were?* A differential run of the shipped expression against the mutant, over every cell configuration I could build — 5 badge shapes × 1–4 personas × every width from 0 to 40:

```
cell configurations compared                 : 3,280
configurations where the DRAWN rows differ   : 0
(row, col) pairs where the CLICK differs     : 0
```

So it was a true equivalent, and the reason is structural: a negative badge width fell straight into `_persona_rows`' own `badge_w <= 0` branch, which already treats it exactly as zero. **A guard whose effect another branch is already providing is a line no input can turn red.**

**The fix is not to delete it — it is to move it somewhere it answers for itself.** The floor now sits *outside* the `min`:

```python
max(0, min(widest_badge, width - _NAME_MIN_W))     # was: min(widest_badge, max(0, width - _NAME_MIN_W))
```

Provably equal wherever the old form was defined (the widest badge is never negative, so both are zero below the floor), but now it is *this function's* contract — *a column is never a negative number of cells* — instead of a property of whatever the caller does next. One assertion pins it, and `test_a_pane_too_narrow_for_a_name_gets_no_badge_column_rather_than_a_negative_one` is it. It also makes `persona_section`'s `width - badge_w` honest: a negative badge width used to publish a badge column starting *past* the pane — unreachable by any click, and wrong in the way that only surfaces later.

Reachability, for the record: `_PAD_MIN_CONTENT` **is** `_NAME_MIN_W`, so `pad_for` already guarantees a padded pane keeps 12 columns. The only route below the floor is a pane narrower than 12, which a plane can ask for with a `[[frame.component]]` of its own — real, and not something the suite happens to build.

### 2 & 3. `SWITCH = "switch"` / `EXPLAIN = "explain"` — the sweep was right, and the constants are gone

`_Chips.hit` returned `_Hit(kind: str, name: str)` against two module constants. Every producer and every consumer went through the constant, so the *value* was unobservable — `'txjudi'` and `'fyqmbjo'` work exactly as well as `"switch"` and `"explain"`. That is a literal nothing can turn red, and no test could honestly pin it without asserting a spelling for its own sake.

**So the strings are deleted rather than pinned.** `_Hit` carries a bool:

```python
class _Hit(NamedTuple):
    explain: bool     # True: the pointer was in the BADGE cell. False: the name cell.
    name: str
```

There is no spelling left to get wrong, and the flag is observable in both directions — flipping it on the badge branch fails 4 cases, on the name branch 11. The column is what matters here, not the vocabulary: there are exactly two things on a persona row, decided by which cell the pointer landed in, and a third would be a different question rather than a third value of this one.

### Hand-verification

`python3 -B`, `__pycache__` cleared between every run:

```
survivor 1: drop the relocated clamp                        -> FAILED (failures=1)
survivor 1b: clamp back inside the min (the old spelling)   -> OK          <- the equivalence, confirmed
survivors 2+3: flip the explain flag on the badge branch    -> FAILED (failures=4)
survivors 2+3: flip the explain flag on the name branch     -> FAILED (failures=11)
```

The second row is deliberate: it is the measurement backing the claim that the old form could not be pinned, rather than an assertion that it could.

### Re-verified on the new head

- Full suite green in six shards — **9,303 tests, 1 skip, 0 failures**.
- `tests/test_a_real_click_opens_the_real_palette.py` on **real tmux 3.7c and the 3.2 floor, twice each**: 10 cases, all green.
- Rebased onto `13b9a2b`. The only conflict across both rebases was #763's `notice` field in `slots._bottom`, resolved by running the door walk over the merged field order — `test_a_field_added_in_front_of_the_hint_moves_its_door_with_it` pins that merge, so the next field added at the head of the attention row moves the `F2 palette` door with it and needs no case of its own.
- Sweep re-fired on `dda1c43` both ways: the `pull_request` run went out on the push (`33366382544`) and I dispatched a second (`33366395312`), since a force-push firing `synchronize` is intermittent.

diazoxide:
**Green on `dda1c435936699ef7f662830b2d27540c8c0e4a3`, rebased onto `13b9a2b`.**

```
no survivors                     -> pass
Add up what the shards found     -> pass
Sweep shard 1 of 2               -> pass
Sweep shard 2 of 2               -> pass
Size the sweep                   -> pass
test (3.11 / 3.12 / 3.13 / 3.14) -> pass
mergeStateStatus                 -> CLEAN
```

An independent local sweep agrees: `dda1c4359366` against `13b9a2bf69e2`, baseline `Ran 9366 tests — OK`, **35 mutations applied, 35 pinned, 0 survived, 0 unresolved** — `gate: no survivors`.

Both sweep runs went out on this head: the `pull_request` one fired on the push (`33366382544`) and I dispatched a second (`33366395312`), since `synchronize` on a force-push is intermittent.

Ready to merge. It closes #751 outright — the identity and attention strips being inert to a click, `F2 palette` included — and closes #753, whose badge legend now answers in the frame through #729's dwell. **#742 stays open on its todo-overflow half**, which is untouched here and has its reasoning recorded on the issue.

Issue #631, opened 2026-08-28
The frame's own surface draws a box around the harness pane


diazoxide:
#628 closed the seam between painted panels by putting the surface on the **window's** two
border options. That reaches every rule in the frame — including the three that separate a
panel from the pane the operator's harness runs in.

Measured on a live frame with `chrome = "dark"` and `bg = "brightblack"` on all four
components:

```
pane-border-style = 'fg=default,dim,bg=brightblack'     <- window-wide
panels:  window-style='bg=brightblack'  active='bg=black'
harness: window-style=''                                <- bare, correctly
```

Every border is grey, so the dark harness sits inside a grey box on three sides. One
report traded for another: charter is drawing chrome around the one rectangle ADR 0018
says it never draws in.

## The cell between two panes belongs to neither

A window-wide value cannot tell a panel apart from the harness, because it is one value for
the whole window. Per-pane border options can, and they exist above tmux 3.7 — with a trap
below it:

| | tmux 3.7c | tmux 3.2–3.6a |
|---|---|---|
| `options-table.c` scope | `OPTIONS_TABLE_WINDOW\|OPTIONS_TABLE_PANE` | `OPTIONS_TABLE_WINDOW` |
| `set -p -t <panel> pane-border-style` | stores on the pane | **rc 0, writes the WINDOW** |
| `show -p` on a pane nobody set | `''` | the window's value |
| `set -p -u` | removes the pane's | removes the **window's** — charter's #514 pin |

The failure below the line is silent rather than refused, so there is nothing for
`tmuxctl.run` to report, and a probe would have to perform the damaging write to discover
it.

Issue #627, opened 2026-08-28
The frame's rules are seams: a coloured pane's background stops at the border tmux draws


diazoxide:
With `chrome = "dark"` and `bg = "brightblack"` on all four components (this plane's own
committed `charter.toml`), the panels come out grey and **a one-cell dark strip runs
between every pair of them**, and around the sidebar. The frame reads as coloured
rectangles separated by seams rather than as an application. Reported off a screenshot.

## The cause, measured on tmux 3.7c

`window-style` paints a **pane's** rectangle. The rule tmux draws between two panes is in
neither rectangle — it comes from `pane-border-style`, which `commands_frame._CHROME` has
pinned to `fg=default,dim` since #514: a foreground and an attribute, and **no
background**. So the border cell stays whatever the operator's terminal already was.

Read off an attached client's wire through a nested tmux, three panes each
`bg = "brightblack"`:

```
ESC[100m                                          ESC[2mESC[49m│ESC[0mESC[100m
                                                          ^^^^^^^^ the seam
```

`ESC[49m` — default background — for the one cell between two `ESC[100m` runs.

## What the fix has to decide

A border sits **between two panes**, and with per-component `bg` those two panes may be
different colours, so there is no border colour that matches both. tmux draws one border,
not two half-borders.

Scope is also not free:

* `pane-border-style` **is** a pane option on 3.7c, and tmux draws each border cell from
  the pane **above or left** of it, ignoring the other side's outright (window `bg=blue`
  + left pane `bg=red` + right pane `bg=green` renders red).
* At `tmuxctl.FLOOR` (3.2) it is **not** a pane option at all: `set -p` returns 0 but
  writes the **window's** value, `show -p` on a pane nobody set answers the window's, and
  `set -p -u` removes charter's own #514 pin for the whole window.

So a per-side design is unavailable at the floor, silently window-wide there, and
one-sided where it exists.

`pane-active-border-style` defaults to a format expression
(`#{?pane_in_mode,fg=yellow,#{?synchronize-panes,fg=red,fg=green}}`) — charter already
replaces it wholesale, which is #514's fix, so whatever is decided has to say what happens
to the active-pane and copy-mode indication rather than losing it by accident.

PR #768, opened 2026-08-30
The frame's rules and its text palette are the plane's to choose

diazoxide:
Two keys the operator hit on their own frame within ten minutes of each other, and one measurement that closes a door behind them. Both are "the frame's appearance is the plane's to choose".

## Part 1 — `[frame] rules`, defaulting to `hidden`

tmux **always** paints a glyph on a border cell: `pane-border-lines` takes `single`/`double`/`heavy`/`simple`/`number` and has no `none`. So the seam reported in #627, #631, #657 — and again against the frame that fixed #657 — was never a character question. It is a contrast question.

`rules = "hidden"` asks for the glyph in the same colour as the cell behind it. Read off an attached client's wire through a nested tmux, three panes each `bg = "brightblack"`:

```
visible  ESC[2m  ESC[100m ─────    a dim foreground glyph, on the surface
hidden   ESC[90m ESC[100m ─────    the glyph IS the surface
```

That is byte-for-byte the style the operator hand-applied to their own live frame and confirmed (`fg=brightblack,bg=brightblack`, no `dim`). This makes it permanent.

**It ships as the default and cannot make an existing frame worse.** A plane with no surface has no colour for the glyph to take, so `rule_style` composes exactly the string charter has drawn since #514 — asserted, not argued (`test_a_surfaceless_frame_is_byte_identical_under_either_word`). It only does anything on a plane that already wrote a `chrome` or a `bg` by hand.

**One arm is reached by an input that reaches only it:** a component with `bg = "default"` keeps its *visible* rule. `default` means the terminal's own background and there is no `fg=` spelling for that — `fg=default` is the terminal's own *foreground* — so honouring `hidden` there would draw the rule at full strength, brighter than the `dim` it replaced.

Follows `FRAME_PANE_BG`'s shape throughout: a **word**, never a style string. Measured again on this option — `set -w pane-border-style 'bg=#{?#{==:1,1},colour196,colour46}'` is rc 0 and reaches the wire as `ESC[48;5;196m`.

### Charter must be told, not compute

Established rather than assumed. The sixteen ANSI names have no fixed RGB; `brightblack` is a dark grey on most themes and a light tan on this operator's. Charter cannot find out which: OSC 11 through tmux answers nothing, and `$COLORTERM` inside a pane describes the terminal that started the **server**. **If charter cannot know what a colour looks like, it cannot compute contrast** — so the plane is asked, in the vocabulary it already answers `bg` in.

### Three keys, not eight — the axis argued

Of `chrome.recipes()`' eight roles, four are theme-safe by construction (`heading` bold, `selected` reverse, `reset`, `inset`) and `chrome.py` already argues why. `ok`/`warn`/`bad` are slots in the operator's *own* palette, which is `FRAME_PANE_COLOURS`' own argument for refusing `colour24`. What is left is the two that are **not** slots:

- **`text`** — the pane's default foreground, set through `window-style fg=`. tmux resolves a pane's default cell from it, so charter's own `ESC[0m` returns to the plane's colour rather than the terminal's, and neither the ~40 `statusline._DIM` sites in `slots.py` nor a provider's component has to be told. Measured on 3.7c and at the 3.2 floor.
- **`dim`** — its own key because it is the one thing in the frame **wrong by construction** on a surface it was not chosen for. `_role_values` claimed bold, dim and reverse "are statements relative to whatever the operator's terminal already is"; that is the wrong half of true for dim, which is relative in one direction — toward the background. This PR corrects that docstring. Measured on the operator's live frame, `ESC[2m` is the most frequent SGR charter paints (12, 8, 6 and 3 occurrences across their four panels).

Default stays `dim = true`: dim is the only thing separating muted text from ordinary text, so off-everywhere is a flat hierarchy.

## The refutation — `reverse` cannot be a default surface

Measured on 3.7c and at `tmuxctl.FLOOR`, through a nested client: tmux **accepts** `window-style reverse`, stores it, reads it back verbatim — and puts **no SGR at all** on the wire. Same for `bold` and `dim`. A pane style honours colour and silently ignores every attribute. A renderer-side reverse is worse rather than harder: `_surface_argvs`' own docstring measures that tmux fills the whole rectangle including cells no renderer wrote and across resize, which a panel's own fill cannot.

So there is no theme-independent surface to ship, `chrome` stays `off`, and the consequence is documented rather than worked around. Pinned by `NoAttributeInAPaneStyleReachesTheWire` **with a control proving tmux accepts the value**, so a future charter that sets it and checks the return code cannot report success while painting nothing.

## `NO_COLOR`

`chrome.no_colour()` remains the one reading; nothing here adds a second. `recipes()` already returned `""` for every role when colour is off, so no-colour is a **superset** of dim-off and they meet in one ordering rather than fighting — asserted both ways. In `panel._write` the dim pass is unreachable under `NO_COLOR` rather than skipped by it.

`chrome.served_params` deliberately still reads `_role_values`, so SGR 2 stays *admitted* whatever the plane said and a provider's own dim is deleted on the way out rather than escaped into its pane as text. Deriving it from `recipes()` — the tidier-looking wiring — would have re-opened #707 for exactly one parameter. Pinned.

## Boundary: refuse or degrade

These are `[frame]` keys, so an unreadable word degrades to the shipped default and leaves the frame whole — `chrome`, `density` and `hotkey`'s existing contract. #535's refuse-the-arrangement-whole rule is about a **component** silently missing, which reads as a plane with no clones; a `rules` typo costs a rule drawn the shipped way in an otherwise intact frame. #687's warning applies in the other direction, so every word in both vocabularies is asserted **accepted** rather than sampled.

## What this does not fix — named, not papered over

`text` fixes a pane whose background is *inverted* relative to the terminal's. On a terminal that is already light it has nothing to do, and what is hard to read there is the **accent** triple — `ok`/`warn`/`bad`, still the operator's own green, yellow and red, and yellow on tan is the combination no palette was designed for. That is a real remaining gap, stated in `docs/frame.md` and the news entry. It is not in this PR because the only mechanism reaching charter's own renderers is routing ~40 `slots.py` call sites through the recipe table — a change worth making on its own.

## Verification

- Full suite green (8930+ tests), `unittest`, `env -u PYTHONSAFEPATH`.
- **59 new tests** across two files. `tests/test_a_planes_frame_really_reads_that_way.py` drives real tmux on **3.7c and the 3.2 floor**, renders the control first every time, and skips rather than passing quietly on a machine whose two readings agree.
- Rebased onto current `main` (through #759). Verified against #760's per-option floor: a floor says which tmuxes are *told* about an option, a `look` says what the two style rows *carry* — orthogonal, and pinned.
- Live assertions are property-based (`bg - fg == 10`), because nesting tmux downsamples `brightblack` at the floor and the spelling is the machine's while the property is charter's.
- `dependencies = []` untouched; nothing in `charter/__init__.py`, `pyproject.toml`, `.claude-plugin/plugin.json`, `hooks/hooks.json`, or any `0.54.0-*` note.
- The operator's frame was read only (`capture-pane -p -e`, `show -pv`); no option was set on socket `charter` and its 11 sessions are intact.

### DIFF
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -2137,7 +2137,21 @@ def _remain_on_exit_argv(*, socket: str, harness_pane: str) -> list[str]:
 #: the 256 and imposed on a theme it cannot see. A terminal too old to honour SGR 2
 #: renders both borders in the plain default instead, which is still ONE colour — the
 #: property here is sameness, and it survives the attribute being dropped.
-_CHROME_STYLE = "fg=default,dim"
+#: The foreground half of it, alone — what a rule is drawn IN before the plane has said
+#: anything. Split out because the ``dim`` beside it is a decision now (`[frame] dim`) and
+#: the surface behind it always was: `instance.rule_style` composes all three, and handing
+#: it the assembled string instead would make it take an attribute back out of one.
+_CHROME_FG = "fg=default"
+
+#: The style every rule in the frame is drawn in when the plane has said nothing — and the
+#: MARKER in :data:`_CHROME` for "this entry is a style", which is how `_chrome_argvs` knows
+#: which of the five options carry a colour without a second list saying so.
+#:
+#: `instance.rule_style(None, _CHROME_FG, instance.look_of({}))` is this string, and
+#: `TheShippedRuleIsStillTheOneCharterAlwaysDrew` asserts it rather than leaving a reader to
+#: check by eye — so the day the shipped `Look` changes, the table stops agreeing with the
+#: assembler loudly instead of quietly.
+_CHROME_STYLE = f"{_CHROME_FG},dim"
 
 #: Every window option tmux consults to draw a pane border, pinned to charter's own
 #: answer. #514: the frame's rules came out in two colours, and neither of them was
@@ -2207,7 +2221,8 @@ def _chrome_values() -> dict[str, str]:
 
 
 def _chrome_argvs(*, socket: str, harness_pane: str, v: tuple[int, int] | None,
-                  surface: str | None = None) -> list[list[str]]:
+                  surface: str | None = None,
+                  look: instance.Look = instance.SHIPPED_LOOK) -> list[list[str]]:
     """`set-option -w`: charter's own answer for every option tmux draws a border from.
 
     **Charter owns the frame's chrome, and tmux draws all of it.** #514 named two
@@ -2280,6 +2295,26 @@ def _chrome_argvs(*, socket: str, harness_pane: str, v: tuple[int, int] | None,
     exactly what it is about. Asked through `chrome.no_colour` so there is one reading of
     the variable (#547), the same way `_surface_argvs` asks.
 
+    **It refuses the plane's own `text` colour by the same line**, and that falls out of
+    the surface being refused rather than needing a second answer: with no surface,
+    `instance.rule_style` has nothing to hide a rule in, and the foreground it composes is
+    the only thing `NO_COLOR` would still have to argue about. It does not — a `text` word
+    IS a colour charter was asked to paint, so it goes with the background. What survives
+    is `_CHROME_FG` and the `dim`, which are the operator's own foreground and an
+    attribute over it, exactly as before.
+
+    ***look* is the plane's own `[frame] rules`, `text` and `dim`** — the frame-wide
+    appearance, resolved once by `instance.look_of` and passed whole rather than as three
+    keywords, so this and the three functions below cannot be handed different halves of
+    one frame's answer. `instance.rule_style` is where all three are spent; nothing here
+    reads a word out of it.
+
+    **It is a different question from *v* and the two do not meet.** A floor says which
+    tmuxes are told about an option at all; a `look` says what the two style rows carry
+    when they are. `pane-border-indicators` is dropped below its floor whatever the plane
+    asked for, and a `hidden` rule is composed identically on every tmux that gets one —
+    which is why the `>= floor` filter below is applied to the row and never to the value.
+
     ***v* is this tmux's version, and every row of `_CHROME` is filtered through its own
     floor by it (#716).** REQUIRED rather than defaulted, because a default is the shape
     the defect had: `pane-border-indicators` does not exist below
@@ -2296,8 +2331,8 @@ def _chrome_argvs(*, socket: str, harness_pane: str, v: tuple[int, int] | None,
     supported baseline: pin everything every tmux charter supports has, and leave the one
     option that arrived after it to a tmux that says so.
     """
-    rule = (f"{_CHROME_STYLE},{surface}"
-            if surface and not chrome_mod.no_colour() else _CHROME_STYLE)
+    rule = instance.rule_style(
+        None if chrome_mod.no_colour() else surface, _CHROME_FG, look)
     known = v if v is not None else tmuxctl.FLOOR
     return [tmuxctl.server_argv(socket, "set-option", "-w", "-t", harness_pane, name,
                                 rule if value == _CHROME_STYLE else value)
@@ -2322,7 +2357,9 @@ def _pane_borders_wanted(v: tuple[int, int] | None) -> bool:
     return v is not None and v >= tmuxctl.PANE_BORDER_FLOOR
 
 
-def _pane_border_pairs(*, chrome, bg, pane_borders: bool) -> tuple[tuple[str, str], ...]:
+def _pane_border_pairs(*, chrome, bg, pane_borders: bool,
+                       look: instance.Look = instance.SHIPPED_LOOK
+                       ) -> tuple[tuple[str, str], ...]:
     """This PANEL pane's own edges, in this panel's own surface — or nothing at all.
 
     **A panel's edges are its own colour, which is the half of #631 that stands.** A
@@ -2357,12 +2394,13 @@ def _pane_border_pairs(*, chrome, bg, pane_borders: bool) -> tuple[tuple[str, st
     `_CHROME_STYLE`, so no operator string reaches a style here any more than it does one
     function up.
     """
-    return (instance.pane_border_options(bg, chrome, _CHROME_STYLE)
+    return (instance.pane_border_options(bg, chrome, _CHROME_FG, look)
             if pane_borders else ())
 
 
 def _harness_rule_argvs(*, socket: str, harness_pane: str, surface: str | None,
-                        pane_borders: bool) -> list[list[str]]:
+                        pane_borders: bool,
+                        look: instance.Look = instance.SHIPPED_LOOK) -> list[list[str]]:
     """`set-option -p`: the rules AROUND the harness pane, in the frame's own surface —
     and never, at any value, anything that reaches INSIDE it.
 
@@ -2449,7 +2487,7 @@ def _harness_rule_argvs(*, socket: str, harness_pane: str, surface: str | None,
     if not pane_borders:
         return []
     pairs = instance.rule_options(None if chrome_mod.no_colour() else surface,
-                                  _CHROME_STYLE)
+                                  _CHROME_FG, look)
     if not pairs:
         return [tmuxctl.server_argv(socket, "set-option", "-p", "-u", "-t", harness_pane,
                                     name) for name in instance.PANE_BORDER_OPTIONS]
@@ -2458,7 +2496,8 @@ def _harness_rule_argvs(*, socket: str, harness_pane: str, surface: str | None,
 
 
 def _surface_argvs(*, socket: str, pane_id: str, chrome, bg=None,
-                   pane_borders: bool = False) -> list[list[str]]:
+                   pane_borders: bool = False,
+                   look: instance.Look = instance.SHIPPED_LOOK) -> list[list[str]]:
     """`set-option -p`: the pane surface `[frame] chrome` asked for, on ONE panel pane.
 
     **tmux paints the background; charter sets an option.** `window-style` and
@@ -2534,9 +2573,8 @@ def _surface_argvs(*, socket: str, pane_id: str, chrome, bg=None,
     # takes the frame's own answer". One expression, so there is no state in which a pane
     # gets one option from here and the other from there.
     return [tmuxctl.server_argv(socket, "set-option", "-p", "-t", pane_id, name, value)
-            for name, value in (*(instance.pane_bg_options(bg)
-                                  or instance.chrome_options(chrome)),
-                                *_pane_border_pairs(chrome=chrome, bg=bg,
+            for name, value in (*instance.surface_options(bg, chrome, look),
+                                *_pane_border_pairs(chrome=chrome, bg=bg, look=look,
                                                     pane_borders=pane_borders))]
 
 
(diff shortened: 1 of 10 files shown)


diazoxide:
## Sweep: `4 survivors, 9 not measured` — classified

Swept commit `81d475f` (the PR merge ref: `5859e28` onto `03fdbd3`), **not** my branch head. Every line number below is against that commit, checked out in a detached worktree, because `Mutation.tag` is line-keyed and this diff moves lines.

### The four survivors — three were real gaps, one could not earn its place

All four were in `charter/frame/chrome.py`, in `undim`/`_without_dim`. Three sit in the same three-line walk, which is why the tool clustered them and pinned none.

| line | operator | mutation | verdict |
|---|---|---|---|
| 260 | `no-fallback` | `{5: 3, 2: 5}.get(space, 1)` → `[space]` | **real gap** — fixed |
| 259 | `collapse-ifexp` | drop the `else -1` | **real gap** — fixed |
| 259 | `shift-boundary` | `i + 1 < len(pieces)` → `<=` | **real gap** — fixed |
| 294 | `drop-if` | delete the `if "\x1b" not in row` fast path | **equivalent — line deleted** |

**The three gaps are crashes, not cosmetics.** `_without_dim` walks an SGR parameter list and must size an extended-colour run (`38`/`48`/`58` are followed by a colour space and its arguments) before it can skip past one. Each mutant turns a row a provider can really write into an exception raised inside `panel._write`, on the repaint path:

- `{5: 3, 2: 5}[space]` — **`KeyError`** on `ESC[38;99m`, a colour space that is neither 5 nor 2.
- dropping the `else -1`, or widening `<` to `<=` — **`IndexError`** on `ESC[38m`, an introducer at the end of its own list.

Charter writes none of those escapes — `instance.FRAME_PANE_COLOURS` argues at length why it names the sixteen — and that is exactly what makes the case real rather than theoretical: this runs over a row a **provider's** component finished, and `tui.sanitize` passes any `ESC[<digits and semicolons>m` through untouched. My existing cases covered `38;5;n` and `38;2;r;g;b` and never a truncated or unknown introducer.

Three new cases cover it, and each was verified to turn its mutant red rather than assumed to:

```
mutant                 ran  verdict
no-fallback   260       49  KILLED (red)
collapse-ifexp 259      49  KILLED (red)
shift-boundary 259      49  KILLED (red)
```

**The fourth is equivalent, and it is deleted rather than defended.** Timed on this machine, 200 000 calls each, against the two rows a repaint actually carries:

```
row with no escape   guard 0.032us   no guard 0.111us   saves 0.079us  (3.5x)
painted row          guard 1.933us   no guard 1.977us   saves 0.044us  (1.0x)
```

3.5× reads like a reason until it is spent. At 0.079 µs a line, a 50-line repaint saves **4 µs against `render`'s own measured 4816 µs** — under a tenth of a percent — on a path that only runs at all for a plane that turned `dim` off. And most rows in this frame carry colour, so the fast path rarely fires. `re.sub` returns the string it was given when nothing matched, so the guard could not change an outcome either. "Equivalent mutant" and "dead code" are one finding, so the line is gone.

That also corrected a test: `test_a_row_with_no_escape_at_all_comes_back_unchanged` asserted `assertIs`, and kept passing with the guard deleted — the identity was CPython's answer, not charter's. It asserts equality now, which is the property.

### The nine not measured — one finding, and not a timeout

**No shard died and nothing was OOM-killed.** All three shards concluded `success`; the collector logged `merged 60 result(s) from 3 of 3 shard(s)`. This is not #773.

**It is also not a timeout, despite the annotation saying so.** All nine are one thing: *the mutant cannot be imported*, so `unittest` never prints a `Ran N tests` line, `_outcome` returns `conclusive=False`, and `decide` retries once and correctly refuses a verdict. The cause is that `instance.SHIPPED_LOOK = look_of({})` is evaluated at **module scope** — it has to be, because it is the default argument for four `commands_frame` functions — so it reaches both `FRAME_DEFAULTS[...]` and the `.get` fallback at import time.

Reproduced locally against `81d475f`, all nine:

```
1021/1022/1023  no-fallback     frame.get("rules", ...) -> frame["rules"]   KeyError: 'rules'
1021/1022/1023  retune-string   FRAME_DEFAULTS["rules"] -> ['svmft']        KeyError: 'svmft'
1527/1535/1555  retune-string   FRAME_FIELDS key "rules" -> 'svmft'         KeyError: 'rules'
```

**A correction to my own first pass, which the shard logs caught.** Lines 1021–1023 each carry the literal **twice** — `frame.get("rules", FRAME_DEFAULTS["rules"])` — so the sweep generates two `retune-string` mutations per line and they landed in different shards. Shard 1 **pinned** the `frame.get(...)` argument on all three lines; shards 2 and 3 got the `FRAME_DEFAULTS[...]` subscript and could not measure it. My first reproduction used `str.replace(..., 1)` and hit occurrence 1, so it showed red where CI showed no-verdict. Verified per occurrence:

```
line 1021  occurrence 1 (frame.get argument)      -> clean import   [shard 1: PINNED]
line 1021  occurrence 2 (FRAME_DEFAULTS subscript) -> KeyError       [shard 2: UNRESOLVED]
```

**Timing rules out a timeout independently.** `SUBSET_TIMEOUT` is 900 s; the 20-module subset runs 1438 tests in **44 s** here, 20× inside the budget. And shard 2's three UNRESOLVED lines are timestamped `00:03:06.166`, `00:03:06.621`, `00:03:32.204` — the first two **0.46 s apart**. Nothing timed out; the interpreter died before collection.

So the honest reading is that these nine are **killed harder than any test could kill them** — the mutant does not merely fail a test, it fails to load. The sweep is right to refuse a verdict (it measures test outcomes, not import outcomes) and I have added no test for them, because a test asserting `import charter.instance` works cannot run in a suite that cannot import it. Worth noting as a tool observation rather than a code change: the annotation text for this bucket reads *"The run timed out rather than failing"*, which is wrong for a mutation that breaks import, and the two are worth telling apart — one says the machine was busy, the other says the mutation is catastrophic.

### State

Rebased onto `c183fe0` (through #774/#763), full suite **9301 tests, OK, 1 skip**, all four Python versions green. Sweep re-fired on the new head.

diazoxide:
## Sweep, second run: **0 survivors**, same 9 not measured — and it is not #773

Run [33359034355](https://github.com/diazoxide/charter/actions/runs/33359034355) on `0ffa3ab`, after the survivor fixes above.

### The four survivors are gone

`4 survivors, 9 not measured` → `no verdict: 9 not measured`. The three new cases and the deleted fast path did what they were meant to.

### Shard accounting — every shard reported

The question was which shard did not report and at what mutation count. **All three did.**

| shard | mutations | pinned | unresolved | job conclusion | artifact |
|---|---|---|---|---|---|
| 1 of 3 | 20 | 16 | 4 | `success` | present |
| 2 of 3 | 20 | 16 | 4 | `success` | present |
| 3 of 3 | 19 | 18 | 1 | `success` | present |
| **total** | **59** | **50** | **9** | — | `merged 59 result(s) from 3 of 3 shard(s)` |

No `exit 143`, no missing shard, no truncated artifact, and no mutation count that stops short. **This is not #773's signature.** The previous run was the same shape — 60 results from 3 of 3, all three `success` — so neither run lost a runner.

### The 9 are one finding: the mutant cannot be imported

Every one of them makes `charter/instance.py` raise at **module scope**, because `SHIPPED_LOOK = look_of({})` is evaluated at import — it has to be, since it is the default argument for four `commands_frame` functions. `unittest` never reaches a `Ran N tests` line, `_outcome` returns `conclusive=False`, `decide` retries once and correctly refuses a verdict.

Reproduced locally, all nine, each in well under a second:

```
1021/1022/1023  no-fallback     frame.get("rules", ...) -> frame["rules"]    KeyError: 'rules'
1021/1022/1023  retune-string   FRAME_DEFAULTS["rules"] -> ['svmft']         KeyError: 'svmft'
1527/1535/1555  retune-string   FRAME_FIELDS key "rules" -> 'svmft'          KeyError: 'rules'
```

Timing rules a timeout out independently: `SUBSET_TIMEOUT` is 900 s and the 20-module subset runs 1438 tests in **44 s** here — 20× inside the budget — while shard 2's three UNRESOLVED lines were timestamped **0.46 s apart**. Nothing was slow; the interpreter died before collection.

**I am not refactoring to make them measurable, and that is a decision rather than a shrug.** The only way to get a verdict on these nine is to stop computing `SHIPPED_LOOK` at import — defer it behind a function and give the four call sites a `None` default. That would convert a charter-internal inconsistency from *fails immediately, at import, in a module every command loads* into *fails later, on a repaint*. `FRAME_DEFAULTS` is charter's own constant and not operator input, so the only thing that can raise here is a charter bug, and failing fast on one is the right response. Changing production design so a mutation tool can score it is the trade this repo refuses everywhere else.

So: these nine are unmeasurable **by construction**, and they are killed harder than any test could kill them — the mutant does not fail a test, it fails to load.

One tool observation, offered rather than filed: the annotation for this bucket reads *"The run timed out rather than failing"*. That is wrong for a mutation that breaks import, and the two are worth telling apart — one says the machine was busy, the other says the mutation is catastrophic.

### #776 checked, and its tell is absent

The plan's 59 mutations fall entirely inside the four files this branch touches:

```
charter/instance.py        35
charter/frame/chrome.py    15
charter/commands_frame.py   7
charter/frame/panel.py      2
```

No file this branch does not touch appears, so the merge-commit-versus-merge-base inflation #776 describes did not charge this branch for `main`'s lines. Rebased onto `5de0794` anyway (through #777/#775) and re-swept, since the rebase is worth having regardless.

### State

Rebased, force-pushed to `cf38ece`, and both `tests` and `deletion sweep` fired on the new head — so `synchronize` did fire here. All five required contexts were green on `0ffa3ab` (`test (3.11–3.14)` and `Add up what the shards found`); re-running now. Local suite 9301 tests OK, and the two live-tmux files re-verified on 3.7c and the 3.2 floor after the rebase.

diazoxide:
## Green on the rebased head — `MERGEABLE CLEAN`

Head `cf38ece`, rebased onto `5de0794`. Re-swept ([33365052096](https://github.com/diazoxide/charter/actions/runs/33365052096)) and the result is identical to the one analysed above:

```
shards reporting: 3 of 3   |  59 mutations  |  50 pinned, 0 survivors, 9 unresolved
  shard 1: 20  {pinned 16, unresolved 4}
  shard 2: 20  {pinned 16, unresolved 4}
  shard 3: 19  {pinned 18, unresolved 1}
unresolved: instance.py 1021, 1022, 1023, 1527, 1535, 1555   (the six lines SHIPPED_LOOK reaches at import)
plan:       instance.py 35, chrome.py 15, commands_frame.py 7, panel.py 2   - all four are files this branch touches
```

All five required contexts pass — `test (3.11-3.14)` and `Add up what the shards found`. Branch protection is `strict: false`, so being two commits behind `main` (#771, #710 landed while this ran) does not block; the branch merges cleanly.

Local suite on the rebased tree: **9316 tests, OK, 1 skip**. Both live-tmux files re-verified after the rebase on 3.7c and at the 3.2 floor.

The `no verdict: 9 not measured` line is informational rather than required, and the nine are accounted for in the comment above: every one makes `charter/instance.py` raise at import, all three shards reported, and nothing timed out or lost a runner.

Issue #767, opened 2026-08-30
A wider `workspaces` bar can draw FEWER tabs, and collapses to the tab you are already on at 228 columns, when the current workspace sorts last


diazoxide:
#759 cuts the name list into pages and draws the page the marked name falls on. Every page
but the last is filled to the brim, so the **last page holds the remainder — and a remainder
shrinks as the pages grow**.

With the marked name sorting last, `_bar("workspaces", names, names[-1], width)` over 15
synthetic names, widths 60→280 in steps of 2, on 03fdbd3:

* **14 widths where widening the bar draws FEWER names**, including `152: 7→6`, `198: 4→3`,
  `228: 2→1`.
* At **228 columns the row is `workspaces  +14  *workspace-14`** — a single tab, the one the
  frame is already on, which `_Tabs.switch_to` correctly refuses. That is #758 returning at a
  width *wider* than the one that fixed it.

It misses this plane only because `harness-wrapper` sorts mid-list. **A workspace named
`zzz-scratch`, or any plane whose current workspace sorts last, gets the bad case.**

**Not covered by the monotonicity property #759 pinned.** That one asks whether the row is
drawn *at all* (`test_a_page_is_cut_inside_the_room_its_own_leading_count_has_left`), and with
the marked name last the row is always drawn — it just gets smaller. Measured: that test is
green while the count drops at 12 widths.

## What was measured before choosing a fix

Four alternatives, over 82 name lists at every room from 5 to 300. "Avoidable inert" counts
pages holding a single tab where some page held two or more.

| cut | avoidable inert | operator plane 100/120/160/200/240 | drops | min tabs ≥150 cols |
|---|---|---|---|---|
| greedy (shipped) | 4775 | 4, 6, 8, 10, 13 | 12 | **1 — inert** |
| equal-sized pages | 6969 | 4, 4, 5, 7, 7 | 0 | 1 |
| last page packed from the right | 4911 | 4, 6, 8, **6**, 12 | 0 | 2 |
| **floor of 2 on the last page** | **4175** | **4, 6, 8, 10, 13** | 10 | **2** |

Balancing every page is **worse**, not better: equal-sized pages ignore what names cost, so
short-name runs go under-packed. Packing the last page from the right removes the drops by
moving them onto a middle name.

A floor of two on the last page costs this plane nothing, removes the inert row, and keeps
the page a pure function of `(names, width)` — measured at **0 of 1,131,087** press/switch/press
triples where a second press at the same column switched again.

[...]

PR #771, opened 2026-08-30
The overflow line says which side the hidden rows are on, and the tab strip stops collapsing to one tab (#741, #767)

### LINKED ISSUE #741: The repo table's `…(+N more)` counts rows hidden above as if they were below, so a scrolled table promises rows that are not there
`…(+N more)` on the repo table counts every row the window is not showing, in both
directions, and reads as "there are N more below". Scrolled to the bottom of the list it
still promises rows that are not there.

## Reproduce

120x24, workspace with 8 clones, `[frame] mouse = true`. Repo pane holds 5 rows.

Unscrolled — correct:

```
▪ repos 8
  ├─ api                               main*
  ├─ auth                              main↑1
  ├─ billing                           main*
  ├─ gateway                           main
  ├─ ledger                            main
  …(+3 more, all clean)                        <- notify, search, web. True.
```

Wheel down six times, the window is at the end of the list:

```
▪ repos 8
  ├─ gateway                           main
  ├─ ledger                            main
  ├─ notify                            main
  ├─ search                            main
  ├─ web                               feat/new-landing-page-with-a-long…
  …(+3 more)                                   <- api, auth, billing. All ABOVE.
```

`web` is the last repo on the plane, there is nothing below it, and the wheel does nothing
from here — but the line under it says three more. The operator keeps scrolling into a
list that has ended.

Two smaller things ride along:

* There is no "above" indicator at all, so a scrolled table gives no sign that its first
  row is not the plane's first row.
* `web` draws `└─` when the whole table fits and `├─` when it is the last visible row with
  a `…(+3 more)` under it. Defensible on its own, and one more reason a scrolled table is
  hard to place.

## What a fix wants

The line is composed by `statusline._pick_rows`/`slots._table_lines` from a count of hidden
rows; the window offset is right there. Either split it (`…3 above`, `…2 below`) or keep
one line and say which way it points. The `, all clean` qualifier is worth keeping — it is
the reason the line is more than a count — and it should describe whichever set it names.

diazoxide:
Closes #741 and #767. Two surfaces, one rule: **an overflow marker that describes a window
has to say where the window is.**

## #741 — the repo table's count was right and its sentence was false

`hidden` is the complement of what was drawn, "true wherever the window is". The arithmetic
was never the bug. `more` is: it asserts *below*, the line sits at the bottom of the table,
and scrolled to the end it pointed underneath itself at nothing.

And it was the same sentence everywhere. Twelve repos in a five-row pane:

| offset | before | after |
|---|---|---|
| 0 | `…(+8 more, all clean)` | `…(8 below, all clean)` |
| 4 | `…(+8 more, all clean)` | `…(4 above, 4 below, all clean)` |
| 8 (bottom) | `…(+8 more, all clean)` | `…(8 above, all clean)` |

Three places to be standing, one string. Scrolling moved a number that never moved.

The table is one window over one list — repo rows then piece rows, the order its budget is
already spent in — so `offset` **is** what is above, and what is left of `hidden` is below.
Taken from the numbers the rows were sliced with, not re-derived.

**Not a marker row, not an arrow.** A row along the top costs a *repo row* out of a budget
counted in rows, at exactly the budgets short enough to scroll; the bottom line was already
reserved out of that budget, so two numbers on it cost nothing. `●`, `◆` and the pointing
triangles are East-Asian Ambiguous and have broken this layout twice (`_BAR_MARK`'s record).

## #767 — a wider tab bar could draw fewer tabs, and collapse to one

Found by the coordinator reviewing #759. Every page but the last is filled to the brim, so
the last holds the **remainder** — and a remainder shrinks as the pages grow. With the
marked name sorting last, 14 widths between 60 and 280 drew *fewer* names as the bar got
wider, and **228 columns collapsed to `+14  *workspace-14`**: one tab, the one the frame is
already on. #758 returning at a width wider than the one that fixed it. This plane escapes
it only because `harness-wrapper` sorts mid-list; `zzz-scratch` would not.

**Not covered by the monotonicity #759 pinned** — that asks whether the row is drawn *at
all*, and with the marked name last it always is. Measured: that case stays green while the
count drops at 12 widths. Different property, and it was unpinned.

Four cuts measured over 82 name lists at every room from 5 to 300. "Avoidable inert" counts
pages holding a single tab while some page held two or more:

| cut | avoidable inert | this plane 100/120/160/200/240 | drops | min tabs ≥150 cols |
|---|---|---|---|---|
| greedy (shipped) | 4775 | 4, 6, 8, 10, 13 | 12 | **1 — inert** |
| equal-sized pages | 6969 | 4, 4, 5, 7, 7 | 0 | 1 |
| last page from the right | 4911 | 4, 6, 8, **6**, 12 | 0 | 2 |
| **floor of 2 on the last page** | **4175** | **4, 6, 8, 10, 13** | 10 | **2** |

Balancing every page is **worse**, not better — equal-sized pages ignore what names cost, so
short-name runs go under-packed. Packing the last page from the right removes the drops by
moving them onto a middle name. A floor of two costs this plane nothing and removes the
inert row.

**The count is still not monotone — 10 drops, each of one name — and that is stated in the
ladder's docstring rather than left to be found.** Every cut that removes the rest was
measured worse on the harm that matters.

The coordinator's central question was whether a last-page merge breaks the column stability
that ruled out a centred window. **It does not**, and that is measured, not reasoned: any
`here`-independent repartition keeps it, at **0 of 1,131,087** press/switch/press triples
where a second press at the same column switched again.

## Verification

- Full suite green; real tmux on **3.7c and the 3.2 floor**, both surfaces.
- #741 swept over **1,414 (repos, pieces, budget, offset) shapes, 1,084 of them overflowing**
  — sides add to what is hidden, the leading count equals the offset, a side with nothing on
  it is never drawn. Three of four new cases go red on the old line.
- Two existing cases **pinned the defect** ("at the top and at the bottom alike", asserting
  the identical string at three offsets) and now assert the total holds while the sides move.
- `test_every_row_of_the_plane_is_on_screen_and_nothing_admits_otherwise` looked for `more)`,
  a word this change deletes — it would have passed against a line it had stopped being able
  to see. It asks for `…(` now.

### DIFF
--- a/charter/frame/slots.py
+++ b/charter/frame/slots.py
@@ -954,6 +954,27 @@ def _table_lines(data: dict, width: int, budget: int, *, offset: int = 0,
                   + pieces[:start] + pieces[start + len(kids):])
         quiet = not any(_needs_attention(r) for r in hidden)
         note = ", all clean" if quiet else ""
+        # **Which SIDE they are on, which `+N more` never said** (#741). The count was
+        # always right — `hidden` is the complement of what was drawn, true at every
+        # offset — but `more` asserts a direction the window has not earned. Driven at
+        # offsets 0, 4 and 8 over twelve repos in five rows, the old line read
+        # `…(+8 more, all clean)` all three times: eight below, four-and-four, and eight
+        # ABOVE are three different places to be standing and the row said one thing.
+        # Scrolled to the bottom it pointed under itself at nothing.
+        #
+        # The window is one window over one list — repo rows then piece rows, the order
+        # the budget above is spent in — so *offset* IS how many rows are above it and
+        # what is left of `hidden` is below. Taken from the numbers the rows were sliced
+        # with rather than re-derived: a second arithmetic for "where is the window" is a
+        # second answer, which is `_Line`'s own reason one column over.
+        above = offset
+        below = len(hidden) - above
+        # **Each count is drawn only for a side that HAS something on it** — `_bar`'s rule
+        # one axis over, where a leading `+3` beside a page starting mid-list is what stops
+        # a lone trailing count claiming the row holds the first names. `0 above` would be
+        # a field that is always false and always drawn.
+        where = ", ".join(f"{n} {side}" for n, side in
+                          ((above, "above"), (below, "below")) if n)
         # **About no repo, so a click here selects nothing.** It stands for the rows that
         # are NOT on screen, and picking one of them because the operator clicked the line
         # that says they exist would be charter answering a question nobody asked.
@@ -963,7 +984,7 @@ def _table_lines(data: dict, width: int, budget: int, *, offset: int = 0,
         # exclusive — they no longer are, and a "there is more below" line with three more
         # rows below it is the note pointing at the wrong place.
         lines.append(_Line(None, tui.truncate(
-            f"  {sl._DIM}…(+{len(hidden)} more{note}){sl._R}", width)))
+            f"  {sl._DIM}…({where}{note}){sl._R}", width)))
     return lines[:budget]
 
 
--- a/charter/frame/slots.py
+++ b/charter/frame/slots.py
@@ -2745,7 +2745,23 @@ def _page(fields: list[str], at: int, room: int) -> tuple[int, int]:
     # while that condition alone stops the loop 757 times. Dropping both leaves every one
     # of those 1,194,017 pages identical. An equivalent mutant and dead code are the same
     # finding, and this repository deletes rather than suppresses.
-    while len(cuts) >= 3 and cuts[-1] - cuts[-2] < 2:
+    #
+    # **A `len(cuts) >= 3` conjunct went the same way, and it is the more interesting of
+    # the three.** It was not unreachable: with a SINGLE field the cuts are `[0, 1]` and
+    # it is what stops the loop, 665 times in the sweep below. What it is not is
+    # OBSERVABLE. A one-name list reaches this rung only when that name did not fit the
+    # row, and the body composed from it is that same name — so `_bar`'s own measurement
+    # refuses the rung whatever this returns. That is the identical masking already
+    # recorded one function down for `if len(names) > 1`, which the sweep found for the
+    # same reason. Measured over 1,963,800 rows, keeping the conjunct and dropping it draw
+    # byte-identical output, with no runaway loop.
+    #
+    # The property it was informally protecting — that a page never starts left of zero,
+    # so no negative column can reach :data:`TABS` — is not left to a masked guard. It is
+    # asserted directly by `tests/test_frame_bars.AClickResolvesAgainstWhatWasDrawn
+    # .test_no_row_at_any_width_ever_draws_or_maps_a_column_left_of_zero`, which stays
+    # true if the measurement above ever stops masking it.
+    while cuts[-1] - cuts[-2] < 2:
         moved = cuts[-2] - 1
         if reach(moved) < cuts[-1]:
             break
--- a/charter/frame/slots.py
+++ b/charter/frame/slots.py
@@ -34,6 +34,7 @@ rather than a plausible number.
 
 from __future__ import annotations
 
+import bisect
 import os
 import time
 from typing import NamedTuple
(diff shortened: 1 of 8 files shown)


diazoxide:
## `2 survivors` — both dead guards, both deleted

```
[17/25] SURVIVED  charter/frame/slots.py:2701 [shift-boundary] reach(cuts[-3]) < moved
[25/25] SURVIVED  charter/frame/slots.py:2701 [shift-boundary] moved <= cuts[-3]
```

Same line — the three-way guard in #767's last-page rescue:

```python
if moved <= cuts[-3] or reach(cuts[-3]) < moved or reach(moved) < cuts[-1]:
    break
```

Neither is an equivalence to argue away. **Both disjuncts are dead**, and an equivalent
mutant and dead code are the same finding, so they are deleted rather than documented.

### Measured, not read

Each disjunct instrumented separately over **1,194,017 pages** — 304 name lists, every room
from 1 to 320, every name — recording which are true independently of `or` short-circuiting:

| disjunct | true | the only reason to break |
|---|---|---|
| `reach(cuts[-3]) < moved` | **0** | 0 |
| `moved <= cuts[-3]` | 11,804 | **0** |
| `reach(moved) < cuts[-1]` | 12,561 | **757** |

* `reach(cuts[-3]) < moved` **can never hold.** `cuts[-3]` does not move inside the loop, so
  `reach(cuts[-3])` is a constant — and it is exactly the greedy `cuts[-2]` the loop only ever
  walks *down* from. The comparison is `constant < something strictly below it`.
* `moved <= cuts[-3]` is true 11,804 times and is **never the only reason to stop**: the third
  condition is true at every one of them. It is also only ever true at exact equality, because
  the loop steps the boundary one name at a time.

Then the whole-output check: the guard reduced to `reach(moved) < cuts[-1]` alone, and each of
the sweep's two mutants, compared against the shipped guard across the same 1,194,017 pages —
**0 differing pages** for all three. That is why they survived, and why deleting them is safe.

### The suspicion about the floor constant is refuted

Reasonable suspicion given my last survivor, but measured false. `< 2` mutated to `< 1` and to
`< 3`, and `len(cuts) >= 3` to `> 3` — **each goes red**. The `1` puts back the collapse to the
tab you are already standing on; the `3` costs this plane its thirteenth tab at 240 columns.
The constant was pinned; its neighbours were not.

### The docstring's "10 drops" is now a test

Correct — it was described and asserted nowhere, which is a number written from a measurement
with nothing keeping it true. Both halves are pinned now, and the magnitude is the half that
matters: **no drop is ever more than one name**. A cut that started losing three at a time
would still satisfy "not monotone" and would be a different, worse thing. The count of ten
sits beside it so a change to the cut has to come back and restate what it costs, and
`_page`'s docstring now names that test.

### Rebased

Onto `c183fe0` — #774 touched `slots.py`, so this was re-checked rather than assumed. Clean
rebase, real tmux re-run green on **3.7c and the 3.2 floor** for both surfaces.

diazoxide:
## `1 survivor` — settled. It is mine, and it is the most interesting of the three

```
charter/frame/slots.py:2748 [drop-conjunct]
len(cuts) >= 3 and cuts[-1] - cuts[-2] < 2 — is the `len(cuts) >= 3` half pinned?
```

**Checked the charge first, per #776.** It is not a mis-charge: `slots.py` is a file this
branch writes, line 2748 is the rescue loop #767 added, and `git log -L` attributes it to
`fefo21` on this branch. Both sweep runs on the same head — the `workflow_dispatch` one and
the `pull_request` one the force-push fired — report the identical single survivor.

### Not unreachable. Unobservable. That is the difference worth naming

The previous two survivors were dead in the strict sense: one condition could never be true,
the other was never the sole reason to break. **This one is genuinely load-bearing for the
loop.** With a single field the cuts are `[0, 1]`, and `len(cuts) >= 3` is exactly what stops
the walk — measured, **665 times**, and in all 665 the other half of the `and` is true, so the
conjunct is doing the stopping.

What it cannot do is change a row. A one-name list reaches this rung only when that name did
not fit, and the body composed from the page **is that same name** — so `_bar` measures it and
refuses the rung whatever `_page` answered. That is the identical masking already recorded one
function down, where the sweep found `if len(names) > 1` for exactly this reason.

Measured over **1,963,800 rows** — 258 lists including singletons and names wider than any
pane, both headings, every width 0–299, every name:

| | |
|---|---|
| rows differing between keeping and dropping | **0** |
| runaway loops in the mutant | **0** |
| `len(cuts) < 3` reached | 665 |
| …of which the conjunct is what stops the loop | **665** |

Equivalent mutant and dead code are the same finding, so it is deleted rather than suppressed.

### The property it protected by accident is kept — as a test, not as a line

Without the conjunct the boundary steps to `-1` on a one-name list and `_page` answers a page
starting left of zero. A `+-1` field is unreadable, and a negative column in the strip is the
wrong-answer-hiding index `_Tabs` refuses on purpose. Today `_bar`'s width check masks it, and
a guard whose effect nothing can observe is not what should be holding that line.

`test_no_row_at_any_width_ever_draws_or_maps_a_column_left_of_zero` asserts it where it *can*
be seen — on the drawn row and on the published map, across singletons, over-wide names and
the fifteen-workspace list, at every width — so it stays true if that masking ever stops.

### Rebased again

`main` moved to `13b9a2b` while this was in flight. Rebased; it touches none of this branch's
files (`switch.py`, `commands_frame.py`, `sweep.yml` — the last a Dependabot cache bump, no
change to how the sweep charges). Re-verifying rather than assuming, and I will re-fire the
sweep on the new head and post the SHA.

Issue #714, opened 2026-08-30
Editing a running frame's arrangement duplicates every placed panel, and only the newest is recorded


diazoxide:
Adding two `[[frame.component]]` tables to a plane whose frame is already running left **six panel panes where there should be two**, and the frame's own state recorded only the newest of each.

## Measured, on the operator's live frame

```
%366 152x1 panel workspaces --session harness-wrapper.1
%367 152x1 panel workspaces --session harness-wrapper.1
%368 152x1 panel workspaces --session harness-wrapper.1
%369 152x1 panel chats      --session harness-wrapper.1
%370 152x1 panel chats      --session harness-wrapper.1
%371 152x1 panel chats      --session harness-wrapper.1

.charter/frame/harness-wrapper.1/panes:
  {"top":"%362","bottom":"%363","repos":"%364","right":"%365",
   "workspaces":"%368","chats":"%371"}
```

Three panes per component; one recorded. `charter.toml` was edited three times during that session — the tables added, reverted, and added again.

**The four orphans were still running**, still drawing, and each holding a ~24 MB panel process. Nothing reaps them: `record_panes` is rewritten on every relayout, so the moment a new pane id is recorded the previous one is unreachable to `_drop_panels` and to every other reader that goes through `state.panes`.

The operator's harness had been squeezed from 38 rows to 30 by panes it did not ask for and could not see the cause of.

## Why this is not "don't edit config while running"

- Nothing says so. `docs/frame.md` documents `[[frame.component]]` as the way to arrange a frame, and an operator arranging one naturally does it with the frame open — that is when they can see the result.
- **It is silent.** The duplicates draw correct content, so the frame looks *plausible*: the bar you asked for is there, three times, and nothing indicates which is real.
- **It is cumulative.** Every relayout after an arrangement change adds another set. There is no upper bound and no reaper.

## Where it goes wrong

`_relayout` splits a pane for a placed component whose recorded pane is missing, and records the new id. It does not ask whether **an unrecorded pane for that component already exists** — and after a config change the previously-recorded id has been overwritten, so the old pane is invisible to the check that would have found it.

The `@charter_panel` option (`_PANEL_OPTION`, from #634) already marks every pane charter split, so the panes are identifiable on the server independently of what the state file remembers. The reconciliation that is missing is *"which panes does this window already have, and which of them does the resolved arrangement still want"* — asked of tmux, not of the record.

## What a fix has to keep

[...]

PR #779, opened 2026-08-31
The top bar stops keeping a roster it drew before the frame's shape was on disk (#748)

### LINKED ISSUE #748: #530's roster duplication comes back on ~1 launch in 6: `_sidebar_live` reads `panes` before `record_panes` writes, and `top` never repaints
#530 stopped the top bar repeating the roster the sidebar already draws, by asking
`state.panes(fid)` live on every repaint. On roughly one launch in six the answer arrives
before `record_panes` has written, `top` draws the roster, and — because `top` is not in
`slots.ANIMATED` — it never repaints again. The duplication is then permanent for the life
of that frame.

## Measured

Eight launches at 120x40, tmux 3.7c, identical plane and identical command each time:

```
run 1: clean
run 2: clean
run 3: DUPLICATED roster
run 4: clean
run 5: clean
run 6: clean
```

plus two more duplicated in earlier ad-hoc launches. What a bad one looks like — the
roster on the top row and the same roster in the sidebar, both on screen:

```
 ⬢ alpha  ◆ steward · ◇ personas docs · forge · reddit · release · statusline    charter 0.54.0
────────────────────────────────────────────────────────────────────────────┬──────────────────
h$                                                                          │▪ personas 6
                                                                            │▸ steward
                                                                            │▫ docs ◦        ⚑
                                                                            │▫ forge ◦       ⚑
```

The record is correct by the time you look at it:

```
$ cat .charter/frame/alpha.1/panes
{"top": "%1", "bottom": "%2", "repos": "%3", "right": "%4"}
$ stat -f '%N %Sm' .charter/frame/alpha.1/{panes,version}
…/panes   01:56:36
…/version 01:56:36
```

Both written in the same second. When the launch gather's `state.bump` lands before
`record_panes`, `top`'s repaint reads a `panes` file that does not yet name `right`,
`_sidebar_live` answers False (its documented "safe direction"), and the roster is drawn.
Nothing bumps the version again, so that paint stands.

## Why the "safe direction" does not save it here

`_sidebar_live`'s docstring reasons about a *corrupt or absent* record, where False is
right because the alternative takes the plane's only roster off a screen with no sidebar.
That is a steady state. This is a startup race, where False is wrong and permanent — and
the panel has no reason to ask again.

## What a fix wants

Ordering, not a different default: `record_panes` should be written before anything can
bump the frame, so the first paint every panel does has the shape already recorded. The
launch gather is detached and deliberately concurrent with the frame coming up, which is
what makes the two writes race at all — so either the gather's bump waits on the panes
record, or the panels' first paint is deferred until it exists.

A second, cheaper safety net: `top` could repaint once on its first tick after start, the
way `resized["flag"] = True` already forces the first pass.

diazoxide:
Closes #748.

#530 stopped the top bar repeating the persona roster the sidebar already draws, by asking
`slots._sidebar_live` — `"right" in state.panes(fid)` — live on every repaint. The
duplication kept coming back on roughly one launch in six. This is why.

## The race

`commands_frame._draw_panels` writes that record, and it can only write it **after** every
`split-window`, because the pane ids in it are what those splits returned. A panel is a
`charter panel` **process** in a pane of its own, and `top` is the first pane split — so it
can be up and painting while the launcher is still carving `bottom`, `repos` and `right`.
When it is, `state.panes` answers `{}`, `_sidebar_live` answers its documented-safe
`False`, and the roster goes on the identity row.

That would be a flicker if anything told the panel to look again. Nothing did. `top` is not
in `slots.ANIMATED`: it repaints on a version bump and on nothing else, and the launch's own
`state.bump` happens *before* the first split.

**The pane's own paint history is the proof, and it is what makes this checkable.** Across
25 alternated launches on the real private-server path, with a real pty and a real
`tmux attach`, `top` on `main` painted **exactly once, every single time**:

```
                                        origin/main      this branch
launches that kept the roster            16 / 25            0 / 25
times the top bar painted at all          1 every time   2 on 22 of 25
median time to first content              0.385 s          0.390 s
```

So this is not "the race is likely" — it is "the state the race lands in is terminal".

## The fix

`_draw_panels` bumps the frame immediately after `state.record_panes`. Writing the shape
down becomes an event rather than a silent file, so a panel that painted before the record
has a reason to paint again, and its second paint reads the shape.

That is the ordering fix #748 asked for, but **neither of the two it proposed**:

- *defer the panels' first paint until the record exists* — puts a wait on the launch path,
  which #728 already measures as the slowest thing charter does;
- *make the gather's bump wait on the record* — puts the correction behind a detached child
  `_spawn_gather` is explicitly allowed to fail to spawn.

Bumping where the shape is written needs neither. It is the order `_apply_arrangement`
already keeps around the other call to `record_panes`, and the one `notify.plane_changed`
states outright: write, then bump, so a poller woken by the bump never reads the old record.

`_sidebar_live` itself is unchanged. #774 is right that reading tmux here would only move
which file the race is against.

## Measured

Forcing the race — the launcher's own `record_panes` write delayed so the panels lose every
time — with the pane's history read back verbatim:

```
origin/main                        this branch
⬢ alpha*  ◆ steward · ◇ perso…     ⬢ alpha*  ◆ steward · ◇ perso…
                                   ⬢ alpha*  ◆ steward
(one paint, 5 of 5)                (two paints, 5 of 5)
```

| rig | main | this branch |
|---|---|---|
| idle machine, operator's tmux, 90 launches | 16 (17.8%) | **0** |
| idle machine, private server, 15 launches | 2 | **0** |
| loaded, alternated, private server, 25 launches | 16 (64%) | **0** |
| forced race, operator's tmux, 20 launches | 20 | **0** |
| forced race, private server, 5 launches | 5 | **0** |

**61 standing duplications on `main`; 0 in 163 launches on this branch.**

Two findings worth recording:

- **The rate is load-dependent, and the filed 1-in-6 is the idle case.** Under load it is
  about two launches in three — which is what a control plane running several frames is.
  Alternating the two trees run-for-run is the only way the comparison holds; two batches
  taken back to back are not one.
- **The panel loses the race just as often on this branch** (19 of 25 launches drew the
  roster at some point, against 16 on `main`). That paint is honest — charter genuinely
  cannot tell yet. What changed is that it is now followed by a paint that can. On 2 of 5
  forced launches the correction landed before the client had even attached, so the
  duplicated row never reached the terminal at all.

## Cost

One `os.replace` of a version file and at most one extra repaint per panel, both spent
before any client is attached and before the frame's window is selected. It does not show
up in the time to first content (0.385 s → 0.390 s, within run-to-run noise).

## Tests

`tests/test_a_frame_repaints_once_its_shape_is_recorded.py`, five tests in two classes —
all five red on `main`:

- the launcher half: `_draw_panels` moves the version, and the bump lands **after** the
  write (a bump on the wrong side reads identically in a diff and closes nothing);
- the operator half, driven through the real `panel._tick` decision and the real
  `slots.render`: a `top` that painted before the record repaints, and the repaint has no
  roster — plus both directions, so `◆ steward` survives it and a frame whose launch
  recorded no sidebar still keeps its roster.

Full suite green: 9251 tests, 1 skip.

### DIFF
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -3368,6 +3368,23 @@ def _draw_panels(socket: str, *, slots: list[str], fid: str, harness_pane: str,
     # meant as which slot. Slots only: `state.record_harness_pane` already owns the
     # harness pane, and one fact recorded twice is one fact free to disagree with itself.
     state.record_panes(fid, panels=panes)
+    # And SAID so — #748. The panels above are processes, and the ids in that record are
+    # what their own `split-window` calls returned, so the record cannot be written until
+    # after the last of them exists and every one of them may already have painted. A
+    # panel repaints on a version bump and on nothing else (`panel._watch`; `top` is not
+    # in `slots.ANIMATED`), and this launch's own bump happened before the first split, so
+    # a shape written with nothing behind it is a shape the panels that lost the race
+    # never read: `slots._sidebar_live` answered `False` off an empty `state.panes`, `top`
+    # drew the roster the sidebar was already drawing, and it stood for the life of the
+    # frame. Measured on 90 real launches at 200x50 before this line: 16 of them.
+    #
+    # AFTER the write and never before it, which is `notify.plane_changed`'s order and its
+    # reason — a poller that saw the new version must never then read the old record —
+    # and the order `_apply_arrangement` already keeps around the OTHER call to
+    # `record_panes`, for this same question one keypress later. Charging the launch one
+    # `os.replace` and at most one
+    # repaint per panel, both spent before any client is attached or any window selected.
+    state.bump(fid)
     return panes
 
 
--- a/charter/frame/slots.py
+++ b/charter/frame/slots.py
@@ -164,14 +164,26 @@ def _sidebar_live(fid: str) -> bool:
     rather than the cost. A panel is a `charter panel` PROCESS in a pane of its own: it
     would be asking the server about a sibling that may not have been split yet. `right` is
     split last on the shipped frame, so `top`'s first paint can precede that pane
-    EXISTING — no reading of tmux, however authoritative, turns that into `True`. #748 is
-    that race, measured at roughly one launch in six, and its fix is an ordering one: get
-    the shape recorded (or the first paint deferred) before anything can bump the frame.
-    Reading tmux here would move which file the race is against and close nothing.
+    EXISTING — no reading of tmux, however authoritative, turns that into `True`. #748 was
+    that race, and reading tmux here would have moved which file it was against and closed
+    nothing.
+
+    **#748 is closed on the WRITING side, and this function is unchanged by it.**
+    Measured on real launches: 16 of 90 on an idle machine and 16 of 25 on a loaded one
+    drew the roster here and kept it, because the launch recorded the frame's shape
+    without moving the frame's version and `top` — not in :data:`ANIMATED` — had no reason
+    to ask a second time. The pane's own history is the proof rather than the rate: on
+    every one of those launches `top` painted exactly ONCE. Neither of the
+    two remedies #748 proposed shipped: deferring the first paint would have made every
+    panel wait on a file, and making the gather's bump wait on the record would have put
+    the correction behind a detached child that `_spawn_gather` is allowed to fail to
+    spawn. `commands_frame._draw_panels` bumps the frame immediately after
+    `state.record_panes` instead, so a paint that lost the race is followed by one that
+    reads the shape. A `False` here is now always either transient or true.
 
     **The cost is one small JSON read on a slot that is not animated.** `top` is not in
-    :data:`ANIMATED`, so it repaints on a version bump, never on `panel.TICK` — and a
-    density change bumps the version precisely so that surviving panels re-read, which is
+    :data:`ANIMATED`, so it repaints on a version bump, never on `panel.TICK` — and both
+    writers of this record bump the version precisely so that the panels re-read, which is
     exactly when this answer can have changed. It reads the same directory `_top` already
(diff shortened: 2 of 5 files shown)


diazoxide:
### #714's author flagged this against their own interest, and they were right

When #714 landed, its author noted that adding a `list-panes` to `_draw_panels` "shifts
every panel's start ~5 ms later relative to the gather's bump, a nudge on the wrong side of
a 1-in-6 race", and declined to claim it was harmless. That is now measured.

Same machine, same plane, same command, **alternated run for run** so both trees see the
same load — comparing the commit immediately before the #714 merge with the merge commit
itself, one step apart:

```
                                     69b59bf (before #714)   d5f4896 (#714)
launches that kept the roster              0 / 20              13 / 20
                                            (0%)                (65%)
```

and against current `main`, twenty more each:

```
                                     69b59bf (before #714)   5de0794 (main)
launches that kept the roster              1 / 20               8 / 20
                                            (5%)                (40%)
```

So #714 did not make a 1-in-6 race slightly likelier. It took a race that barely fired into
one that fires most of the time, and the issue's own 3-of-8 was measured on the tree that
already had it. That is the difference between "this is annoying" and "this is what the
frame does".

Nothing here is an argument against #714 — asking the window rather than the record is
right, and #774 restates why. It is an argument that the timing it perturbed was load-
bearing and nothing was watching it. This branch removes the dependence rather than
restoring the old margin: with the bump in place, the same rigs give 0 in 163 launches, and
where the race is *forced* the top bar simply paints twice instead of once.

Measurement rig, for anyone re-running it: one fresh tmux server per launch, the real
private-server path under a real pty with a real `tmux attach`, a frame counted as
duplicated only if the roster was still on the top bar after the frame's version stopped
moving, and the two trees alternated rather than batched — batching them is not a
comparison, because this race moves with machine load by a factor of three.

Issue #572, opened 2026-08-27
tools/sweep.py cannot sweep at all on macOS: the default workdir is a symlink, so the selection map comes back empty


diazoxide:
`python3 tools/sweep.py` refuses to run on macOS with its **default** workdir. It traces all 329 test modules, produces a map of **0 files and 0 broken modules**, and then correctly refuses:

```
  workdir: /var/folders/wl/hs3pdt2j0t9gyc_l73py38l80000gp/T/charter-sweep-855eddc2fb8a
sweeping ebeb62177349 (paths: charter)
  diff against 1ddd6b05f00b: 4 file(s), 384 added line(s)
  preparing the reference sandbox…
  selection map: tracing 329 test modules (7 at a time)…
    traced 320/329 modules
  ! the trace produced 0 file(s) and 0 broken module(s). The map is not usable:
RuntimeError: the selection map is empty — refusing to sweep blind
```

`0 broken modules` is the tell: every runner loaded its suite, ran it, and reported no error. The tracer simply matched nothing.

## Cause — measured

`build_map` computes `prefix = str(root) + os.sep` from the sandbox path, and `_TRACE_RUNNER` matches `code.co_filename.startswith(prefix)`.

The default workdir is `Path(tempfile.gettempdir()) / f"charter-sweep-{...}"`. On macOS `$TMPDIR` is `/var/folders/…`, and `/var` is a **symlink** to `/private/var`. The runner does `sys.path.insert(0, os.getcwd())`, and `os.getcwd()` returns the *resolved* path:

```
$ python3 -c "
import tempfile, os, pathlib
d = tempfile.mkdtemp(prefix='sym-')
print('mkdtemp   :', d)
print('resolved  :', str(pathlib.Path(d).resolve()))
os.chdir(d); print('getcwd    :', os.getcwd())"
mkdtemp   : /var/folders/wl/hs3pdt2j0t9gyc_l73py38l80000gp/T/sym-puy2r65o
resolved  : /private/var/folders/wl/hs3pdt2j0t9gyc_l73py38l80000gp/T/sym-puy2r65o
getcwd    : /private/var/folders/wl/hs3pdt2j0t9gyc_l73py38l80000gp/T/sym-puy2r65o
```

So every `co_filename` starts with `/private/var/folders/…` while `prefix` starts with `/var/folders/…`. `startswith` is never true, `seen` stays empty for all 329 modules, and the map is empty.

[...]

PR #790, opened 2026-08-31
The sweep charges you for your own lines, says when it measured nothing, and stops reporting its own runaway mutants as a dead runner (#776, #782, #773)

### LINKED ISSUE #773: A mutation can consume its runner, and the sweep reports that as infrastructure
`sweep.py` bounds what a mutation may spend in **time** and nothing else. A mutant that
allocates without bound exhausts its runner before either timeout can fire, the host kills
the runner, and GitHub reports:

```
##[error]The runner has received a shutdown signal. This can happen when the runner
service is stopped, or a manually started runner is canceled.
##[error]Process completed with exit code 143.
```

That is byte-for-byte what a **spot reclaim** looks like. From the report there is no way to
tell "the pool took my runner" from "my mutant ate my runner", so the gate charges it to
infrastructure and the shard's mutations are silently never measured.

#630 named *a mutation that hangs* as one of the ways a sweep lies, and `SUBSET_TIMEOUT`
(900 s) and `FULL_TIMEOUT` (2400 s) with a process-group kill are the answer to it. That
answer is correct and it does not cover this: **a timeout cannot save a runner that has
already been consumed.**

## Measured, on #710

Enumerating every mutation that branch offers — 118 of them, applied one at a time and run
under a 30 s cap — **four do not terminate**:

| line | operator | why it never ends |
|---|---|---|
| `while i < n:` | `shift-boundary` → `i <= n` | at `i == n`, `find` returns −1 so `i = n` again — and `lines.append(...)` every pass |
| `i = n if end < 0 else end + 1` | `collapse-ifexp` → `end + 1` | `end == −1` ⇒ `i = 0`, back to the start of the string, forever |
| (a second site, same shape) | `collapse-ifexp` → `end + 1` | as above |
| `if end < 0: return None` | `drop-if` | `j = end + 1 = 0`, rescans forever |

Three of the four append on every iteration. Rate, measured directly:

```
shift-boundary i<=n   5,134,576 iterations in 4.0s
                      -> ~252 MB retained, ~3,763 MB/min
```

**~3.8 GB/min exhausts a 16 GB runner in about four minutes** — against a 900 s subset
timeout. The timeout never gets to fire.

## The shard map says the same thing

Dealing `04bf8e6`'s 118-mutation plan round-robin the way `shard_of` does, and marking the
mutations that actually run away rather than the sites that could:

| shard | memory-exhausting mutations | outcome of run 33331151759 |
|---|---|---|
| 1 | 0 | success |
| **2** | **3** | **failure — "runner has received a shutdown signal"** |
| 3 | 0 (holds the CPU-only `drop-if`) | success |
| 4 | 0 | success |
| 5 | 0 | success |

Shard 2 is the only shard holding a memory-eater and the only shard that died. Shard 3 drew
the one that spins **without** allocating, and survived — which is the timeout working
exactly as designed, and is the control that makes this more than a coincidence.

## Why this matters beyond one branch

The failure is **silent and self-concealing**. The shard reports nothing, so `collect`
correctly says *"no verdict: N survivors so far, 1 of 5 shards did not report"* — the honest
answer, and the reason I could not read the run as clean. But nobody looking at that line
would suspect the branch's own mutants caused it, and the natural reading ("flaky pool,
re-run it") produces the same failure again on the same slice. It cost three runs here
before the shape was even suspected.

It also gets worse the better the tool gets: a scanner with index arithmetic is exactly what
`shift-boundary` and `collapse-ifexp` are for, and any parser, tokenizer or cursor loop has
the same shape. #710 is a character-by-character scanner; there will be more.

## What would close it

1. **Cap the child's address space.** The sandbox already runs the suite in its own process
   group (`_Sandbox.run`); adding `resource.setrlimit(RLIMIT_AS, …)` in a `preexec_fn`
   turns "runner dies, reported as infrastructure" into `MemoryError` in the child, which is
   a red, which is a **pin** — the correct verdict for a mutant that destroys the process.
2. **Failing that, report it.** If a shard's own resident set crosses a threshold while a
   mutation is running, name the mutation in the log before the host kills it. Even one line
   would have turned three runs of guesswork into a diagnosis.
3. **A fifth way a sweep lies, in the spec.** #630 lists the ways; this is one it does not
   have, and the distinguishing property is that the mutant takes the *measuring apparatus*
   with it rather than producing a wrong answer.

Filed as a hazard against the tool. It is not specific to #710 — that branch is only where
it happened to be measurable, and the same four operators over any cursor loop reproduce it.

### LINKED ISSUE #776: The sweep charges a PR for main's lines too, and the staleness multiplies #773's dead shards
The sweep charges a branch for lines it did not write, and the older the branch the more it
is charged. Measured on #710 today.

## The evidence, which is conclusive

The sweep swept merge commit `2ec368a8` against merge-base `f77ece96` and planned:

```
149 mutations across 8 file(s)
```

That branch touches **7 files, exactly one of them Python**. One of the survivors it reported:

```
SURVIVED  charter/frame/state.py:454 [narrow-except] OSError — is the catch pinned…
```

`charter/frame/state.py` is not a file that branch touches. It came from `main`, via commits
merged after the branch's merge-base.

## Why it happens

A PR's merge commit contains both the branch's changes *and* everything `main` gained since
the merge-base. Diffing `merge-base..merge-commit` therefore reports main's post-merge-base
lines as added by the branch. The correct charge is the branch's own diff — `merge-base..head`.

## Why it is worse than an inflated number

Three compounding effects, all measured on this one run:

1. **Foreign survivors.** A survivor in someone else's file lands on an author who cannot
   judge it, on a PR where it is noise. The whole value of the gate is that a survivor is
   *your* untested line.
2. **The plan grows, so the shard count grows** — here ~118 mutations over 5 shards became
   149 over 6.
3. **More shards means the runaway mutations are spread over more of them.** #773 records
   that a mutation can consume its runner (~3.8 GB/min against a 900 s timeout) and report
   `exit 143`, indistinguishable from a spot reclaim. On this run **three of six shards died**
   with the #773 signature, where the same branch on a correct plan had one of five. So the
   over-charge does not merely add noise, it multiplies the failure mode that produces
   unexplained no-verdicts.

The interaction is the point: each of these is tolerable alone, and together they turn a
stale branch into a sweep that cannot report.

## What a fix has to be careful about

- **`merge-base..head` is the intended charge**, but confirm that is what `DEFAULT_PATHS`
  and the gate's own "the guards this branch adds" wording already promise — if so this is a
  straightforward correctness fix; if the merge commit was deliberate, that reasoning needs
  to be written down, because nothing states it now.
- **Plan order and `shard_of` move when the plan shortens.** #754's verified property was
  that plan order does not move, and #755 was deliberately kept separate for exactly this
  reason. A fix here *will* re-deal every shard, so it should land alone and say so.
- Rebasing is the current workaround and it works, but it is invisible: nothing tells an
  author their survivor list contains another branch's lines. Until this is fixed, the gate
  should at least **name the file set it charged**, so a foreign path is visible.

### LINKED ISSUE #782: A sweep that planned zero mutations reports 'no survivors', which is the words for having checked
`no survivors: pass` is published for a branch on which **zero mutations were applied**. A gate
that measured nothing reports the same words as a gate that measured everything and found
nothing wrong.

## Measured

On #779, both on CI and locally:

```
no survivors: pass
mutations applied: 0
```

The branch's one added statement offered no mutation, so the plan was empty and the gate
passed. The author caught it and said so — *"the sweep verdict is vacuous"* — and supplied
real evidence by hand instead: deleting the added bump turns **5 tests red**, and moving it
before the record turns the ordering test red **alone**.

## This is #617 and #630, one level further in

The lineage is already written down in this repository:

* **#617** — a green check was not "no survivors". Fixed by #630 making the check's **name**
  carry the verdict, because GitHub's conclusion vocabulary cannot express it.
* **#646** — a gate that never ran looked identical to one that found nothing. Closed by #561
  making `Add up what the shards found` a **required** check, so a missing row blocks.
* **This** — a gate that ran, planned nothing, and says the words that mean "I checked."

Each previous fix removed one way for silence to read as success. This is the remaining one,
and it is the most misleading of the three, because the row is present, the job succeeded,
and the sentence is affirmative.

## Why it matters more than it looks

The empty plan is not a rare accident. It is the **normal** result for exactly the changes a
reviewer is least able to check by eye:

* a one-line statement that no operator mutates,
* a pure-documentation commit,
* a change whose only Python edit is inside a file the gate does not charge.

So the branches most likely to publish a vacuous `no survivors` are the ones where a human is
most likely to lean on it.

## The fix is a distinct verdict, not a failure

Do **not** make an empty plan fail. A change with nothing mutable is legitimate and common —
docs, config, a news entry — and failing it would train people to bypass the gate.

Make it *say what happened*, in the same place the other verdicts live — the check name:

```
no survivors                     N mutations, none survived
nothing to sweep                 0 mutations planned      <- new, and honest
no verdict: 2 of 5 shards ...    the run could not answer
```

`nothing to sweep` and `no survivors` are different facts and a reader needs to tell them
apart. One says the guards you added are tested; the other says you added nothing the tool
knows how to test — which on a security branch, or on a branch whose whole content is a
one-line ordering fix, is the more important sentence of the two.

## Note on scope

Also worth checking whether the count is available at the point the name is composed. #754
established that `plan` order and shard membership are settled before any shard runs, so the
plan length is known to the sizing job — which is where the distinction can be drawn without
waiting for a shard to report.

diazoxide:
Three things the deletion sweep said that were not true. One pull request, because they
interact: #776 lengthens the plan, a longer plan is more shards, and more shards spread
#773's runaway mutations across more of them.

Closes #776, closes #782, closes #773.

---

## #776 — it charged your branch for `main`'s lines

**The mechanism is not what the issue says, and the difference decides the fix.**

The issue reads as if the tool's scoping were wrong — "the correct charge is
`merge-base..HEAD`". The tool's scoping was already exactly that. `base_for`'s default is
the merge-base of the swept ref with `origin/main`, and for a merge commit **that is the
merge's own first parent**, because nothing later than the first parent is reachable from
the merge. It has always been right.

What was wrong is that `sweep.yml` overrode it with
`github.event.pull_request.base.sha`. That field is the payload's record of where the
branch started; it is **not** the tree `actions/checkout` puts on disk, and it lags. From
run 33331151759's own log, both lines from the same job:

```
HEAD is now at 58411a9 Merge 04bf8e6 into c29f3a8
diff against d40d998e06bd: 3 file(s), 458 added line(s)
```

`d40d998` is three `main` commits and three hours behind `c29f3a8`. So that branch — which
touches exactly one Python file — was swept for **three**, the extra two arriving from
`main`'s own commits, one of them a nine-file `charter save`.

**The fix is a deletion.** `plan` and `sweep` pass no `--base` at all and let the tool
resolve it; `plan` publishes the sha it charged as an output; `collect` takes its page
header from that output instead of from the payload. `tests/test_sweep.py` refuses
`pull_request.base.sha` by name in *any* `env:` of the file, so the one-line convenience
meets an assertion rather than a reviewer.

Reproduced as a real repository shaped like the tree CI checks out —
`ABranchIsNotChargedForWhatMainGainedWhileItWasOpen` — where the payload's base charges
three files and the tool's default charges one.

**This re-deals every shard**, and that is the reason it is worth saying out loud. `shard_of`
deals the plan round-robin, so a shorter plan puts different mutations on different
machines. Nothing is dropped and nothing is added — #754's property (order and membership
are settled before any shard runs, and identical across the plan and shard jobs) still
holds; membership simply moves once, here.

**A hazard that was checked and is not one.** Plan and shards resolve the base
independently, so they could in principle disagree. They cannot: `actions/checkout` fetches
`+<github.sha>:refs/remotes/pull/N/merge`, so every job in the run checks out the *same*
fixed merge commit, and the merge-base of a fixed merge commit with a fast-forward-only
`main` is invariant however far `main` moves mid-run.

---

## #782 — it said `no survivors` for a branch it never measured

`no survivors: pass` published beside `mutations applied: 0`. There is now a fourth verdict
beside the other three, in the same place they live — the check's name:

```
no survivors                 N mutations, none survived
nothing to sweep             0 mutations planned            <- new
no verdict: 2 of 5 shards …  the run could not answer
```

It **does not fail**, deliberately: docs, config, a news entry and a one-line reordering all
land here and failing them would train people around the gate. The run summary gains a
section saying in as many words that the page is not evidence the branch is tested, and the
terminal report says the same thing rather than "Every mutation this diff offered goes red"
over a plan of none.

`nothing to sweep` can never be worn by a run that lost a shard: `missing` is asked first
and still answers `no verdict`.

**The issue's suggested source for the count was not used, and using it would have been a
new lie of the same kind.** It proposed taking the plan length from the sizing job. That
output is the *empty string* when `plan` is cancelled, and an empty string read as zero
publishes `nothing to sweep` for a sweep that was superseded — exactly the shape #782 is
about. The merged result set answers the question exactly and cannot be empty for any other
reason: `Gate.measured == 0` with `missing == 0` is true if and only if no mutation reached
a sandbox.

Reproduced end to end at `test_a_diff_that_offers_no_mutation_is_not_a_branch_that_was_checked`
— #779's shape, a branch that adds `rows.append(1)`: a file changes, lines are added, the
scope is not empty, and no operator has anything to say about it.

---

## #773 — a mutant could eat the runner, and the report blamed the pool

Every run now gets an address-space cap: one share of the machine per sandbox and one left
over (`total // (jobs + 1)`), floored at the 2 GiB the suite was measured to need. A mutant
that crosses it raises `MemoryError` in the child — a red, on tests that were green
unmutated, which `decide` turns into a **pin**. The baseline runs under the same cap as the
mutants, because a cap the mutants have and the baseline does not could only ever
manufacture a red the baseline never had.

**The proposed delivery does not survive contact, and it fails toward "the tool does not
run".** `resource.setrlimit(RLIMIT_AS, …)` *raises* on macOS, and a `preexec_fn` that raises
surfaces as `SubprocessError` **in the parent** — so `RLIMIT_AS` via `preexec_fn`, written
as the issue proposes it, makes every mutation on the operator's own machine fail to run at
all. #572 is the issue about this tool being unusable on macOS. Second, `preexec_fn` calls
back into Python between `fork` and `exec`, which the standard library says is unsafe in the
presence of threads, and this sweep runs its sandboxes from a `ThreadPoolExecutor`.

Shipped instead as `/bin/sh -c 'ulimit -v "$1" || exit 89; shift; exec "$@"'` — thread-safe,
and `exec` keeps the pid the sweep holds, which `_kill_group`, the process group and both
timeouts all work through. `cap_holds()` measures once, in one process, whether the platform
takes a limit at all, and the sweep says which answer it got. macOS takes none, and says so.

**Measured, not assumed** (a Debian container, since the mechanism is Linux-only):

| what | result |
|---|---|
| whole suite, uncapped | 9,581 tests, peak tree RSS **253 MB**, peak tree VSZ **1,580 MB** |
| whole suite under a runner-sized cap (`ulimit -v 3355443`, = 3.2 GiB) | same failures, same skips — the cap changes nothing legitimate |
| a mutant allocating without bound, capped at 512 MiB | red, `conclusive`, one named failing test — a **pin** |
| the same mutant, cap removed | the child is killed with no verdict at all — #773's own signature |

The last two are `AMutantThatEatsTheMachineIsARedAndNotAnOutage`, which runs for real on
Linux (all four CI Pythons) and skips loudly on macOS.

**A static detector was not built, and the issue is right that it would fail open** — the
fastest of #710's four runaways leaves every loop inside its own function terminating and
corrupts a cursor returned *across* a function boundary, so the caller is what spins.

`docs/superpowers/specs/2026-08-27-deletion-sweep-harness.md` now lists six ways a sweep
lies rather than five, with this one's distinguishing property written down: it takes the
measuring apparatus with it rather than producing a wrong answer.

---

## Checks

* Every new case was run with its guard removed and confirmed red — including the two
  Linux-only ones, checked in a container.
* `tools/sweep.py --gate --path tools` on this branch: the sweep's own verdict on its own
  new code, reported in the pull request thread.

### DIFF
--- a/.github/workflows/sweep.yml
+++ b/.github/workflows/sweep.yml
@@ -18,11 +18,12 @@ name: deletion sweep
 # repository rather than read out of the documentation:
 #
 #   * **The two events sweep different trees.** On `pull_request`, `github.sha` is the merge
-#     commit and `--base` is the pull request's base sha — the tree that will land. A `push`
-#     run has neither: `BASE_SHA` is empty, so `sweep.py` takes its own default of the
-#     merge-base with `origin/main` and charges the branch TIP. Both answers are defensible.
-#     They are not the same answer, and under `push:` they would arrive on the same pull
-#     request under the same check names.
+#     commit — the tree that will land — and the sweep charges it against its merge-base
+#     with `origin/main`, which for a merge commit is the base branch tip it was merged
+#     into. A `push` run has no merge commit: the same default charges the branch TIP
+#     against where it was cut from. Both answers are defensible. They are not the same
+#     answer, and under `push:` they would arrive on the same pull request under the same
+#     check names.
 #   * **That is what makes it worse rather than better.** The remedy #646 settles on is to
 #     require this workflow's `collect` job, so that a missing row BLOCKS instead of passing
 #     silently. The moment `push:` can also produce that check, its presence stops proving
@@ -102,6 +103,11 @@ jobs:
       mutations: ${{ steps.size.outputs.mutations }}
       shards: ${{ steps.size.outputs.shards }}
       matrix: ${{ steps.size.outputs.matrix }}
+      # The sha the mutations were actually read against, resolved by the tool from the
+      # tree it swept. The merge step runs at `fetch-depth: 1` on its own machine, where
+      # neither `HEAD^` nor `origin/main` is an object it has, so it cannot work this out
+      # and must be told — and the value it used to be told was the one #776 is about.
+      base: ${{ steps.size.outputs.base }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
@@ -129,25 +135,39 @@ jobs:
         with:
           path: ${{ runner.temp }}/sweep/cache
           key: sweep-map-${{ hashFiles('charter/**/*.py', 'tests/**/*.py') }}
+      # **No `--base`, and `github.event.pull_request.base.sha` is refused by name (#776).**
+      #
+      # That field is the payload's idea of where the branch started, and it is not the
+      # tree `actions/checkout` just put on disk. Measured, from run 33331151759's own log
+      # rather than from documentation:
+      #
+      #     HEAD is now at 58411a9 Merge 04bf8e6 into c29f3a8
+      #     diff against d40d998e06bd: 3 file(s), 458 added line(s)
+      #
+      # The checkout merged the branch into `c29f3a8`; the payload said the base was
+      # `d40d998`, three commits and three hours earlier. So the sweep charged that branch
+      # for `d40d998..c29f3a8` — main's own commits, one of them a nine-file `charter
+      # save` — and reported **3 files** for a branch that touches exactly one Python file.
+      # A survivor in a file the author has never opened destroys the one premise this
+      # gate runs on, which is that a survivor is YOUR untested line.
+      #
+      # The tool's own default is already the right answer and always was: the merge-base
+      # of the checked-out merge commit with `origin/main` **is** that merge's first
+      # parent, because nothing later than the first parent is reachable from the merge.
+      # `fetch-depth: 0` above fetches `+refs/heads/*`, so `origin/main` is there to ask.
+      # On a manual dispatch there is no merge commit and no base sha either, and the same
+      # default gives the branch tip's merge-base — which is what that run wants.
+      #
+      # The over-charge also multiplies #773: a longer plan is more shards, and more
+      # shards spread the runaway mutations across more of them. Three of six shards died
+      # on the over-charged plan where the same branch on a correct one lost one of five.
       - name: How many mutations, and how many jobs they need
         id: size
-        env:
-          # Through the environment and not interpolated into the script. A `${{ }}` inside
-          # `run:` is textual substitution before bash ever sees it, which is how a value
-          # somebody else controls becomes a command; the same rule `release.yml` follows.
-          BASE_SHA: ${{ github.event.pull_request.base.sha }}
         run: |
           set -euo pipefail
-          # No --base on a manual dispatch: with no pull request there is no base sha, and
-          # the tool's own default (the merge-base with origin/main) is the right answer.
-          if [ -n "${BASE_SHA:-}" ]; then
-            set -- --base "$BASE_SHA"
-          else
-            set --
-          fi
           python3 tools/sweep.py --plan --warm-map --jobs 4 \
             --workdir "$RUNNER_TEMP/sweep" \
-            --github-output "$GITHUB_OUTPUT" "$@"
+            --github-output "$GITHUB_OUTPUT"
 
   # ------------------------------------------------------------------------------------
   # 2. The sweep itself, on as many machines as the plan asked for.
@@ -183,22 +203,23 @@ jobs:
       # answer and must never write a page that looks like the answer: five partial
       # tables on one pull request is the same defect #617 is about, arriving five times.
       # The merge step below is the only thing that tells this branch anything.
+      # No `--base` here either, and for a stronger reason than in `plan`: a shard's
+      # slice is `shard_of` over the whole plan, so a shard that charged a different base
+      # would deal a different plan and sweep a slice nobody asked for — while the merge
+      # step added the pieces up and reported a complete sweep. The two jobs agree because
+      # they ask the same question of the same tree: `github.sha` is fixed for the run, so
+      # both check out the same merge commit, and its merge-base with a fast-forward-only
+      # `main` cannot move while the run is going.
       - name: Sweep the guards this branch adds
         env:
-          BASE_SHA: ${{ github.event.pull_request.base.sha }}
           SHARD: ${{ matrix.shard }}
           SHARDS: ${{ needs.plan.outputs.shards }}
         run: |
           set -euo pipefail
-          if [ -n "${BASE_SHA:-}" ]; then
-            set -- --base "$BASE_SHA"
-          else
-            set --
-          fi
           python3 tools/sweep.py --gate --jobs 4 \
             --shard "$SHARD/$SHARDS" \
             --workdir "$RUNNER_TEMP/sweep" \
-            --json "sweep-results-$SHARD.json" "$@"
+            --json "sweep-results-$SHARD.json"
       - name: Keep the result set, whatever the sweep said
         if: always()
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
@@ -302,7 +323,12 @@ jobs:
           # is the tree they actually swept. Naming the branch head on a page about the
           # merge result would be one sha out on every run with a moving base.
           SWEPT_SHA: ${{ github.sha }}
-          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          # From the job that resolved it, and never from the pull request payload (#776).
+          # This step is a page header away from the sweep itself, and a header naming a
+          # base the mutations were not read against is the quietest way for the two to
+          # drift apart again. Empty when `plan` did not finish, which `--base` renders as
+          # "merge-base" rather than as a sha nobody charged.
+          BASE_SHA: ${{ needs.plan.outputs.base }}
         run: |
           set -euo pipefail
           # `--enforce` is the one flag that makes any of this block a merge, and it is
--- a/docs/news/unreleased-the-sweep-stops-reporting-things-that-are-not-true.md
+++ b/docs/news/unreleased-the-sweep-stops-reporting-things-that-are-not-true.md
@@ -1,0 +1,115 @@
+---
+version: unreleased
+headline: The deletion sweep charges you for your own lines, says when it measured nothing, and stops reporting its own runaway mutants as a dead runner
+---
+
+Three things the sweep said that were not true. They are one change because they interact:
+the first makes the third worse, and the second is the one that reads best while saying
+least.
+
+## It charged your branch for `main`'s lines (#776)
+
+The workflow told the tool where the branch started by passing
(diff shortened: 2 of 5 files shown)


diazoxide:
## The three fixes, verified on this pull request's own run

**#776 — the base is resolved from the tree, not from the payload.** `Size the sweep` on
run 33386163073, two lines from the same job:

```
HEAD is now at 40aaf84 Merge d5dbddd7387b620076628c23168c34956ab02c54 into c140c463b2d30adf76d5da01e507bcbf10cdf2ed
diff against c140c463b2d3: 0 file(s), 0 added line(s)
```

The sha it charged is exactly the merge commit's first parent, worked out by the tool with
no `--base` passed to it at all. That is the property the fix rests on, live.

**#782 — the check on this pull request reads `nothing to sweep`.** This branch touches
nothing under `charter/`, so the plan is empty, and the row a reviewer sees is:

```
nothing to sweep    pass
Add up what the shards found    pass
```

Before this change that row said `no survivors` — over zero mutations. The gate now says
which of the two happened, and it still blocks nothing.

**#773 — the cap is exercised for real on every CI Python.** `AMutantThatEatsTheMachineIsARedAndNotAnOutage`
runs on Linux and skips loudly on macOS; `test (3.11)`, `(3.12)`, `(3.13)` and `(3.14)` are
all green, which means the runaway-mutant case ran four times and came back a red on a named
test rather than a dead process.

Measured separately in a Debian container, since the mechanism is Linux-only:

| | |
|---|---|
| whole suite, uncapped, on `main` | 9,581 tests — 5 failures (all container-as-root artefacts) |
| whole suite on this branch under a runner-sized cap (`ulimit -v 3355443`, 3.2 GiB) | 9,608 tests — the same 5, by name |
| a mutant allocating without bound, capped at 512 MiB | red, conclusive, one named failing test — a **pin** |
| the same mutant with `under_cap` removed | the child dies with no verdict at all — #773's own signature |

Peak of the whole process tree, uncapped, on ten cores: **253 MB resident, 1,580 MB of
address space**. `SUITE_ADDRESS_SPACE` is 2 GiB, above that figure, and the cap on a 16 GB
runner at `--jobs 4` is 3.2 GiB.

Issue #788, opened 2026-08-31
After a workspace switch, the chats bar offers rows that cmd_chat refuses every time, and the pre-flight approves them


diazoxide:
After `F2 → workspace`, the chats bar draws the target workspace's other chats as tabs, the
bar's own pre-flight approves them, and pressing one is then refused every time.

## The sequence

1. `switch.to_workspace(fid, "gamma")` re-points chat `alpha.1`'s record to `gamma`
   (`charter/frame/switch.py:196`). The tmux window does not move — charter calls no
   `move-window` anywhere.
2. `chats.roster(fid)` correctly lists `gamma`'s chats, because `fid` now says it is in
   `gamma`.
3. `chats.check` (`charter/frame/chats.py:305`) approves those rows — same server, valid
   pane record.
4. `cmd_chat` (`charter/commands_frame.py:6166`) refuses **every one** of them:

   > `cannot switch: chat 'gamma.1' is a window of another tmux session, so selecting it
   > would move that session and leave this client here`

So the bar offers rows that are guaranteed to be rejected, and the pre-flight that exists to
prevent exactly that says they are fine.

## Why it is worth its own issue

`#733` records the **backward** direction of this strand — the moved chat's siblings can never
see it again. This is the **forward** direction: the moved chat can see its new neighbours,
is told they are reachable, and cannot reach any of them. Neither is a subset of the other,
and only one of them is written down.

It is also the sharpest available demonstration that the three surfaces disagree:

* `roster` answers *"which chats are in my workspace"* — from the record;
* `check` answers *"can I switch to this"* — from the server and the pane;
* `cmd_chat` answers the same question again — from tmux's session identity.

Two say yes and the third says no, and the third is right. A pre-flight that approves what
the action refuses is worse than no pre-flight, because it moves the failure from the moment
of choosing to the moment of pressing.

## What fixing the cause does to this

The settled design (`§4j`, *"a chat belongs to its workspace for life"*, spec
`2026-08-25-agentic-ide-foundation.md:661`) makes step 1 impossible, and this defect cannot
form. So this issue should close *with* that work rather than being fixed on its own —
patching `check` to agree with `cmd_chat` would hide the disagreement while leaving the
re-pointing that causes it.

**But it should stay filed**, because it is the test worth keeping afterwards: no surface may
offer a chat row that `cmd_chat` would refuse, and that property is checkable independently of
how the switch is implemented.

Issue #785, opened 2026-08-31
A suite run from a worktree asserts against the parent clone's live plane, so tests can pass or fail for reasons the branch never touched


diazoxide:
A test run from a git worktree reads the **parent clone's live plane**, not the worktree's own
committed one. So a suite that passes on CI can fail locally for reasons that have nothing to
do with the branch, and — worse — a suite that passes locally may be asserting against an
operator's uncommitted configuration.

## Measured

`charter/root.py` deliberately redirects a worktree's plane to the clone it was cut from. Its
own comment says so:

> `_plane_of` handles the case where the worktree HAS a `charter.toml` — it redirects

Consequence, measured from a **pristine detached worktree of `origin/main`**:

* `config.ROOT` resolves to `/Users/aharon/IdeaProjects/charter` — the operator's live plane;
* `test_frame_border_surface.test_this_planes_own_committed_frame_answers_the_report` fails,
  because that plane's `[frame]` block is not the committed one (6 components, `chrome=dark`
  giving `bg=black`, where the test wants `brightblack`);
* the same test **passes on CI**, where the checkout has no parent plane to be redirected to.

It is not caused by any branch. It reproduces on a clean worktree of `main`.

## Why this is worth fixing rather than working around

The redirect is almost certainly correct for the *product*: a worktree of your plane is still
your plane, and charter should not treat `charter clone`'s worktrees as separate control
planes. The defect is that **the test suite inherits a product behaviour that makes its own
assertions non-hermetic.**

Two failure directions, and the second is the dangerous one:

1. **False red.** A test asserting "this repository's committed `charter.toml` says X" fails in
   any worktree whose parent plane says Y. Every agent working in a worktree hits this, and
   the failure names a file the branch never touched — so the first reading is always "did I
   break this?"
2. **False green.** The same mechanism can make a test pass because the *operator's* live
   configuration happens to satisfy it. That failure is silent, survives review, and is the
   shape this repository has spent a long time learning to distrust: a test that is not
   measuring what its name says.

There is a third, quieter cost: it makes local suite runs depend on uncommitted state, so two
agents on the same machine can get different results from the same tree.

## What a fix has to decide

* **Whether the redirect should apply when charter is being TESTED.** `tests/_planeguard.py`
  and `tests/_isolation.py` already exist to keep the suite off the real plane; this is a hole
  in that boundary rather than a new problem.
* **Which tests legitimately want the committed file.** `test_this_planes_own_committed_frame_
[...]

PR #794, opened 2026-09-01
Typing a workspace command inside a chat no longer moves the chat (#791)

### LINKED ISSUE #731: A recycled chat id inherits the previous frame's workspace pointer and lock, so `--workspace alpha` can come up labelled `gamma`

### LINKED ISSUE #791: charter workspace use still re-homes a chat, because the per-session pointer is a membership rung
`#789` removed the two writes in `switch.to_workspace`, so `F2 → workspace` can no longer
re-home a chat. **The same defect is still reachable through `charter workspace use`**, and
`#733`/`#788` closed without covering it because both named `switch.to_workspace` as the cause.

## Measured, on `#789`'s own branch, three chat dirs, no tmux

```
--- F2 -> workspace -> gamma, inside alpha.1        (fixed by #789)
  outcome: ok=False, 'a chat belongs to its workspace for life — …'
  alpha.1 draws: alpha    alpha.2 sees: ['alpha.1']

--- `charter workspace use gamma`, typed inside alpha.1   (STILL BROKEN)
  alpha.1 draws: gamma    alpha.1 owns: gamma    frame record: alpha
  alpha.2 sees: []        alpha.1 sees: ['gamma.1']
  check alpha.2 -> alpha.1: no chat 'alpha.1' here — have: alpha.2
  check alpha.1 -> gamma.1: chat → gamma.1     <- and cmd_chat then refuses it
```

The last two lines are `#733` and `#788` verbatim.

## Why it still reaches

`commands_workspace.py` writes `workspace.set_active(..., session_id=session.current())`.
Inside a frame `session.current()` is **the chat id** (ADR 0019), so the write lands on
`workspace.for_session(<chat>)` — which `e590566` made the **middle rung of
`state.own_workspace`**, the membership ladder. A command about "which workspace is this
session working in" therefore silently answers "which workspace does this chat belong to".

## The fork

**Option A — refuse `workspace use` inside a frame.** Rejected once already, on evidence: the
only available test for "inside a frame" is `session.current()`, which reads
`$CHARTER_SESSION_ID`, and **every agent spawned from a frame inherits it** (measured:
a subagent shell three levels down still reports `CHARTER_SESSION_ID=harness-wrapper.1`). The
refusal would fire on agents doing ordinary CLI work in isolated worktrees, none of which is
in a frame in any meaningful sense.

**Option B — drop the per-session pointer rung from `own_workspace`.** Membership becomes the
launch record alone. `workspace use` keeps writing the pointer and keeps working everywhere,
for every CLI purpose; it simply stops deciding a chat's identity.

**Recommended: B.** It needs no detection, breaks no command, and is what `§4j` already says —
*"`{workspace}-{hash}` is identity, not a property."* A pointer written by whoever last typed
a command is the definition of a property, and identity cannot be one.

## What B costs, stated rather than discovered

`docs/frame.md:767` promises *"`charter workspace use <name>` typed at the agent moves the
panels too"*. Under B that becomes false and the line must be corrected — the same correction
`#789` deliberately did **not** make, because at that point the write that made it true was
the one still standing.

Anyone taking this should also re-check `#752`: a chat with a session pointer follows a
workspace rename today, and one without is orphaned. Dropping the rung makes every chat behave
like the second kind. That is more consistent, and it is a behaviour change worth naming.

diazoxide:
Closes #791. **Does NOT close #731** — measured; see below.

## The defect

`#789` removed `switch.to_workspace`'s two writes, so `F2 → workspace` can no longer
re-home a chat. **The same defect was still reachable by typing a command**, and #733/#788
closed without covering it because both named the switcher as the cause.

`commands_workspace` calls `workspace.set_active(...)`, which keys its pointer on
`session.current()`. Inside a frame `session.current()` is **the chat's id** (ADR 0019), so
the write landed on `workspace.for_session(<chat>)` — which `e590566` had made the **middle
rung of `state.own_workspace`**, the membership ladder. A command whose subject is "which
workspace is this session working in" was also answering "which workspace does this chat
belong to".

Measured on three chat directories, no tmux, `charter workspace use gamma` typed inside
`alpha.1`:

```
alpha.1 draws: gamma    alpha.1 owns: gamma    frame record: alpha
alpha.2 sees: []        alpha.1 sees: ['gamma.1']
check alpha.2 -> alpha.1: no chat 'alpha.1' here — have: alpha.2
check alpha.1 -> gamma.1: chat → gamma.1     <- and cmd_chat then refuses it
```

The last two lines are #733 and #788 verbatim. After this change, same fixture:

```
alpha.1 draws: alpha    alpha.1 owns: alpha    frame record: alpha
alpha.2 sees: ['alpha.1']   alpha.1 sees: ['alpha.2']
check alpha.2 -> alpha.1: True   check alpha.1 -> gamma.1: False
of_workspace alpha: ['alpha.1', 'alpha.2']   of_workspace gamma: ['gamma.1']
```

## What the ladder looks like now

`state.own_workspace` — the membership question — loses its middle rung:

| | before | after |
|---|---|---|
| 1 | the pin the launcher recorded (`state.identity`) | the pin the launcher recorded |
| 2 | `workspace.for_session(<chat>)` — the per-session pointer | *(gone)* |
| 3 | `state.frame_workspace` — what the launch resolved | what the launch resolved |

Both remaining rungs are written by the **launch**, so nothing typed after a launch moves a
chat between workspaces — which is §4j: *"`{workspace}-{hash}` is identity, not a
property."* `state.workspace_for` is unchanged in shape (rung 0 `$CHARTER_WORKSPACE`, then
`own_workspace`, then a local resolve), so the roster and the panels still walk **one**
ladder — #733's actual fix — it is just one rung shorter.

**Refusing `workspace use` inside a frame was the alternative and it is rejected on
evidence.** The only available test for "inside a frame" is `session.current()`, which reads
`$CHARTER_SESSION_ID`, and **every agent spawned from a frame inherits it** (measured: a
subagent shell three levels down still reports the frame's id). A refusal keyed on that
fires on agents doing ordinary CLI work in isolated worktrees. Dropping the rung needs no
detection and breaks no command: `set_active` still writes the pointer and the lock,
`workspace.resolve` still reads them, so `charter clone`, `charter repos` and
`charter ws current` in the frame's own shell all act on the name that was typed. A test
pins that explicitly, so a later "fix" cannot satisfy this by deleting the write.

## Two behaviour changes, stated rather than left to be found

**1. `charter workspace use <name>` no longer moves a frame's panels.** `docs/frame.md:804`
promised it did; the promise is corrected in this PR rather than quietly broken. #789
deliberately did not correct it because at that point the write making it true was the one
still standing. `F2 → workspace` moves your terminal instead and leaves every chat where it
is. One frame still follows the pointer and it is the migration case — a frame that recorded
no workspace at all, where the panels fall through to resolving for themselves;
`tests/test_frame_owns_the_surface` is renamed for that and its new last case pins the
boundary. ADR 0019's illustrative claim ("panels follow `ws use` *only* because of the id
collision") is corrected too; the decision itself is unchanged and still load-bearing for
the persona, the lock and the harness-session mapping.

**2. A workspace rename now orphans every chat, not just some of them.**
`workspace._rename_active_pointers` rewrites every per-session pointer holding the old name,
so a chat that had one followed a rename automatically while a chat that never had one was
left naming a workspace that no longer exists (#752's neighbourhood). Membership no longer
reads that pointer, so **every** chat behaves like the second kind — more uniform, and a real
change. The repair route went with it: typing `charter workspace use <newname>` inside an
orphaned chat used to fix it and now does nothing. Making `rename` repoint the frame records
is the fix and it is deliberately **not** in here — whether a rename moves a chat's identity
is a decision that deserves its own argument beside #752, not a line in a bugfix.

## Does this close #731? No — and which half survives

#731 is that `.charter/sessions/<fid>.workspace` and `.lock` are not reaped with the frame,
so a recycled chat ordinal inherits the previous frame's pointer *and outranks its own
launcher*. Measured both ways on the same fixture (write the pointer, `rmtree` the frame
directory, relaunch onto the same ordinal):

**Gone — the half the issue's title is about.** Before: the relaunch that asked for `alpha`
explicitly drew `gamma`, joined `gamma`'s roster and left `alpha`'s — which is the identity
row, the todo count and the repo pane height in the issue's screenshot. After:

```
relaunched alpha.1 draws: alpha   owns: alpha   frame record: alpha
of_workspace alpha: ['alpha.1', 'alpha.2']   of_workspace gamma: ['gamma.1']
```

**Survives — the lock and the frame shell's own resolution**, neither of which this touches:

```
is_locked: gamma
shell resolve (cwd=plane root): gamma
set_active('alpha') now returns: locked
```

So the operator who just typed `--workspace alpha` gets a frame drawing `alpha` whose
commands act on `gamma` and which **refuses** `charter workspace use alpha` as locked. #731
is now exactly *"reap the session pointer and the lock with the frame"*, which is what PR
#757's body already recommended as the right fix. Leaving it open.

## This is a prerequisite for quit → reopen (spec §4e/§4f)

A reopened chat relaunches into its **existing** directory, so before this change a stale
`.charter/sessions/<fid>.workspace` would have decided its membership over whatever the
reopen manifest said — silently, and only for chats somebody had once typed
`workspace use` in. **This PR unblocks that work**, and one of the new cases is written as
that scenario (`test_a_pointer_left_behind_by_a_reaped_chat_decides_nothing`).

## Four tests near the change had stopped measuring their own guards

All four are restaged rather than left green — three had leaned on the pointer being the
disagreeing rung, and with the pointer gone they passed whichever way the guard went:

- the recorded pin's `.strip()` (`test_a_recorded_pin_with_a_TRAILING_space_is_still_that_pin`) — the padded pin now names a *different* workspace from the launch record
- `workspace_for`'s rung 0, twice (`test_frame_state.ThePinOutranksTheFramesOwnRungs`, `test_frame_slots.TheChipAndItsStarNameTheSameWorkspace`) — the launch record now names a third workspace
- `of_workspace`'s `is_dir()` filter — **no measurement left at all.** #733 restored that
  line on the grounds that membership had grown a rung outside the frame's directory; #791
  removes that rung, so a loose file is refused by the read whatever the filter does
  (checked: nothing in the suite goes red without it). It stays, and its comment now says
  plainly that it is a floor rather than a measured guard, so the next deletion sweep decides
  on the facts instead of on a stale paragraph.

## Verification

- **Reproduced without tmux first**, through the real `workspace.set_active` with the chat's
  id in the environment and no explicit `session_id=` — the defect is *which id the command
  resolves for itself*, so passing one would remove the thing under test.
- **7 new cases** in `tests/test_a_chat_belongs_to_its_workspace_for_life.py`
  (`TheOtherDoorMovesNoChatEither`), **5 of them red before the change**; the two that were
  green are the ones asserting what must NOT change (the pointer is still written; the pin
  still wins).
- **Full suite: 9761 tests, OK** in the sweep's clean sandbox. Locally one failure,
  `test_frame_border_surface.test_this_planes_own_committed_frame_answers_the_report`, which
  is #785 — it reads the operator's uncommitted `charter.toml` through `$CHARTER_ROOT`;
  reproduced at `9cd519e` with this branch's changes stashed, and green in the sandbox that
  has no such env.
- **Real tmux: 177 tests, 0 skipped, OK** (`test_frame_tmux_integration`,
  `test_a_real_click_on_a_real_tab_bar_switches`, `test_a_workspace_switch_moves_the_client`,
  `test_a_second_launch_focuses_instead_of_dragging`), on the tests' own sockets.
- **Deletion sweep: `NOTHING TO SWEEP` — 0 mutations applied**, and per #782 that is the
  count to read rather than the verdict: the sweep's own words are *"a record of a question
  nobody asked rather than of an answer"*. The change is one deleted disjunct plus prose, and
  `return frame_workspace(fid)` offers no operator anything. Baseline in that run: 9761 tests
  OK.
- **So the mutation evidence is by hand**, five mutations, `python3 -B` with `__pycache__`
  cleared, each one red on the case that names it:

  | mutation | result |
  |---|---|
  | restore `for_session(fid) or …` (the defect) | 8 failures |
  | drop the pin rung — membership is the record alone | 6 failures |
  | pin `.strip()` → `.lstrip()` | 1 failure |
  | pin checked for truth instead of for a name | 1 failure |
  | drop `workspace_for`'s rung 0 | 4 failures |

### DIFF
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -6095,9 +6095,12 @@ def _pane_place(socket: str, pane: str | None) -> tuple[str, str] | None:
 
     **The question `chats.py` cannot ask and `select-window`'s return code does not
     answer.** `chats.of_workspace` decides membership from RECORDS — `state.own_workspace`,
-    which reads the pin the launcher wrote down, then the per-session pointer `charter
-    workspace use` writes, then the workspace the launch recorded — so two chats can share
-    a roster while their windows are in two different tmux sessions.
+    which reads the pin the launcher wrote down, then the workspace the launch recorded —
+    so two chats can share a roster while their windows are in two different tmux
+    sessions. (Both records are a launch's; the per-session pointer `charter workspace use`
+    writes was a third rung until #791, and this guard is what stood between it and a
+    `select-window` across sessions. It is not softened by the rung going away — a
+    migration or a hand-edited record reaches the same state.)
     The session is where a chat actually lives (`cmd_launch` makes the workspace the
     session), it is not written down anywhere charter can read, and it moves at runtime.
     So it is asked of tmux, at the moment it matters.
@@ -6190,7 +6193,9 @@ def cmd_chat(args) -> int:
        :func:`_pane_place`). Not one of the four, and it is here because every one of them
        assumes something no record on disk can promise: that the target's window is a
        window this client can be moved to. `chats.of_workspace` matches RECORDS
-       (`state.own_workspace`), whose pointer rung `charter workspace use` writes, so two
+       (`state.own_workspace`), and a record can name a workspace whose tmux session this
+       chat's window is not in — a migration, a hand-edited file, or (until #791) a
+       pointer `charter workspace use` wrote — so two
        chats can share a roster with their windows in two different tmux sessions — and
        `select-window` at another session's pane returns **0**, moves that session, and
        leaves this client exactly where it was. Steps 1 and 2 then reported a switch that
@@ -6290,12 +6295,15 @@ def cmd_chat(args) -> int:
     # anything** (#684). `chats.check` has established that both are chats of one
     # workspace and — since this issue — that both are on one tmux SERVER, and neither of
     # those is "in one tmux session". `cmd_launch` makes the workspace the session, but
-    # what membership is read from is RECORDS that move (`state.own_workspace`): `charter
-    # workspace use beta` typed inside chat `api.1` writes its pointer rung, and after
-    # that `of_workspace("beta")` returns `api.1` — a window of session `api` — beside
-    # `beta.1`, in one roster. (`switch.to_workspace` wrote that rung and the launch
-    # record both until §4j made it a refusal; this guard is what stood between that and a
-    # `select-window` across sessions, and it is not softened by the writer going away.)
+    # what membership is read from is RECORDS (`state.own_workspace`), and a record naming
+    # a workspace whose tmux session this chat's window is not in puts `api.1` — a window
+    # of session `api` — into `of_workspace("beta")` beside `beta.1`, in one roster.
+    # (Two writers reached that state and both are gone: `switch.to_workspace` wrote the
+    # pointer and the launch record until §4j made it a refusal, and `charter workspace
+    # use beta` typed inside `api.1` wrote the pointer until #791 stopped membership
+    # reading it. This guard is what stood between them and a `select-window` across
+    # sessions, and it is not softened by either writer going away — a migration or a
+    # hand-edited record reaches the same state with nothing having written anything.)
     here_place = _pane_place(socket, chats.pane_of(fid))
     there_place = _pane_place(socket, pane)
     if there_place is None:
--- a/charter/frame/chats.py
+++ b/charter/frame/chats.py
@@ -116,13 +116,21 @@ def of_workspace(workspace: str) -> list[str]:
     the fix for a chat missing from a roster is never to call it here.**
 
     What was wrong was the DEPTH, not the direction. This read `state.frame_workspace`
-    alone — the launch record, rung 2 — while `roster` keyed on all four, and
-    `charter workspace use <name>` typed at the agent writes rung 1 and never rung 2. So a
-    chat repaired that way was DRAWING `alpha` and excluded from `alpha`'s roster at the
-    same time, and the chats it had left could never see it again however it was repaired.
-    `state.own_workspace` is those chat-owned rungs — the recorded pin, the pointer, the
-    record — in the ladder's own order, so this and `roster` now ask one question at one
-    depth.
+    alone — the launch record — while `roster` keyed on every rung above it, so a chat
+    could be DRAWING `alpha` and excluded from `alpha`'s roster at the same time, and the
+    chats it had left could never see it again however it was repaired.
+    `state.own_workspace` is the chat-owned rungs — the recorded pin, then the record — in
+    the ladder's own order, so this and `roster` now ask one question at one depth.
+
+    **The per-session pointer is NOT one of those rungs, and #791 is why.** It was, for
+    one release. `charter workspace use <name>` typed at the agent writes it under the
+    CHAT's id (inside a frame `session.current()` is the chat — ADR 0019), so while it was
+    a rung that command re-homed the chat: measured, `charter workspace use gamma` inside
+    `alpha.1` took it out of `alpha`'s roster and put `gamma`'s chats on its bar, where
+    `check` approved rows `cmd_chat` then refuses. That is #733 and #788 through a door
+    neither issue named, and §4j settles it — a chat's workspace is its identity, and a
+    pointer written by whoever last typed a command is a property. Both rungs left are
+    written by the LAUNCH, so nothing after a launch moves a chat between workspaces.
 
     **Ordinal order, and it is not `created`'s.** Spec §3.5 asks for a `created` stamp
     per chat "so tab order and the ordinal allocator do not depend on directory mtime".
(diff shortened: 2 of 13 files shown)


diazoxide:
Filed #795 for the rename half named in "Two behaviour changes" above — the mechanical fix is
small, but whether a rename may move a chat's recorded workspace is a §4j question and belongs
beside #752 rather than inside this PR.

Issue #695, opened 2026-08-30
A workspace whose name contains a dot cannot open a second chat, and its frame's identity is set on another workspace's session


diazoxide:
Found while auditing #684's finding as a **class** — "what else in `commands_frame` targets
a pane or window without establishing it belongs to this frame". #684 is one instance
(fixed). This is the other, and it has a single root.

## The root

**tmux parses a `-t` target on `.` as `window.pane`.** So a target that is a *name*
containing a dot is not the name — it is a window and a pane index.

`instance.WORKSPACE_NAME_RE` is `^[A-Za-z0-9][A-Za-z0-9._-]*$`, so `api.2` is a workspace
an operator may create today. `state.workspace_prefix` keeps the dot (`_UNSAFE` allows
`.`), and `cmd_launch` makes the workspace the tmux SESSION name and passes that name as
`-t` in four places.

## Measured, tmux 3.7c, own socket

Two sessions: `api.2` (the workspace's) and `api` (a sibling workspace's).

**1. The second chat can never be opened.** `layout.chat_window_argv`, the production
builder, verbatim:

```
$ tmux -L … new-window -d -a -t api.2 -n api.2.2 -c /tmp -P -F '#{pane_id}' -- sleep 60
can't specify pane here
rc=1
```

`cmd_launch` returns 1 and the operator gets nothing. Every workspace named with a dot is
a one-chat workspace, and Phase 5's headline use — `charter claude` twice in one workspace
— is unreachable in it.

**2. The frame's identity is written on ANOTHER workspace's session.**
`_session_id_env_argv` / `_exit_path_env_argv` / `_charter_py_env_argv` all send
`set-environment -t <session>`:

```
$ tmux -L … set-environment -t api.2 FOO bar
rc=0
$ tmux -L … show-environment -t api  FOO
FOO=bar
```

rc 0, and it landed on session `api` — a different workspace's frame. `$CHARTER_SESSION_ID`
is exactly what the hotkey bind and every palette action resolve themselves from, so that
frame's palette would act for a frame it is not in. This is #684's shape again: a command
that succeeded on the wrong target.

**3. The frame cannot be attached to.** `cmd_launch`'s `attach -t <session>`:

```
$ tmux -L … has-session -t api.2
can't find pane: 2
rc=1
```

**4. A refusal is drawn on another workspace's screen.** `_say_on_screen` sends
`display-message -d 4000 -t <fid>`, and a chat id is `<session>.<n>`:

```
[...]

Issue #442, opened 2026-08-23
[workspace] default` and `workspaces/.default` are committed reading sites with no containment check


diazoxide:
`workspace.declared_default` (`charter/workspace.py:204-210`) and `instance.default_workspace_of` (`charter/instance.py:83-84`) return the committed value verbatim — neither calls `workspace.valid_name` nor `contain.child`. `workspace.resolve()` (`:269-272`) hands it to every `workspace_dir()` caller, and the read side joins it unguarded: `read_charter` (`:918-925`), `read_manifest` (`:676-681`).

Reproduced via both routes in a throwaway plane with fabricated content:

```
[workspace] default = "../../esc"
  resolve()     : ../../esc
  read_charter  : '---\nname: esc\n---\n\n## Vision\nCANARY-VISION-CONTENT\n'
  read_manifest : {'repos': [{'name': 'CANARY-REPO', 'branch': 'main'}]}
```

`charter workspace default ../../esc` was also **accepted** and written to `workspaces/.default`, after which `charter workspace current` and `charter workspace vision` both printed the out-of-plane content.

Why `contain` misses it: `file_refusal` → `_path_refusal` asks `within_data` only inside `if stat.S_ISLNK(...)` (`contain.py:449`), and a lexical `../` to a real regular file is not a symlink. The write and directory-listing sides *are* correctly refused (`workspace remember`, `workspace todo` both return `NOT_PLANE_DATA` naming the resolved outside path).

The persona twin does gate: `persona.declared_default` (`persona.py:653-659`) wraps the value in `reference_ok`. `contain.py:5-7` enumerates the five covered reading sites and this is not one; `tests/test_names_from_files_are_contained.py` claims to cover *"every entry point that takes a name from a file"* and has no row for it. The setter at `commands_workspace.py:268` gates only on `workspace_dir(name).exists()` — the exact mistake `persona.py:550-551` calls out in its own comment (*"'a path that exists' was never the question being asked (#337)"*).

**Severity: low, honestly.** No exfiltration channel — the bytes land in the victim's own terminal. It does **not** reach SessionStart (verified: the neighbour digest iterates `list_workspaces()`, which enumerates real directories). It cannot reach a credential — targets are files literally named `workspace.md`/`workspace.json`, and `contain.data_roots()` (`:207-210`) excludes the vault home. Every surface prints the name verbatim (`Active workspace: ../../esc`; statusline `⬢ ../../esc`). It is a containment-invariant break, not a confidentiality one.

[...]

PR #796, opened 2026-09-01
Quitting charter records the plane, so reopening puts it back — and says, per chat, what it cannot (§4e, §4f, §4i)

### LINKED ISSUE #785: A suite run from a worktree asserts against the parent clone's live plane, so tests can pass or fail for reasons the branch never touched

diazoxide:
**Rebased onto `9c0d0a9`.** Three commits landed under this branch, and one of them matters
to it: `2e6eecb` (#791/#794) took the per-session `.charter/sessions/<id>.workspace` pointer
out of `state.own_workspace`. That was this design's stated dependency and the residual it
could previously only *report* — see the last bullet under *Six things this found about
itself*. It is now a test.

Closes nothing on its own — this is §4e, §4f and §4i of
`docs/superpowers/specs/2026-08-30-charter-opens-like-an-ide.md`, plus the one edit of
stage 4 that had to come with them (`chat: close`).

> *"when user closing charter session we need to show warning, that all harness sessions
> will be closed, and when opening again we need to resume harness sessions as well. so for
> user it should be 'less invasive' — to not lose sessions, state etc."*

The operator restarted, got an empty frame, and recovered their conversation by typing
`/resume` by hand. **Three verbs were missing and none existed anywhere**: `charter --help`
had ten `frame-*` commands and neither a quit nor a close, and the palette registered
detach, two repo rows, three densities and three surfaces.

## What ships

| verb | what it does |
|---|---|
| `F2 → charter: quit` | records the plane, then stops every harness on it |
| `F2 → chat: close` | stops one chat and marks it so nothing brings it back |
| `charter reopen` | puts the recorded plane back — and resumes what it can |
| `F2 → chat: previous transcript` | opens what a chat had on screen before it stopped |

## The load-bearing trick, and why it was necessary

#757 keeps a chat's harness session id in `session.durable` so `clear_shape` cannot delete
it. **Measured: it does not survive a restart.** `reap` deletes a chat's whole directory
when its launcher pid is dead, and after a restart *every* launcher pid is dead — so the id
a resume needs is removed by the next launch.

Inverting `reap` is stage 4 of the delivery order, six edits wide, with ~60 tests asserting
the current rule. So the record is a **file**:

```
.charter/frame/
  alpha.1/                  <- a chat: still a liveness marker, still reaped exactly as now
  alpha.1.transcript        <- what was on its screen
  reopen.json               <- the plane
```

`reap` skips anything that is not a directory, and so does the tab bar's own scan. **`reap`
is not inverted and stage 4 is untouched.**

## Order is the feature

1. read the plane off disk (workspace, persona, harness, cwd, durable id per chat)
2. ask each tmux server which chats are live and where each window is
3. **capture each pane** while it still exists
4. **write the manifest**
5. drop transcripts no manifest names
6. **kill**, one `kill-window` per chat
7. **prune `inflight`**

4 before 6 because a record that depends on the thing it records succeeding is not a record
— a quit that cannot write the manifest **refuses and stops nothing**. 7 after 6 so the
tracker cannot end up quieter than the plane.

## Three things the spec got wrong, corrected in writing rather than by overriding

* **The gauge gate is dropped.** §4e's `exit_code(fid) is None` reads a `kill-window`ed chat
  as *live* (§2.17: that call writes no `exit`), which is the inverse of the case it was
  written for — and nothing needs suppressing anyway, because a reopened chat gets a fresh
  chat id and has no `session` mapping to inherit. The ordering hazard §4e names disappears
  with the gate.
* **Phase 5 Task 9 is amended in this PR**, so one of the two documents stops being true on
  paper: it said a reopened chat *"starts a fresh harness session"* and *"must not inherit
  the closed one's token gauge"*. Its Step 4 — *do not add a resume member to `Harness`* —
  **stands and is now load-bearing**: the flag rides the existing `extra` pass-through.
* **ADR 0018 is amended in this PR.** It said charter never touches the harness's pane; the
  measurement showed charter has been reading it since #384 (`_pane_last_words`, on both
  launch paths). The rule now says charter reads that pane at exactly **two** moments, both
  of them moments the pane is about to be destroyed, and writes nothing back — with the
  distinguishing question written down so the next exception has to argue for itself.

## Six things this found about itself, after the first commit

Recorded because each was found by looking rather than by a test failing, and each is a
commit of its own on this branch:

* **The confirming row had to move to the top.** Under the list it is about reads better and
  is wrong: `palette.narrow` puts the cursor on the first row when nothing has been typed,
  refused or not, so the confirmation opened with a *chat* row selected and Enter bound to
  nothing at all. Found by rendering the real surface.
* **The quit had nowhere to say its own sentence.** It runs detached with all three streams
  on `/dev/null` (the palette's pane is killed the instant a row is invoked) and by the time
  it finishes there is no frame left for a notice either. `cmd_launch` — the process that
  hands the operator their shell back — now says it, gated on the manifest naming its own
  chat.
* **…and that gate needed a second one.** `new_chat_id` walks upward from 1 and `reap` frees
  the ordinals a quit's chats held, so a reopen usually gets the *same* ids back — every one
  of its own launches was named in the manifest it was in the middle of acting on, and
  announced the quit it was undoing.
* **`charter reopen` refuses inside a tmux you already have.** That launcher stays awake for
  the life of each frame (it reads the exit status itself rather than installing the
  `pane-died` hooks), so it blocks exactly as `attach` does and a reopen would stop at the
  first chat. Suppressing that wait is its own change on a path whose exit-code contract
  depends on it.
* **A chat with no workspace record cannot be reopened**, and now says so instead of
  promising a resume: `reopen.read` holds a manifest entry's workspace to `valid_name`, so
  such a line would be discarded by every reader. It is stopped like the rest and honestly
  not recorded.
* **Two identical `list-windows` calls became one**, and two helpers with no caller were
  deleted.
* **A partial reopen now leaves behind exactly the retry.** Dropping the manifest whole once
  anything had launched threw away the record of the chats that did *not* come back; leaving
  it whole would have opened the ones that did a second time.
* **`chat: previous transcript` was targeting the wrong thing entirely**, and only a
  hand-run found it. `kill-window -t %N` resolves a pane to its own window — charter's
  early-death path relies on that — so `new-window -t %N` looked like it should too. It does
  not, on either version:

  ```
  new-window -t %0        rc 1   can't specify pane here
  new-window -t @0        rc 1   create window failed: index 0 in use
  new-window -t alpha.2   rc 1   can't specify pane here      <- #695, again
  new-window -t $0        rc 0
  ```

  `-t` there is a target-*window*; a window id is read as the index to insert at, and a
  dotted session name is parsed as `window.pane`. The session id is the one unambiguous
  spelling, and it is pinned on a real server now — a pager window really opens beside the
  chat, really shows the captured text with its colours rendered rather than printed, and the
  harness's own pane is left exactly as it was.
* **The stale-pointer residual is closed, and it was sharper than "a pointer might win".**
  A reopen mints a *fresh* chat id — but `new_chat_id` counts up from 1 and `reap` frees the
  ordinals a quit's chats held, so it very often gets the same *name* back. While the
  per-session pointer was a rung of `own_workspace`, a `charter workspace use gamma` typed in
  the OLD `alpha.1` decided the NEW `alpha.1`'s membership, over the manifest *and* over the
  launcher's own record. #791 removed that rung; pinned here with the stale pointer
  deliberately planted, and verified RED against a `state.py` with the rung put back.
* **And one this branch's own guard caught rather than a human**: a test class appended after
  a module's `if __name__` trailer, which `unittest.main()`'s `SystemExit` makes invisible to
  `python3 tests/<module>.py`. `test_no_test_hides_below_the_main_trailer` failed on CI 3.11
  and 3.12 and nothing local had seen it, because running a module *by name* is the spelling
  that runs them all. Fixed in `58d954e`; the guard is exactly the kind of thing it exists
  for and it worked.

## Verification

* **112 new cases in five modules**, 102 of them with no tmux at all — including one that
  drives the four hops from the keypress to the detached command through the real `_picker`
  and the real `_draw_palette`, with `own_the_tty` replaced by a script rather than a pty.
* **Two guards mutated and confirmed RED**: dropping `state.record_cwd` from the launch
  path, and pruning `inflight` before the kill instead of after.
* **A third property is over-determined and is documented as such** rather than claimed:
  three independent rules in `reap` keep a non-directory entry, so no single mutation can
  take the manifest's durability away.
* **A correction to the design's own §2.12, amended in this PR: the real-tmux tests DO run in
  CI.** `test.yml` installs no tmux, but `ubuntu-latest` ships one — so they ran there, and
  one of them failed on 3.11 and 3.13 while passing on 3.14 in the same run. A race (a pane's
  process has not exec'd when `new-window` returns), found by CI and not by this machine. The
  honest statement is narrower than "no tmux in CI": **CI answers on whatever tmux the runner
  image happens to ship, which is neither of the two versions charter promises** — a third
  machine's opinion on top of the hand-runs rather than a substitute for them.
* **Real tmux on 3.7c and at the 3.2 floor** — the capture is real bytes off a real pane
  with its trailing spaces intact (`-N`, which `-e` alone trims); the kill really ends the
  chats and, with the last window, the session and then the server; and **two planes on one
  server, both with a workspace called `default`, both with a window in the one tmux session
  that name can have — quitting one leaves the other's harness running** (§3.3, the case a
  single-plane test is blind to).
* **Full local suite green.** Before the rebase it ran three times at 9844 / 9852 / 9861
  tests with one failure each — `test_frame_border_surface`, which reproduced at `9cd519e` in
  a clean worktree but **not** in the sweep sandbox's fresh `git clone` (9852 green). That
  observation is now closed upstream: #799 fixed #785, so a suite run from a linked worktree
  no longer reads the operator's live plane, and this branch runs clean.
* Per §2.12, **none of the tmux half runs in CI**; every tmux claim above is a hand-run on
  this machine on both versions.

## The deletion sweep: 225 mutations, every one of them with a verdict

**Including the 39 the CI shards ran out of time before reaching** — those were run locally
by explicit line number, same tool, same sandboxes, same full-suite confirmation for a
survivor, just a smaller plan. The arithmetic, so it can be checked:

```
100  survivor mutants   (77 from the CI shards, 23 from the local 39-line run)
-23  pinned since, each re-run against the branch head rather than assumed
────
 77  residual from those two runs
 +2  first seen in this head's own sweep, which measured a different subset
────
 79  merging unpinned, named and filed as #810
```

**This head's sweep is a partial, and its check name says so** — `no verdict: 54 survivors
so far, 171 of 225 measured, 54 out of time` (#803/#804 doing exactly its job: six of eight
shards hit the hour and reported what they had rather than a verdict they had not earned).
Two of its 54 had never appeared in the earlier run, which is worth stating plainly: **each
run samples a different subset, so a survivor list is a union across runs and never a
snapshot.** Both new ones are in #810.

Three clusters were closed here, and each is recorded by its **cause** rather than by its
line list, because each was one missing habit and not N missing tests:

* **The fallback cluster (5) — not five gaps, one fixture that never lies.** `_plant` always
  wrote a server, a workspace, a harness pane and an identity, so every fallback in
  `leave.record` (`leave.py:152/156/157/159/160`) read a truthy value in every test. Closed
  by a fixture that records nothing at all.
* **The on-disk names (8), and this is the rule worth carrying off this branch: a round trip
  cannot pin a name.** Every test wrote with `reopen.write` and read with `reopen.read`, so
  renaming the manifest, the `.transcript` suffix, the `cwd` file, the `closed` marker or any
  JSON key moved both sides together and stayed green — while on a real plane it silently
  orphans everything already on disk. The fix is to spell the literal a second time, by hand,
  in the test: **the duplication *is* the assertion.**
* **The manifest parser (10).** Every case fed a bad **value** inside the right **container**
  — a bad version, a chat id with a semicolon in it, a workspace that is `../etc` — so no
  type guard was ever handed the wrong container.

**And one real defect, found by a mutant that could not be reddened at all.** A `>` in the
transcript byte cap resisted every test, because at exactly the cap `text[-N:]` returns the
whole string either way. An unpinnable comparison is a pointer, not a line to delete, and
what this one pointed at was bytes-versus-characters. Fixed, with the multi-byte case pinned.

### Why the parser is the cluster that had to close before merge

It is the only one on this branch whose blast radius leaves the frame. Those values drive
`os.chdir` and a tmux argv (`_reopen_one`, `_attach_after_reopen`), and the manifest is a
plain file that outlives the process that wrote it and can be **older than the charter
reading it** — §5 of the IDE spec names exactly that as a cost the durability design creates,
and a hand-edited or half-written one is the ordinary case, not the exotic one. A guard
missing there is not a wrong answer that degrades; it is an `AttributeError` or a `KeyError`
out of a function whose entire contract is that it degrades, raised in the process that hands
the operator their shell back — `cmd_launch` asks `read()` whether this chat was named by a
quit. This is `state.frame_workspace`'s position (#442) and `chats.of_workspace`'s (#475), at
the one boundary this module owns.

Ten mutants, each confirmed RED against the shipped file by hand — one mutant at a time,
before its test was trusted:

| line | mutant | the case that kills it |
|---|---|---|
| `reopen.py:193` | `chat.chat or ""` → `chat.chat` | a record whose chat id is not even text |
| `reopen.py:260` ×2 | the `frames` type gate, deleted / `False` | a manifest with no `frames` key, and one where it is a number |
| `reopen.py:264` | the frame and `chats` type gate, deleted | a frame that is not a dict, and a `chats` that is not a list |
| `reopen.py:269` | `f.get("workspace") or ""` → `f["workspace"]` | a frame that never recorded a workspace, and one whose workspace is `null` |
| `reopen.py:272` ×2 | `at if isinstance(…)` → `0`, and → `at` | the recorded `at` read back whole, and an `at` that is a string |
| `reopen.py:273` | `focus if isinstance(…)` → `focus` | a `focus` that is a number |
| `reopen.py:287` ×2 | the chat-entry type gate, deleted / `False` | a chat entry that is not a dict |

**CI reached five of these ten on this head and calls all five pinned** — `264`, `269`,
`272`, `273` and `287` come back `pinned` in the shard logs, where four of them were
survivors before. The other five were among the 54 the shards ran out of time on, which is
the same set that forced the local by-line run in the first place; those rest on the hand-run
above, which is the stronger evidence anyway because it applies one mutant at a time to the
shipped file and names the case that kills it.

`260` nearly stayed open, and the reason is worth keeping: **two guards in sequence mask each
other.** It and the per-frame gate below it answer identically for every `frames` that is
*iterable* — a string or a dict walks into the loop and is refused one line further down with
the same `None`. What tells them apart is a `frames` that cannot be iterated at all: an absent
key, and a number.

**The refusal is asymmetric, and the tests now pin both halves of it.** A bad *shape* — a
top level that is not a dict, a `frames` that is not a list, a frame that is not a dict, a
`chats` that is not a list — takes the WHOLE manifest down, because *a manifest charter
cannot read whole is not a plane charter may half-restore*. A bad *value* — a chat entry
that is not a dict, a chat id or workspace that could not be one — drops that record and
leaves the rest, because `cmd_reopen` reports the difference between what was recorded and
what came back, so a dropped chat is a sentence rather than a silence. Both rules were
already written down in `read()`'s docstring; neither had a test on the wrong-container side
until now.

Worth stating for a reader who will notice it: the **frame's** own `workspace` is bounded by
`str(… or "")` and not by `valid_name`, unlike the chat's. That is deliberate and was
re-checked here — `f.workspace` is read in exactly two places, both of which copy it straight
back into a rewritten manifest. Every operative use is the CHAT's workspace (`_reopen_one`'s
launch argument, `_attach_after_reopen`'s `workspace_prefix`), and that one *is* held to
`valid_name` by `_usable`. The frame label is a grouping key that round-trips and reaches
nothing.

One case calls `_usable` directly and says why in full: no shipped caller can hand it a
non-string, because `_chat` normalises all seven text fields on the way in. The fallback is
there so the predicate is **total** over `Chat` — it is the last gate before a recorded name
becomes a transcript path and a tmux target, and a gate that raises on the one input it was
written to refuse is not a gate.

### What merges unpinned, and why these are the acceptable ones

**79 survivors, named rather than left to be rediscovered — #810**, in four groups, each with
its cause:

| group | cause, in one line |
|---|---|
| `except OSError` (13) | no test in the suite makes the filesystem refuse, so the clause is never entered — and the sweep's own `PLATFORM` note says some of these are *unreachable* on darwin rather than unpinned, which CI has to answer before a test is worth writing |
| refusal branches (45) | every case drives the working path and asserts what came back, so a refusal is only ever the branch NOT taken — the fixture is always well-formed enough to get past the gate |
| operator-visible sentences (16) | the round-trip rule again, with a sentence in place of a file name: every assertion goes through the module constant on **both** sides (`assertEqual(rows[0].title, leave.NOTHING_OPEN)`), so rewording moves the test with it |
| `contain` wrappers (5) | every `harness`, `cwd` and chat id in the suite is already one line of printable ASCII, so `one_line` and `readable` are the identity function on every input the tests supply |

**The line here is not a count, it is a question: does this decide where a process runs, or
what gets executed?** For all 79, no. A missing `except OSError` costs a traceback where a
warning belonged. An untested refusal costs one chat not reopened, on a path whose working
half is tested and whose failure the operator is told about. An unpinned sentence costs
wording. An unpinned `contain` costs a report line, and only for a value hostile enough to
carry an escape. **The two that would have answered yes are the two that are pinned** — the
parser, whose values reach `chdir` and an argv, and the on-disk names, whose drift orphans
every plane already recorded.

Three of the 79 are flagged in #810 as the ones to take first, and they are not in the
biggest group: `leave.py:397–399` are palette **action ids**, and an action id is a binding
surface a `charter.toml` can name. Renaming one silently breaks an operator's keybinding,
which puts them with the manifest's names rather than with a warning's wording.

## Not done here, stated rather than implied

* **The chat directory is not durable** (stage 4, six edits) — which is why a reopened chat
  is a new tab with a new id rather than the same directory woken up.
* **Bare `charter` does not reopen.** `charter reopen` is a thing you ask for.
* **Resume is Claude Code only**, and the warning says so per chat.
* **`chat: close` cannot refuse a busy chat** — `inflight` records name no chat, so there is
  no reading to refuse on. Said out loud rather than faked.
* **A chat with no workspace record cannot be reopened.** Stopped like the rest, honestly not
  recorded, and its warning row says so; choosing a workspace for it would be §4j's
  re-homing arriving as a convenience.

### DIFF
--- a/charter/cli.py
+++ b/charter/cli.py
@@ -359,6 +359,18 @@ def build_parser() -> argparse.ArgumentParser:
     tr.add_argument("-n", type=int, default=0, help="Show only the last N raw events.")
     tr.set_defaults(func=commands_persona.cmd_trace)
 
+    # §4e/§4i's third verb, and the only one of the three an operator types rather than
+    # presses: quit and close are palette rows in a running frame, and by the time a reopen
+    # is wanted there is no frame to press a key in. Registered HERE, above
+    # `_add_frame_parsers`, so it is one of the commands the harness collision guard checks
+    # a `cli_name` against rather than one added after that snapshot is taken.
+    ro = sub.add_parser(
+        "reopen",
+        help="Put back the plane your last `F2 → charter: quit` recorded: every "
+             "workspace, chat, persona and directory, and — for Claude Code — the "
+             "conversation.")
+    ro.set_defaults(func=commands_frame.cmd_reopen)
+
     _add_harness_parser(sub)
     _add_workspace_parser(sub)
     _add_worktree_parser(sub)
@@ -756,7 +768,9 @@ def _add_frame_parsers(sub) -> None:
     _core_commands = set(sub.choices) | {"frame", "panel", "frame-palette",
                                          "frame-probe", "frame-respawn", "frame-density",
                                          "frame-resize", "frame-gather", "frame-switch",
-                                         "frame-toggle", "frame-chrome", "frame-chat"}
+                                         "frame-toggle", "frame-chrome", "frame-chat",
+                                         "frame-quit", "frame-close",
+                                         "frame-transcript"}
 
     # Which harness (by `.name`, never `.cli_name` — that's the dict key below) has
     # already claimed each word, so a SECOND harness wanting it is told who got there
@@ -990,6 +1004,43 @@ def _add_frame_parsers(sub) -> None:
     ch.add_argument("chat_id")
     ch.add_argument("--chat", dest="chat", default="")
     ch.set_defaults(func=commands_frame.cmd_chat)
+
+    # §4i's quit. Started DETACHED by the palette's confirmation row
+    # (`commands_frame._start_leaving`) and typeable by hand, which is why it warns on its
+    # own stderr as well as on the surface that confirmed it: §4i is explicit that quit
+    # *warns and proceeds*, so the record of what was lost has to survive the keypress.
+    #
+    # **It takes no target, and that is the design rather than an omission.** Quit is
+    # plane-scoped — the operator's own words were "all harness sessions will be closed" —
+    # and §3.3 bounds it from the other side: one tmux server serves every plane on the
+    # machine and session names carry no plane, so the set of things it stops can only come
+    # from this plane's own chat directories. `--chat` is where the keypress came FROM, for
+    # `frame-toggle`'s reason, and is used only to decide which workspace a later
+    # `charter reopen` puts the operator back on.
+    qt = sub.add_parser("frame-quit")
+    qt.add_argument("--chat", dest="chat", default="")
+    qt.set_defaults(func=commands_frame.cmd_quit)
+
+    # §4d's `chat: close`, which is stage 4's new reaper arriving on its own. Quit's
+    # teardown with one target and one file more (`state.record_closed`), because a missing
+    # `exit` file means "was open" and closing writes no exit either — see `cmd_close`.
+    #
+    # **Two chat ids, and they are two different questions**, exactly as `frame-chat` has:
+    # the POSITIONAL is which chat to close, optional so the bare command closes the one
+    # you are in; `--chat` is where the keypress came from. Deliberately NOT `choices=`,
+    # for `frame-chat`'s reason — which chats exist is a property of the frame root at the
+    # moment the key fires, so there is no set for argparse to hold.
+    cl = sub.add_parser("frame-close")
+    cl.add_argument("chat_id", nargs="?", default="")
+    cl.add_argument("--chat", dest="chat", default="")
+    cl.set_defaults(func=commands_frame.cmd_close)
+
+    # §4f's other half: the captured scrollback, offered rather than replayed. Opens a
+    # pager in a window of its own; nothing is ever written into the harness's pane. See
+    # the ADR 0018 amendment this change carries.
+    tr = sub.add_parser("frame-transcript")
+    tr.add_argument("--chat", dest="chat", default="")
+    tr.set_defaults(func=commands_frame.cmd_transcript)
 
     # A TOP-LEVEL sibling of `frame` for a DIFFERENT reason than `frame-palette` above:
     # that one exists because `_split_frame_argv` eats everything
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -7659,7 +7659,7 @@ def _plane_servers() -> list[str]:
     live there" by the caller.
     """
     found = [SOCKET]
-    for fid in leave._plane_chats():
+    for fid in leave.plane_chats():
         server = state.frame_server(fid) or SOCKET
         if server not in found:
             found.append(server)
@@ -7774,6 +7774,17 @@ def _record_the_plane(doomed, *, focus: str, active, windows) -> int | None:
     entries: dict[str, list] = {}
     order: list[str] = []
     for c in doomed:
+        if not c.workspace:
+            # **Stopped, and honestly not recorded.** A reopen rebuilds one tmux session per
+            # workspace and hands the launcher a `--workspace`; there is nothing to rebuild a
+            # chat into that says nothing about its workspace, and choosing one for it would
+            # be §4j's re-homing arriving as a convenience. `reopen.read` refuses such a
+            # record anyway (`_usable` holds the workspace to `valid_name`), so writing one
+            # would put a line in the manifest that every reader discards — a record that
+            # looks like a promise and is not. The warning already told the operator
+            # (`leave.NOT_REOPENED`), and `cmd_quit`'s own line says how many of how many
+            # were kept.
+            continue
         server = state.frame_server(c.chat) or SOCKET
         transcript = ""
         dest = reopen_state.transcript_path(c.chat)
@@ -7793,7 +7804,10 @@ def _record_the_plane(doomed, *, focus: str, active, windows) -> int | None:
         frames.append(reopen_state.Frame(workspace=ws, chats=tuple(entries[ws])))
     if not reopen_state.write(frames, focus=focus):
         return None
-    reopen_state.prune_transcripts({c.chat for c in doomed})
+    # Keyed on what was RECORDED and not on what was stopped: a transcript is a file in the
+    # frame root, and the only collector it has is this call. Keeping one for a chat no
+    # manifest names would leave it there for good.
+    reopen_state.prune_transcripts({c.chat for f in frames for c in f.chats})
     return sum(len(f.chats) for f in frames)
 
 
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -8020,10 +8020,13 @@ def cmd_reopen(args) -> int:
     whichever session tmux happened to make current would put the operator somewhere they
     did not leave.
 
-    **The manifest is consumed.** A record describes one quit, and a second `charter reopen`
-    against the same file would open every chat a second time with nothing on screen to tell
-    the duplicates apart. It is dropped after the launches and before the attach, because
-    the attach does not return until the operator leaves.
+    **The manifest is consumed, chat by chat.** A record describes one quit, and a second
+    `charter reopen` against the same file would open every chat a second time with nothing
+    on screen to tell the duplicates apart — but deleting it whole would throw away the
+    record of the chats that did NOT come back, which is exactly the state an operator would
+    want to retry. So what is left behind is precisely the retry (:func:`_consume`), and it
+    is written after the launches and before the attach, because the attach does not return
+    until the operator leaves.
     """
     m = reopen_state.read()
     if m is None or not m.frames:
@@ -8070,8 +8073,34 @@ def cmd_reopen(args) -> int:
                  "record is left in place so this can be tried again.")
         return 1
     util.ok(f"charter: reopened {len(back)} of {len(m.all_chats())} chats")
-    reopen_state.forget()
+    _consume(m, {r.chat.chat for r in back})
     return _attach_after_reopen(m, back)
+
+
+def _consume(m, done) -> None:
+    """Take the chats in *done* out of the manifest, and drop it when nothing is left.
+
+    **Consumed rather than deleted, because a partial reopen has two halves and only one of
+    them is over.** A record describes one quit: leaving it whole would open every chat a
+    second time on the next `charter reopen`, with nothing on screen to tell the duplicates
+    apart — but deleting it whole would throw away the record of the chats that did *not*
+    come back, which is exactly the state an operator would want to retry. So what is left
+    behind is precisely the retry: the chats still owed, in their own frames, with the same
+    focus.
+
+    Called after the launches and before the attach, because the attach does not return
+    until the operator leaves.
+    """
+    left = [reopen_state.Frame(workspace=f.workspace,
+                               chats=tuple(c for c in f.chats if c.chat not in done))
+            for f in m.frames]
+    left = [f for f in left if f.chats]
+    if not left:
+        reopen_state.forget()
+        return
+    if reopen_state.write(left, focus=m.focus, at=m.at):
+        util.warn(f"charter reopen: {sum(len(f.chats) for f in left)} chat(s) still "
+                  "recorded — `charter reopen` again to retry just those")
 
 
 def _reopen_one(c) -> "Reopening | None":
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -147,10 +147,14 @@ import subprocess
 import sys
 import time
 
-from . import config, contain, harness, instance, tui, util, workspace
+from . import config, contain, harness, inflight, instance, tui, util, workspace
 from .frame import (actions as frame_actions, builtin_actions, chats, choose, component,
-                    gather, layout, overlay, pane, palette, picker, state, switch,
+                    gather, layout, leave, overlay, pane, palette, picker, state, switch,
                     tmuxctl)
+# Aliased because `cmd_reopen` is a function in this module and `reopen` reads as one: the
+# module answers what a quit RECORDED and the command is what puts it back, and a bare
+# `reopen.read()` beside `cmd_reopen` invites a reader to think one is the other.
+from .frame import reopen as reopen_state
 # Aliased: `cmd_launch` already has a local variable named `slots` (the VISIBLE slot
 # list `layout.visible_slots` returns) — importing the renderer registry under its own
 # name would be shadowed by that local the moment it's assigned, and a `slots.SLOTS`
@@ -3152,6 +3156,14 @@ def _launch_in_operator_tmux(socket: str, session: str, *, ws: str,
     # one's.
     state.record_identity(fid, _frame_identity_env(env))
     cwd = os.getcwd()
+    # And WHERE, which until now was read on this line, handed to tmux and dropped
+    # (`state.record_cwd`). It is the fourth of §4e's four restore items and the only one
+    # with nowhere to live: `workspace` is a file, `identity` carries the harness and the
+    # persona pin, and a cwd could not join `identity` because every value in that record
+    # goes onto a world-readable tmux `-e` argv. Recorded on BOTH launch paths, from the
+    # same `os.getcwd()` each one already reads, because a chat reopened out of the wrong
+    # directory is a harness given a different plane to look at.
+    state.record_cwd(fid, cwd)
     opened = tmuxctl.run(
         "opening a window for the frame",
         layout.window_argv(socket=socket, session=session, window=fid, cwd=cwd))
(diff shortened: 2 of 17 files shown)


diazoxide:
## Where verification stands, at `28d17cf`

**Green, on this machine:**

* full suite three times over this branch's heads — 9844 / 9852 / 9861 tests, each with the
  same single failure: `test_frame_border_surface::
  test_this_planes_own_committed_frame_answers_the_report`, which fails identically at
  `9cd519e` in a clean worktree (**#785**). Worth adding to that issue: it is a *worktree*
  artefact, not a machine one — the deletion sweep's own sandbox, which is a fresh `git clone`
  of this branch, ran **9852 tests green**. The test's disagreement is with the committed
  `charter.toml` a worktree checks out.
* 735 targeted tests on the exact head — the five new modules plus `frame_launcher`,
  `frame_palette`, `frame_state`, `builtin_components`, `action_registry` and both `inflight`
  modules.
* all 13 real-tmux cases on **3.7c and at the 3.2 floor**, the transcript ones run four times
  over to shake the races out.

**CI, on this branch:** the four-Python matrix was green on `377d854`, `a25fd66` and
`0563fa2`. Two heads went red and both were mine, both test-side, both fixed and re-verified:

1. a class appended below a module's `if __name__` trailer (`test_no_test_hides_below_the_main_trailer`
   on 3.11/3.12) — invisible to `python3 tests/<module>.py`, and invisible to me because I had
   been running the module *by name*, which is the spelling that runs them all;
2. two races in one assertion in the new pager test — `new-window` returns before the pane's
   process has exec'd, and a pager that *has* exec'd has not necessarily painted. Failed on
   3.11/3.13 then on 3.14 in two successive runs; both polled now.

Neither touched production code.

**Still owed and stated rather than assumed:** at the time of writing, the matrix and the
sweep for this head (`28d17cf`, sweep run `33497714817`) are **queued** — the runner pool is
backed up by several concurrent branches. The last sweep that got as far as planning sized
this diff at **230 mutations across 7 files / 8 shards**; its shards were cancelled by my own
successive pushes, and the local equivalent had a green baseline (9852 tests) and reported
`pinned` for every mutation it reached before I stopped it for machine contention. **So there
is no sweep verdict yet, and that is a gap rather than a pass.** Please read
`gh run view 33497714817` before merging.

diazoxide:
## Rebased onto `9224a68` — head is now `36996e6`

Clean rebase, no conflicts; merge-base is `9224a68` exactly, so the sweep charges only this
branch's lines.

**`2e6eecb` (#791/#794) landed, and it closes the one residual this design could previously
only report.** I had written it up as "the stale `.charter/sessions/<id>.workspace` pointer
can still bite via ordinal recycling". Looked at against the landed change, that was
understated: a reopen mints a *fresh* chat id, but `new_chat_id` counts up from 1 and `reap`
frees the ordinals a quit's chats held — so a reopened chat very often gets the **same name**
back, and while the pointer was a rung of `state.own_workspace` a `charter workspace use
gamma` typed in the old `alpha.1` would have decided the new `alpha.1`'s membership over the
manifest *and* over the launcher's own `record_workspace`.

That rung is gone. It is pinned now with the stale pointer deliberately planted, and verified
**RED** against a `state.py` with the rung put back (`gamma` instead of `alpha`). Two
docstrings that counted `own_workspace`'s rungs are corrected, and the Task 9 amendment's
"except where the ordinal is recycled" caveat is replaced by what actually settles it.

## Verification at `36996e6`

* **Full local suite: `Ran 9895 tests … OK (skipped=8)`** — the first fully green local run
  this branch has had. #799 is why: the `test_frame_border_surface` failure that dogged every
  earlier run was #785, and it is fixed.
* **CI, four Pythons: `test (3.11)` `test (3.12)` `test (3.13)` `test (3.14)` — all
  `success`.** `CodeQL`, `Analyze (python)`, `Analyze (actions)`, `Size the sweep` — all
  `success`.
* **Real tmux on 3.7c and at the 3.2 floor** — 13/13, re-run after the rebase.
* **Deletion sweep: 8 shards in progress**, run `33500900581`. No verdict yet; that is the
  one thing still owed.

Nothing has been pushed since these checks started.

diazoxide:
## The CI sweep cannot answer a diff this size, and it said so rather than pretending

Run [`33500900581`](https://github.com/diazoxide/charter/actions/runs/33500900581) at `36996e6`
finished as **`no verdict: 8 of 8 shards did not report`**. Nothing was pushed to cancel it —
`origin/quit-records-the-plane` is still `36996e6` and `origin/main` is still `9224a68`. The
timing settles it:

```
Size the sweep        11:09:49 → 11:15:10   success
Sweep shard 1..8      11:15:12 → 12:15:2x   cancelled   ← 60 minutes, to the second, all eight
Add up what the shards found                success
no verdict: 8 of 8 shards did not report    success
```

Every shard hit `timeout-minutes: 60` simultaneously. **`plan` predicted it in the same run:**

> `231 mutations → 8 shard(s) of at most 29 each`
> ⚠ `231 mutations is past the 224 that 8 shards can measure inside 40 minutes each. Every one
> of them is still swept — nothing here is dropped — but a shard may be cancelled at the job
> timeout, and a cancelled shard reports no verdict rather than a short one.`

So this is the gate working exactly as designed — *completed-and-clean*, *completed-with-N-
survivors* and *did not complete* are three answers, and it gave the third — and it is a
**structural limit rather than a signal about this branch**: at 8 shards the sweep tops out
around 224 mutations, and this diff is 231. Any branch of this size gets no verdict.

**A green tick on `no verdict: … did not report` must not be read as a clean sweep.** It
concluded `success` only because `--gate` is non-enforcing; the sweep found nothing because it
never finished measuring.

### So it is being run here instead

A local `tools/sweep.py --gate` on `36996e6`, charged against `9224a68` — 7 files, 2,227 added
lines, the same 231 mutations — with no 60-minute cap. That is the only route to an actual
verdict for a diff this size, and it will be reported by name when it lands.

*Worth someone's attention beyond this PR:* the shard cap is a `plan`-time constant and the
warning is already written; what is missing is a shard count that scales past 224, or a
`timeout-minutes` that matches what a shard is actually asked to do. Filing that is not this
branch's to do, but it will bite the next large one identically.

Issue #766, opened 2026-08-30
Hover is deliverable to a panel — measured on 3.7c and the 3.2 floor — and the affordable shape is an in-pane highlight, not a footer tooltip


diazoxide:
Filed to keep a measurement rather than to ask for a feature: the question *"can charter receive hover at all"* now has an answer, and the answer is yes with one condition and one shape constraint. Without this written down the next attempt re-derives it, and the obvious design (the operator's own suggestion of a tooltip on the footer) is the expensive one.

## What was measured

Real tmux **3.7c** and the **3.2 floor** (`~/.local/share/charter-testing/tmux-3.2`), own socket, real client on a real pty, two panes split `-h -l 40`, harness pane active, SGR reports injected the way a reporting terminal sends them. Every mode request tmux wrote to the outer terminal was recorded off the client's master fd; the non-active pane ran a probe that asked for a named mode and logged every byte it read off fd 1.

```
harness pane asks 1000+1006 (what an agent harness asks). panel pane asks 1003+1006.

tmux mouse OFF
  outer terminal told:  1000h 1006h        <- the ACTIVE pane's request, only
  the panel's 1003 request is NOT propagated. no motion is ever sent.

tmux mouse ON
  outer terminal told:  ... 1003h          <- BECAUSE the non-active panel asked
  motion over the PANEL    ->  panel READ  \x1b[<35;11;5M   (pane-relative, 1-based)
  motion over the HARNESS  ->  nobody
```

Byte-identical on both versions (3.2 asks the outer terminal for a slightly different mode set with `mouse on` — `1003h` alone rather than `1000h 1002h 1003h` — but delivery is the same).

## Three findings

1. **Hover IS deliverable to a non-active `charter panel` process**, conditional on `[frame] mouse = true`. A pane asking `\x1b[?1003h` makes tmux propagate 1003 to the outer terminal *even though that pane is not active*, and the motion reports then arrive in that pane's own cells — the same coordinate contract `click` already has.
2. **It cannot poison the harness.** tmux filters per-pane by what each pane asked for. With 1003 on for the whole terminal, motion over the harness pane (which asks 1000/1002) reaches **nobody**. This was the measurement that mattered most: the harness is the thing the frame exists to protect.
3. **With `mouse = false` hover is dead**, and not merely quiet: only the ACTIVE pane's request is propagated, so no motion is ever sent by the terminal at all. This is the same rule `docs/frame.md` already states for `click`, one mode over.

[...]

PR #801, opened 2026-09-01
A short terminal keeps the row that says where you are, and the sidebar's hidden todos have a keyboard route (#740, #742)

### LINKED ISSUE #740: A short terminal drops the identity strip and keeps a one-row `repos` pane that can only draw its own heading
On a short terminal `layout.visible_slots` drops `right` and then `top`, and keeps `repos`
— which at that point has room for its heading and nothing else. The result spends three
rows (two rules and a count) on a terminal that has just been judged too short for the
one-row strip that names the workspace and the persona.

## Measured

120 columns, tmux 3.7c, alpha (8 clones). Shrinking the terminal:

```
h=20  panes: top 1  harness 12  sidebar 12  repos 3  attention 1
h=19  panes:        harness 12              repos 4  attention 1     <- `top` dropped
h=15  panes:        harness 11              repos 1  attention 1
h=10  panes:        harness  6              repos 1  attention 1
```

At h=10 the whole screen is:

```
h$
                        <- five rows of harness
────────────────────────────────────────
▪ repos 8
────────────────────────────────────────
7 todos · F2 palette
```

Six rows of agent session, three rows to say the number 8, one row of attention strip.
The workspace name and the active persona — the two things `top` exists to tell you — are
gone.

## Why this is the wrong trade

`docs/frame.md` ranks `bottom` above everything ("the whole reason a cramped terminal is
worth framing at all") and that is clearly right. It does not rank `top` against `repos`,
and the implicit answer the floors give — drop the identity strip, keep a heading-only
table — inverts what each one is worth at that size. `top` is one row and carries two
facts; `repos` at one row carries a count and costs three rows to do it.

Concretely: a `repos` pane whose content does not fit **one repo row** should drop like
`right` and `top` do, rather than existing as a heading. That is the same rule the table
already applies horizontally — below 95 columns it is not drawn at all rather than drawn
with columns cut off — applied to its other axis.

### LINKED ISSUE #742: The sidebar handles no events: its `…(+N more)` todos cannot be reached by mouse, keyboard or palette

diazoxide:
Closes #740. Closes #742.

Two frame-UX reports from the same audit. The third, #766, is closed separately as *measured, deliberately not shipped* — its numbers and mine are on the issue so the next person does not re-measure it.

## #740 — the ladder was inverted at the bottom

`_ROW_DROPS` took `top` the instant rows fell below `min-rows`, while `repos` had **no rows threshold at all**. So a 120x10 frame spent three rows — two pane rules and `▪ repos 8` — on a repo count, having just decided it could not afford the one row that names the workspace and the persona.

**The rule I chose, stated because the old one was unexamined rather than wrong:** *a rung drops when the pane it would get cannot carry the thing the rung is for; among rungs that still can, the ones whose facts another surface reaches go first.*

That gives the ladder `_DROP_ORDER` always claimed, now honoured on both axes:

| rung | goes when |
|---|---|
| `right` | any shortage — it costs columns *and* a divider |
| `repos` | its pane is too narrow for a table (`repos_fits`) **or too short for one repo row** (`repos_fits_rows`, new) |
| `chats`, `workspaces` | at `min-rows`, whole — `F2` reaches every chat and workspace in two keystrokes |
| `top` | only at the floor, with everything else |
| `bottom` | never |

Two changes:

* **`repos` gets a rows threshold, the mirror of its columns one.** `repos_fits_rows` over a new `repos_row_cap`, tested against `_table_min_rows()` which READS `slots._TABLE_MIN_ROWS` exactly the way `_table_min_cols()` reads `statusline._LEFT_W`. `repos_rows` applies the same `repos_row_cap`, so "is this pane worth splitting" and "how tall is it" come from one arithmetic instead of two.
* **`top` leaves `_ROW_DROPS`.** Nothing else on a short terminal says which workspace you are in or which persona you are being, and with the sidebar gone it is also the plane's only roster (#530).

### Measured, on real tmux

3.7c and the 3.2 floor (`~/.local/share/charter-testing/tmux-3.2`), own socket, 120 columns, a plane with eight clones. Byte-identical on both versions. The splits are built from `layout`'s own edge and size answers and the panes read back with `list-panes`.

| rows | before | now |
|---|---|---|
| 20 | identity 1 · harness 12 · sidebar 12 · table 3 · attention 1 | *unchanged* |
| 19 | harness 12 · table 4 · attention 1 | identity 1 · harness 12 · table 2 · attention 1 |
| 18 | harness 12 · table 3 · attention 1 | identity 1 · **harness 14** · attention 1 |
| 17 | harness 12 · table 2 · attention 1 | identity 1 · **harness 13** · attention 1 |
| 16 | harness 12 · table 1 · attention 1 | identity 1 · harness 12 · attention 1 |
| 15 | harness 11 · table 1 · attention 1 | identity 1 · harness 11 · attention 1 |
| 10 | harness 6 · table 1 · attention 1 | identity 1 · harness 6 · attention 1 |

The "before" column reproduces the issue's own measurements exactly. **The harness never loses a row** — a table pane starved to its floor paid a border plus one row, which is what `top` costs — and it gains one at 17 and two at 18. **What this costs is one or two repo rows between 17 and 19**, which is the trade `_DROP_ORDER` already ranked and the rows axis had no test to make.

`tests/test_frame_layout.py` pins that whole table as literals through `harness_rows`.

## #742 — what remained, and how it closes

Confirmed before touching anything: **#765 shipped the persona half** — the sidebar declares `click`, `slots.persona_section` publishes a row→persona map on every rung, a click on a name runs `charter frame-switch --persona`, and a click on the badges puts the glyph legend on the attention row through #729's dwell. Only the todo overflow remained, which is why it was not auto-closed.

`F2` → **`todo: read the next open todo`** is the route. One repeatable palette row that reads the open todos out on the palette's own header:

```
charter /todo · todo 3/7: cut 0.55.0
```

It repeats like the two `repo: select the …` rows, so five hidden todos are five Enters and one palette. It wraps rather than pressing nothing; it says where in the list it is, because a wrap that did not would read as a repeat; and on a plane whose gather cache holds part of a very long list it says that too — `todo 3/20 of 400: …`, counted through a new shared `slots.todo_total` so the row and the `▪ todos N` heading cannot disagree about a clipped cache.

**A key rather than a wheel, and that ordering is the design.** `[frame] mouse` ships off, so a pointer-only answer would be inert on exactly the planes this was reported from — and `docs/frame.md`'s own rule runs that way anyway: a pointer affordance always has a key or a palette row beside it, so the keyboard half is the one that has to exist. A wheel over the section can be added on top of this; it could not have stood in for it.

**The `…(+N more)` row still answers nothing**, and a test asserts it: it stands for todos it does not name, so there is no todo for a click on it to be about — the same case the persona column's own overflow row carries.

It is free when nobody presses it: `cmd_palette` already loads the gather snapshot to draw itself, and the row writes no file, bumps no version and starts no process. The cursor is a cell closed over in the registry, which `build` makes fresh per palette — so it starts at the top each time `F2` is pressed, and no `state.bump` wakes four panels to move a number nobody else reads.

## Verification

* Full suite green on the rebased tree: **9,808 tests, 0 failures** (`python3 -B -m unittest discover -s tests`). `test_frame_border_surface` now passes — #785's fix in #799 landed, so that failure is no longer an expected worktree artefact and this branch does not produce one.
* **Every new guard was checked to fail without its change**, not merely to pass with it: `_ROW_DROPS` with `top` back in (16 failures), `visible_slots` without the rows test (11), `_table_min_rows` as a literal `2` (1), `todo.next` unregistered (13), the fraction counted off the clipped list (2), no wrap (1), `repeat=False` (1), `todo_total` ignoring `todo_count` (3), `_HEAD_ROWS = 2` (45), `_TABLE_MIN_ROWS = _HEAD_ROWS` (2), the title fallback dropped (1), `max`→`min` in `todo_total` (4), the `isinstance` guard dropped (4).
* **Two lines were rewritten because the sweep would have been right about them.** `todo_total` moved out of `_todo_rows` carrying `raw if raw >= len(items) else len(items)` — at `raw == len(items)` both arms answer the same number, so `>=` versus `>` is unobservable and no test could ever turn it red. It is `max(raw, len(items))` now, with no branch to get wrong. And the `isinstance` beside it survived its own deletion until a test fed a `todo_count` that is not a number — which a JSON cache another process wrote can be, and which would otherwise have raised `TypeError` on a sidebar repaint.
* No rendered string was changed by this branch, so no `assertNotIn` needle was deleted out from under an absence assertion. The one absence assertion added (`assertNotIn("repos", …)`) is paired with the `assertIn` from the other side of the same boundary, and the `top` entry removed from a `gone` tuple in `test_frame_bars.py` is replaced by an explicit `assertIn("top", kept)`.
* Real tmux on 3.7c and the 3.2 floor, own socket, reaped.

### DIFF
--- a/charter/frame/builtin_actions.py
+++ b/charter/frame/builtin_actions.py
@@ -255,6 +255,128 @@ def _select(ctx, step: int, start: int):
     return f"selected {name}"
 
 
+#: What a workspace with nothing open is told instead of a row that does nothing.
+#: :data:`NO_REPOS`' shape and its argument — the row carries the fix, because a refusal
+#: that names no way out is a row an operator reads once and never again. `charter
+#: workspace todo "…"` with a quoted argument RECORDS one; bare it lists (`cli._wire`),
+#: which is why the sentence quotes the recording form and not the reading one.
+NO_TODOS = ("this workspace has nothing open to read out — "
+            "charter workspace todo \"<what to do>\"")
+
+
+def _register_todos(reg: actions.ActionRegistry) -> None:
+    """One row that reads this workspace's open todos out, one press at a time — **the
+    keyboard route to the sidebar's `…(+N more)`** (#742).
+
+    **The defect this closes is a row with no route beside it.** `slots._todo_rows` draws
+    `…(+5 more)` on a sidebar too short for seven todos, and until this row those five were
+    reachable only by typing `charter workspace todo` into the harness — the surface the
+    frame exists to replace. The sidebar's persona column got its route in #765 (a click
+    switches, a click on the badges explains); the todo list had no gesture at all, and its
+    own overflow line must go on having none, because a row standing for items it does not
+    name cannot resolve to one (`slots._Chips.hit`'s rule).
+
+    **A KEY and not a wheel, and that ordering is the whole design.** `[frame] mouse` ships
+    off, so a pointer-only answer is inert on exactly the planes this was reported from —
+    and `docs/frame.md`'s rule runs the other way anyway: a pointer affordance always has a
+    key or a palette row beside it, so the keyboard route is the half that must exist. A
+    wheel over the section can be added on top of this later; it cannot stand in for it.
+
+    **Why the palette's HEADER rather than a list surface of its own.** `commands_frame
+    ._again` argues this in full for the repo-selection rows and every word applies: the
+    overlay pane is zoomed over the whole window, so the sidebar is not on screen behind
+    the palette, and `Invocation.note` on the header is the surface the operator is
+    actually looking at. A read-only list surface would be the nicer answer and is a real
+    change — a fifth `frame/choose.py` noun breaks that module's "a noun is a thing you
+    switch to" contract at the point where picking a row tries to switch to it.
+
+    **`repeat=True`, so five hidden todos cost five keypresses and one palette.** Without
+    it each Enter closes the pane, kills the process and re-splits for the next — the
+    fourteen-keystroke, three-cycle cost #746 measured on the repo rows.
+
+    **The cursor lives in a CELL closed over here, and that is not laziness about state.**
+    `_select` records the repo selection through `state` because a PANE has to redraw from
+    it: two processes, one fact. Nothing redraws from this — the answer is the sentence
+    `run` returns — so a `state` write plus the `state.bump` that wakes every panel would
+    be four panels repainting per keypress to move a number nobody else reads. The
+    registry is rebuilt per palette (`build`'s "a fresh registry per call") and
+    `commands_frame._again` invokes through that same registry, so the cell lives exactly
+    as long as the palette the operator is reading and starts again at the top next time —
+    which is what an operator who opened `F2` to read their todos means.
+
+    A list rather than an `itertools.count`: :func:`_read_todo` needs the value it is
+    about to report, wrapped against a list whose length it only learns from `ctx`.
+    """
+    seen = [0]
+    reg.register(action.Action(
+        id="todo.next", title="todo: read the next open todo",
+        touches=("todos", "gather"), repeat=True,
+        run=(lambda ctx: _read_todo(ctx, seen)),
+        available=lambda ctx: bool(_open_todos(ctx)),
+        reason_unavailable=lambda ctx: NO_TODOS))
+
+
+def _open_todos(ctx) -> list[dict]:
+    """The todos on *ctx* that are actually rows — `slots._todo_rows`' own filter.
+
+    A gather cache is a JSON file written by another process and may hold anything;
+    `ctx.todos` contains it no further than making the list a tuple (`ctx.SERVES` says the
+    containment is shallow in as many words). Asked in one place so `available` and
+    :func:`_read_todo` cannot disagree about whether there is a row to read — a row listed
+    as available that then answers "nothing" is the shape `ActionRegistry._check` builds one
+    ctx to prevent.
+    """
+    return [t for t in ctx.todos if isinstance(t, dict)]
+
+
+def _read_todo(ctx, seen: list[int]) -> str:
+    """The next open todo, as one sentence for the palette's header.
+
+    Split out of the row above for :func:`_select`'s reason: the walk is exercised against
+    a list of titles with no palette, no tmux server and no frame directory under it.
+
+    **The total is `slots.todo_total`, READ rather than counted here**, and #742 is why it
+    is worth a shared function: `gather._MAX_TODOS` bounds the LIST the cache holds while
+    `todo_count` records what was there before the bound, so counting `ctx.todos` would
+    report `3/20` under a sidebar heading saying `todos 400`. One answer, two surfaces.
+
+    **It wraps, and the position is what makes the wrap readable.** `_select`'s argument
+    for wrapping holds — a row that visibly does nothing at the end of a list reads as
+    broken and costs a whole `F2` to find out — and `3/7` is what stops a wrap looking like
+    a repeat. The clip is stated too: on a plane whose cache holds twenty of four hundred,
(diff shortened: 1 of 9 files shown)


diazoxide:
**Sweep on `629ec88f`: `1 survivor`, it was mine, and it is fixed in `ca9900a5`.**

The plan charged this branch correctly — `diff against 9224a682: 3 file(s), 354 added line(s)`, 41 mutations over `builtin_actions.py`, `layout.py` and `slots.py`, no #776 over-charge — and 40 of 41 were pinned. The one that lived:

```
charter/frame/layout.py:693  in `visible_slots`  [drop-conjunct]
  "repos" in keep and not repos_fits_rows(keep, window_rows=rows)
  — is the `"repos" in keep` half pinned?
```

It was right. The body was `keep = [s for s in keep if s != "repos"]`, which rebuilds an identical list when `repos` is not in it — so the membership half could not change an outcome, and no input could ever turn it red. It was decoration copied from the line above it, and I had added a second copy of that line rather than noticing.

The fix is not a test for an equivalent mutant; it is the duplication going away. The two drops were one decision with two reasons, so they are one `if` now, and the membership test does real work instead of guarding a no-op:

```python
if "repos" in keep and not (repos_fits(keep, window_cols=cols)
                            and repos_fits_rows(keep, window_rows=rows)):
    keep.remove("repos")
```

`remove` raises on a frame with no table, so deleting that conjunct is 4 errors rather than a green suite — and it reads as *take the table out* rather than *rebuild the list without it*.

Checked before pushing:

* **Behaviour is identical**, not merely "tests still pass": `visible_slots` was compared old-against-new over **915,876 combinations** — every permutation of the six slot names, twelve widths across the `_LEFT_W` and `min-cols` boundaries, thirteen heights across the drop boundaries, three `min-cols`/`min-rows` pairs. Zero differences.
* **All three conjuncts are now killable**: membership → 4 errors, the width half → 4 failures, the rows half → 11 failures.
* Full suite green: 9,808 tests, 0 failures.

Issue #698, opened 2026-08-30
retune-string inverts a character range, so a regex holding a digit class is a mutation the sweep cannot score


diazoxide:
Found on #693, whose gate read `no verdict: 2 not measured` while the other three branches
of the same night read `no survivors`. Naming the two turned out to be a finding about the
tool rather than about the branch.

## The two mutations

```
[1/11] UNRESOLVED  charter/commands_frame.py:289 [retune-string] r"@[0-9]+"    (46 module(s))
[2/11] UNRESOLVED  charter/commands_frame.py:288 [retune-string] r"\$[0-9]+"   (46 module(s))
…
  charter/commands_frame.py:288  in `<module>`  [retune-string]  no tests ran
  charter/commands_frame.py:289  in `<module>`  [retune-string]  no tests ran
```

## Why they could not be scored

`retune` shifts each character one along inside its class. For a digit that is
`0 -> 1` and `9 -> 0`, and inside a character class that **inverts the range**:

```
$ python3 -c "import sweep, re; print(repr(sweep.retune(r'\$[0-9]+')))"
'\\$[1-0]+'
$ python3 -c "import re; re.compile(r'\$[1-0]+')"
re.error: bad character range 1-0 at position 3
```

So the mutant is not a program. `re.compile` raises while `charter/commands_frame.py` is
being imported, every selected test module fails to load, **zero tests run**, and the
sweep — correctly — refuses to call that a pin. Reproduced by hand on the branch: applying
the tool's own mutant to `_SESSION_ID_RE` and running one covering module ends in
`re.PatternError: bad character range 1-0`, with no test result at all.

**The report's prose is wrong about the reason, though.** The UNRESOLVED section says:

> Twice the run could not be measured: it timed out rather than failing. … Re-run these on
> a quieter machine, or raise the timeout, before reading this sweep as clean.

It did not time out. The per-mutation label right below it says `no tests ran`, and the
timings agree — the second of the two finished in **32 seconds**, against ~55s for a
`retune-string` on a plain constant with the same 46-module selection. A quieter machine
and a bigger timeout will produce this same result forever.

## It is not specific to that branch

Any `[0-9]` (or `[a-z]`, `[A-Z]`) in a regex the tool reaches has this shape.
`tmuxctl.PANE_ID_RE` is `r"%[0-9]+"` and retunes to `r"%[1-0]+"`. Those constants are not
in a recent diff, so the sweep has not asked about them yet; it will the day one of them
moves.

[...]

Issue #670, opened 2026-08-29
The selection map's trace cost is recorded twice and not identically — 250 s in sweep.py, 350 s in sweep.yml


diazoxide:
The same measurement — one trace of the whole suite, which is the largest fixed cost a deletion-sweep shard pays — is written down twice with different numbers:

- `tools/sweep.py:2161` (inside `SHARD_FIXED`'s budget note): `` #: * the selection map, traced — **250 s**, and restored from cache — under 5 s; ``
- `.github/workflows/sweep.yml:72` (the cache step's comment): `# The selection map is one trace of the whole suite — 350 s measured, and the`

Nothing downstream breaks on it: `SHARD_FIXED` is `12 * 60` = 720 s and the itemised costs come to 508 s at 250 and 608 s at 350, so the budget covers either reading. But it is a number a reader will quote, and today there is no way to know which one to quote.

The release notes entry that cited 250 s (`docs/news/unreleased-a-green-sweep-is-not-no-survivors.md`) has been changed to name both figures and say they have not been reconciled, rather than assert one. Re-measuring on `ubuntu-latest` and collapsing the two comments to one number would let that entry state a figure again.

PR #804, opened 2026-09-01
A shard too big to finish says what it measured, instead of saying nothing and calling it success (#803)

### LINKED ISSUE #803: Past 224 mutations the sweep cannot answer, and the no-verdict it returns instead reports success
Past **224 mutations** the deletion sweep cannot return a verdict, and the `no verdict` it
returns instead concludes **`success`**. The gate predicts this at plan time, warns, and runs
anyway.

## Measured, on run `33500900581` (PR #796, head `36996e6`)

```
Size the sweep        11:09:49 → 11:15:10   success
Sweep shard 1..8      11:15:12 → 12:15:2x   cancelled   ← 60 min, to the second, all eight
Add up what the shards found                success
no verdict: 8 of 8 shards did not report    success
```

The plan job had already said so, in the same run:

> `231 mutations → 8 shard(s) of at most 29 each`
> ⚠ `231 mutations is past the 224 that 8 shards can measure inside 40 minutes each … a shard
> may be cancelled at the job timeout, and a cancelled shard reports no verdict rather than a
> short one.`

## The arithmetic, and the mismatch inside it

* `tools/sweep.py:2876` — `MAX_SHARDS = 8`, documented as *"a ceiling on machines and never on
  questions, so a diff past it…"*
* `tools/sweep.py:2823` — `SHARD_BUDGET = 40 * 60`, what `per_shard()` sizes each shard to fit.
* `.github/workflows/sweep.yml:184` — `timeout-minutes: 60`.

So the ceiling is `8 × per_shard()` ≈ **224 mutations**, and this diff was 231. **Any branch of
this size gets this result**; it is not a signal about the branch.

Note the two numbers do not agree with each other. `sweep.py` plans a shard to 40 minutes; the
workflow kills it at 60. A shard that overruns its own budget therefore gets twenty further
minutes of runway and *then* dies with nothing to report — the worst of both, since the same
wall clock spent inside the budget would have produced a partial answer.

## Why the green tick is the dangerous half

`no verdict: 8 of 8 shards did not report` concluded **`success`**, because `--gate` is
non-enforcing. On the checks list it sits beside seven real passes. This repository has closed
this exact shape three times:

* **#617** — a green check was not "no survivors". Fixed by #630 putting the verdict in the
  check **name**.
* **#646/#561** — a gate that never ran looked like one that found nothing. Fixed by making
  `Add up what the shards found` a **required** check, so a missing row blocks.
* **#782** — a sweep that planned zero mutations still said `no survivors`. Fixed by
  `nothing to sweep`.

This is the fourth: a sweep that **ran and could not finish**, reporting `success`. The name is
honest — that part works — but nothing stops a merge, and the name is the only thing that says
so.

## Three fixes, and the trade between them

1. **Raise `MAX_SHARDS`.** Simplest. Costs runners, and `MAX_SHARDS`'s own docstring calls it a
   ceiling on machines — so this is the number it was written to be raised.
2. **Make `timeout-minutes` agree with `SHARD_BUDGET`.** Right regardless of the above: two
   numbers describing one deadline should not differ by 20 minutes. Note this makes a shard
   die *sooner*, so it is a fix for the mismatch and not for the ceiling.
3. **Have a shard that will not fit report what it measured.** The current design is all-or-
   nothing, and #630's whole argument was that a partial answer honestly labelled beats silence.
   *"23 of 29 measured, 0 survived"* is worth more than nothing, and it is what a killed shard
   already has in hand when it dies.

**Whichever is taken, the enforcement question is separate and worth settling with it:** should
`no verdict` block a merge? It currently does not, and every one of the three fixes above still
leaves a way to reach it.

diazoxide:
Closes #803.

## The defect, measured

Run [`33500900581`](https://github.com/diazoxide/charter/actions/runs/33500900581) (PR #796, head `36996e6`):

```
Size the sweep        11:09:49 → 11:15:10   success
Sweep shard 1..8      11:15:12 → 12:15:2x   cancelled   ← 60 min, to the second, all eight
Add up what the shards found                success
no verdict: 8 of 8 shards did not report    success      ← reports SUCCESS
```

231 mutations, eight shards, every one killed at `timeout-minutes: 60`. **How much answer
that threw away is countable, from the shards' own logs** — they print a line per mutation:

```
Sweep shard 1 of 8:  17 of 29 measured, 10 survived
Sweep shard 2 of 8:  22 of 29 measured,  8 survived
Sweep shard 3 of 8:  27 of 29 measured, 12 survived
Sweep shard 4 of 8:  19 of 29 measured,  8 survived
Sweep shard 5 of 8:  12 of 29 measured, 10 survived
Sweep shard 6 of 8:  21 of 29 measured,  7 survived
Sweep shard 7 of 8:  16 of 29 measured,  9 survived
Sweep shard 8 of 8:  18 of 28 measured,  7 survived

TOTAL: 152 of 231 mutations measured, 71 survivors printed
```

**Two thirds of the branch had been answered and 71 survivors had been printed when the
eight processes were killed.** All of it went with them, because the result set lived in
memory until the sweep's last line. The pull request was told `success`.

(71 is the raw `SURVIVED` count as each shard printed it, before the merge step sorts
survivors into `unpinned` / `masked` / `platform-deferred`. The actionable number is some
part of it, and nobody has ever seen which part.)

The plan job predicted it in the same run and proceeded — and the warning was accurate
rather than alarmist: `MAX_SHARDS = 8`, `per_shard() = 28`, so `8 x 28 = 224`, and **any
branch past that got this result regardless of what was in it.**

## Can a cancelled job report at all? Yes — measured, and it already did

This was the question that decided the design, so it was answered from the run rather than
from the documentation:

```
$ gh api repos/diazoxide/charter/actions/runs/33500900581/jobs \
    -q '.jobs[] | select(.name|test("Sweep shard 1 ")) | .steps[]'

"Sweep the guards this branch adds"              cancelled   11:15:17 -> 12:15:27
"Keep the result set, whatever the sweep said"   success     12:15:27 -> 12:15:27   <- ran
```

**The `if: always()` upload step ran on every one of the eight cancelled shards and
concluded `success`.** GitHub was never the obstacle. It uploaded nothing because there was
no file: `sweep.py` held the whole result set in memory until its last line, so
`if-no-files-found: ignore` silently decided what a cancelled shard says, and it said
nothing. `artifacts?total_count` on that run is `0`.

Reproduced locally too: a shard on a 38-mutation slice, killed 25 minutes in, left **no
`--json` at all**.

## The fix — option 3, plus a correction to option 2

**A shard reports what it measured.** The result file is written *before the first
mutation* — the whole plan, every row marked `out of time` — and rewritten (atomically, via
`os.replace`) as each answer lands. So a shard killed at any instant leaves the **complete
plan with holes in it**, not a short list that would read as a complete sweep of a short
plan. And it stops on its own ten minutes before the runner's cap, so the ordinary case is
a clean exit rather than a kill.

The check on a pull request now reads

```
deletion sweep / no verdict: 23 of 29 measured, 6 out of time
```

which is #630's own argument turned on itself: *a partial answer honestly labelled beats
silence.*

`out of time` is **its own bucket**, not folded into `unresolved` — #698's rule again.
`unresolved` is *I looked and could not tell*; re-run it. `out of time` is *I never
looked*; re-running the same plan on the same fan-out stops at the same place, so what
answers it is a smaller branch or more machines. It ranks with a **missing shard** and
above a survivor in `gate_conclusion`, because "3 survivors" said over a third of a plan is
a number nobody should quote — without that, this fix would have *downgraded* the verdict
it was written to protect.

### What was rejected

**Raising `MAX_SHARDS` (option 1).** Its docstring calls it a ceiling on machines and never
on questions, so it is the number written to be raised — but raising it moves the ceiling
without removing it, and the next large diff hits the same wall with more runners burned.
The ceiling is now *survivable*, which is the property that was missing.

**Lowering `timeout-minutes` to 40 (option 2 as stated).** The issue is right that the
twenty-minute gap was a defect — a shard that overran its sizing got twenty further minutes
of runway and *then* died with nothing. But the two never described one deadline, and
**setting them equal would have been a regression**: `SHARD_BUDGET` is a sizing estimate
built from an *average* mutation, and a shard whose slice holds five survivors pays a
full-suite run for each and legitimately needs longer. Cutting it off at forty turns a
complete answer arriving at forty-eight minutes into a partial one — on exactly the
branches that have something to report. Cutting the *runner* off at forty is worse: it
kills the shard at the instant it stops to write.

So the gap is closed by a **third** number between them:

| constant | value | what it decides |
|---|---|---|
| `SHARD_BUDGET` | 40 min | sizes the plan — how many mutations a machine may be dealt |
| `SHARD_REPORT_AT` | 50 min | when the shard stops and writes what it measured |
| `SHARD_TIMEOUT` | 60 min | the runner's cap, now in `sweep.py` and asserted against the YAML (#670) |

`SHARD_REPORT_AT > SHARD_BUDGET` is asserted, because that inequality is what says a
correctly sized shard is never cut short by this.

## Should `no verdict` block a merge?

**Recommended, and deliberately not done here** — enforcement is a repository setting and
the operator's call. The reasoning is on #803; the short version is that `collect` is
already the required check and `--enforce` already has a distinct exit code (`3`) for
exactly this, so it is one flag and no new machinery. But it should wait until a few real
branches have published a partial verdict and the numbers have been read and believed,
which is the same staging argument `gate_exit_code` makes about itself.

## Verification

* `tests.test_sweep` + `tests.test_workflows`: 491 tests, green.
* **Every guard added here was removed one at a time and the named tests watched go red** —
  18 un-guards, 18 red. One assertion was found vacuous that way
  (`assertIn("if: always()", yaml)` matches the `collect` job) and replaced with one that
  reads the step's own block.
* Ceiling reproduced before fixing: a synthetic 605-line diff planned
  `300 mutations -> 8 shard(s) of at most 28 each` with the over-budget warning, and a
  shard of that plan killed mid-run left nothing behind.
* The tool run on itself — `tools/sweep.py --path tools`, since CI's sweep charges
  `charter/` only and its verdict is worthless for this change — reported in a comment
  below.

Do not merge on CI alone: the `deletion sweep` check on this pull request charges `charter/`
and this branch does not touch it, so it will say `nothing to sweep`. That is correct and
it is not evidence about this change.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/sweep.yml
+++ b/.github/workflows/sweep.yml
@@ -176,20 +176,21 @@ jobs:
     name: Sweep shard ${{ matrix.shard }} of ${{ needs.plan.outputs.shards }}
     needs: plan
     runs-on: ubuntu-latest
-    # **The backstop, and since #803 nothing but the backstop.** `sweep.py` sizes a shard
-    # for `SHARD_BUDGET` — forty minutes — and now STOPS there and writes what it
-    # measured, so this number is only what catches a shard that could not keep its own
-    # deadline: a hung subprocess, a mutant that took the machine (#773).
+    # The runner's cap, and `sweep.py` reads it: `SHARD_TIMEOUT` is this number, and
+    # `SHARD_REPORT_AT` is ten minutes short of it — the point at which a shard stops
+    # dealing mutations and writes what it measured (#803). One deadline written in two
+    # files and drifting is #670, so the tool holds the constant and
+    # `test_sweep.TheReportingDeadlineSitsBetweenTheBudgetAndTheBackstop` holds this line
+    # to it. Change one and the suite asks for the other.
     #
-    # Twenty minutes behind the budget, and the gap is asserted by
-    # `test_sweep.TheWorkflowsBackstopIsBehindTheBudget` rather than left to two files
-    # agreeing by habit. **Do not close the gap by lowering this to forty.** #803 reads
-    # the mismatch as a defect and it was one — a shard that overran its budget used to
-    # get twenty further minutes of runway and *then* die with nothing — but the two
-    # numbers no longer describe one deadline, and setting them equal would have the
-    # runner kill the shard at the exact moment it stops to write its answer. The room
-    # is for the mutation still in flight when the budget expires (`FULL_TIMEOUT` bounds
-    # it) and for the file to be written and uploaded.
+    # **Do not lower this to `SHARD_BUDGET`'s forty.** #803 reads the twenty-minute gap
+    # as a defect and it was one — a shard that overran its sizing got twenty further
+    # minutes of runway and *then* died with nothing. But the two were never one
+    # deadline: forty sizes the PLAN, from an average mutation, and a shard whose slice
+    # holds five survivors pays a full-suite run for each and legitimately needs longer.
+    # Cutting it off at forty would turn complete answers into partial ones, and cutting
+    # the runner off there would kill the shard at the instant it stops to write. What
+    # closed the gap is the third number, between them.
     timeout-minutes: 60
     strategy:
       # One shard's trouble must not cancel the others: three quarters of an answer, said
--- a/.github/workflows/sweep.yml
+++ b/.github/workflows/sweep.yml
@@ -176,11 +176,20 @@ jobs:
     name: Sweep shard ${{ matrix.shard }} of ${{ needs.plan.outputs.shards }}
     needs: plan
     runs-on: ubuntu-latest
-    # The backstop and not the budget. `sweep.py` sizes a shard for forty minutes; this is
-    # what catches a box that turned out to be slower than that, and it is deliberately
-    # the same hour the unsharded job had — the fix for running out of time is more
-    # machines, not a longer rope, because a shard cancelled here reports nothing at all
-    # and a longer rope only moves where that happens.
+    # **The backstop, and since #803 nothing but the backstop.** `sweep.py` sizes a shard
+    # for `SHARD_BUDGET` — forty minutes — and now STOPS there and writes what it
+    # measured, so this number is only what catches a shard that could not keep its own
+    # deadline: a hung subprocess, a mutant that took the machine (#773).
+    #
+    # Twenty minutes behind the budget, and the gap is asserted by
+    # `test_sweep.TheWorkflowsBackstopIsBehindTheBudget` rather than left to two files
+    # agreeing by habit. **Do not close the gap by lowering this to forty.** #803 reads
+    # the mismatch as a defect and it was one — a shard that overran its budget used to
+    # get twenty further minutes of runway and *then* die with nothing — but the two
+    # numbers no longer describe one deadline, and setting them equal would have the
+    # runner kill the shard at the exact moment it stops to write its answer. The room
+    # is for the mutation still in flight when the budget expires (`FULL_TIMEOUT` bounds
+    # it) and for the file to be written and uploaded.
     timeout-minutes: 60
     strategy:
       # One shard's trouble must not cancel the others: three quarters of an answer, said
@@ -220,13 +229,25 @@ jobs:
             --shard "$SHARD/$SHARDS" \
             --workdir "$RUNNER_TEMP/sweep" \
             --json "sweep-results-$SHARD.json"
+      # **This step already worked, and that is the measurement #803 turns on.** On run
+      # 33500900581 all eight shards were cancelled at `timeout-minutes` and this step
+      # ran on every one of them, concluding `success` — GitHub does run an `always()`
+      # step after a job timeout. It uploaded nothing because `sweep.py` held the whole
+      # result set in memory until its last line, so the killed process took the answer
+      # with it and the run published `no verdict: 8 of 8 shards did not report`.
+      #
+      # `--json` is now written as each answer arrives, so what this finds is whatever
+      # the shard had measured at the instant it died. `if-no-files-found: warn` and no
+      # longer `ignore`: a missing file used to be the ordinary outcome of a long sweep
+      # and is now an anomaly worth a line in the log — the shard died before it reached
+      # its first mutation, during the map or the baseline.
       - name: Keep the result set, whatever the sweep said
         if: always()
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: deletion-sweep-${{ matrix.shard }}
           path: sweep-results-${{ matrix.shard }}.json
-          if-no-files-found: ignore
+          if-no-files-found: warn
 
   # ------------------------------------------------------------------------------------
   # 3. One answer, from however many shards came back.
--- a/docs/news/unreleased-a-sweep-that-cannot-finish-says-what-it-measured.md
+++ b/docs/news/unreleased-a-sweep-that-cannot-finish-says-what-it-measured.md
@@ -33,8 +33,8 @@ result regardless of what was in it**. It is not a signal about the branch.
 
 ## What was fixed, and what was deliberately not
 
-**A shard now stops at its own budget and reports what it measured.** Forty minutes after
-the run started it deals no further mutation; the ones it never reached come back
+**A shard now stops short of being killed and reports what it measured.** Ten minutes before
+the runner's own cap it deals no further mutation; the ones it never reached come back
 `out of time`, named, in the same result set as the answers. The check on the pull request
 reads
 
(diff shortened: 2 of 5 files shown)


diazoxide:
## The tool on itself — and the fix proving itself by accident

`tools/sweep.py --path tools`, on this branch, because CI's sweep charges `charter/` and
this branch does not touch it. (CI's own verdict here is `nothing to sweep`, which is
correct and is #782 working, and is not evidence about this change.)

**The run was killed by the machine at mutation 13 of 39** — load average 16, two agents
sweeping at once, the process gone with `exit 144`. That is the exact shape of the defect
this pull request is about, arriving unrehearsed. Before this change the file it wrote
would not have existed. What it actually left on disk:

```
$ python3 -c 'sweep.classify(sweep.results_from_json(open("self.json").read()))'

rows in the killed shard file: 39
Counter({'out_of_time': 26, 'pinned': 9, 'survived': 4})

check name would be: no verdict: 4 survivors so far, 13 of 39 measured, 26 out of time
conclusion         : no-verdict
```

Nine pinned, four survivors, twenty-six named as never reached, from a process that was
killed. The old code would have published `no verdict: 1 of 1 shards did not report`, with
a conclusion of `success`, and the four findings would have been lost.

### The four survivors were real, and all four are now pinned

The sweep found them in this pull request's own new code:

| operator | line | what was not pinned |
|---|---|---|
| `retune-string` | `OUT_OF_TIME = "out_of_time"` | the **wire spelling**. That string is written by one process and read verbatim by another (`results_from_json` takes `row["verdict"]` as-is), so a shard built by an older sweep than the merge agrees on this literal or does not agree at all. Every test said `sweep.OUT_OF_TIME`, so renaming it kept them all green. |
| `shift-boundary` | `now() >= deadline` | the boundary. `>` and `>=` differ by exactly one mutation — the one dealt at the instant the budget expires, which `>` sends to a sandbox for a full-suite run on a machine that is about to be killed. The scripted clocks elsewhere step over the boundary rather than landing on it. |
| `drop-if` | the `record(results)` after the evidence pass | the third write. `evidence` is what `_summary_rows` prints under every survivor and it does not exist at the second write; a file without it round-trips as `naming: null`, which the merge step reads as "the evidence pass never ran" — true of a killed shard and a lie about a finished one. The test that watches the file watches it from inside `decide`, so it could not see this one. |
| `collapse-ifexp` | `f" of {len(results)}" if short else ""` | the denominator. `mutations applied : N` is #782's line and a reader knows it; it gains a denominator exactly when there is something to be short of, because `0 of 0` on a docs branch is a question where `0` was a statement. |

Each is now pinned by a test that was run against **the exact mutation the tool applied**
and watched go red.

## What the check name will say

```
all measured, clean       -> clean       no survivors
partial, clean so far     -> no-verdict  no verdict: 23 of 29 measured, 6 out of time
partial with survivors    -> no-verdict  no verdict: 3 survivors so far, 23 of 29 measured, 6 out of time
nothing reached           -> no-verdict  no verdict: 0 of 29 measured, 29 out of time
nothing to sweep          -> nothing     nothing to sweep
```

Note the third row: `out of time` ranks **with a missing shard and above a survivor**, so a
partial sweep never publishes a survivor count as if it were the branch's total. Ranked with
`unresolved` instead, this change would have turned run `33500900581`'s honest
`no verdict` into a confident `71 survivors` measured over two thirds of the plan — a
downgrade wearing the shape of a fix.

## The un-guard ledger

Every guard added here was removed, one at a time, and the named tests watched go red.
**22 un-guards, 22 red.** One assertion was found vacuous that way and replaced:
`assertIn("if: always()", yaml)` is green with the upload step's condition deleted, because
`collect` carries an `always()` too.

diazoxide:
## The sweep's verdict on this branch: `no survivors`

Three runs of `tools/sweep.py --path tools` over this branch, each one over the tree the
previous one criticised. The rule that produced this is the repository's own: *after fixing
a class of defect, re-run the measurement that found it over the fixed tree.* It paid out
twice here.

| run | | measured | pinned | survivors | wall clock |
|---|---|---:|---:|---:|---:|
| 1 | **killed by the machine at 13 of 39** | 13 | 9 | **4** | — |
| 2 | complete, over the tree that fixed those 4 | 39 | 32 | **7** | 47.9 min |
| 3 | complete, over the tree that fixed those 7 | **37** | **37** | **0** | 21.6 min |

```
mutations applied : 37
pinned            : 37
SURVIVED          : 0
UNRESOLVED        : 0
wall clock        : 21.6 min

gate: 0 unpinned, 0 in masked cluster(s), 0 platform-deferred, 0 unresolved,
      0 out of time, 0 not applied, 0 withheld, 37 pinned
gate: no survivors
```

The plan is 37 and not 39 because three of run 2's survivors were closed by **deleting**
rather than by testing.

### What run 2 found, and what was done about each

Seven survivors, every one of them in reporting code this pull request adds.

**Four pinned**, because each is a fact a reader acts on:

* the terminal report's `OUT OF TIME : N` count line, and its `OUT OF TIME — these
  mutations were planned and never measured` section. Both were deletable with the whole
  suite green, because the tests until then read the *check name* and the *markdown page*
  and nobody read the report a person gets at a terminal.
* the run summary's `| **out of time** | N |` table row. Deleting it left a table whose
  rows all read zero, directly above a section saying three quarters of the branch was
  never measured.
* the twenty-line cap on the unswept list, **and its boundary**. A truncated list that
  does not say it was truncated is the silent-truncation shape this whole tool exists to
  refuse, so the cap announces itself exactly as `ANNOTATION_CAP` does.

**Three deleted.** They were all inside one log line — a `max(0.0, left)` clamp twice over
and a `.0f` format spec — computing how much of the budget remained after the map and the
baseline. None of it was load bearing and none of it was worth a test. This repository's
rule is that *"equivalent mutant" and "dead code" are one finding*, and the honest answer to
three survivors in a log line is a shorter log line. The deadline is the fact; a reader can
subtract.

### And run 1's four, for the record

`OUT_OF_TIME`'s wire spelling, the `>=` in `now() >= deadline`, the file rewrite after the
evidence pass, and the report's conditional denominator — all four pinned, each against the
exact mutation the tool applied.

## The un-guard ledger, final

**27 un-guards, 27 red.** Every guard this branch adds was removed one at a time and the
named tests watched go red — including all eleven the sweep itself found, using the tool's
own mutation as the un-guard. One assertion was found vacuous that way and replaced.

diazoxide:
**This merged at `5a2e7fe`, one commit before the tool was run on itself.** The two commits
that close what that run found are not on `main`, so `main` is carrying **eleven unpinned
guards** in the reporting code this pull request added.

They are in #805, rebased onto `main`, along with the paragraph this branch's news entry is
missing. Nothing in it changes behaviour except three deleted lines in one log message;
everything else is tests.

The short version of what the sweep found:

| run | | measured | pinned | survivors |
|---|---|---:|---:|---:|
| 1 | **killed by the machine at 13 of 39** | 13 | 9 | **4** |
| 2 | complete, over the tree that fixed those 4 | 39 | 32 | **7** |
| 3 | complete, over the tree that fixed those 7 | **37** | **37** | **0** |

Run 1 is this issue arriving unrehearsed — the process was killed mid-sweep and the file it
left behind said `no verdict: 4 survivors so far, 13 of 39 measured, 26 out of time`. Under
the code that was on `main` before this merge, there would have been no file.

One other thing worth knowing: **GitHub swallowed the `pull_request` event on this branch's
last two pushes.** The runs list has `push` runs for `fec2e38` and `e0b8f4d` and no
`pull_request` run for either — #646/#561 happening live, and the reason to read
`gh api .../commits/<head>/check-runs` rather than `gh pr checks` before a merge. The head
sha's own checks are green (four Pythons, plus a `workflow_dispatch` sweep); CodeQL's last
run on this branch was on `5a2e7fe`.

Issue #780, opened 2026-08-31
A launch spends ~76% of its pre-attach time in ~42 sequential tmux round trips, 29 of which read nothing back


diazoxide:
Split out of #728, which was filed as a display defect ("nothing says the launch is
starting") and is not one — during the blank there is no client attached, so there is no
surface to write on. What is left of it is latency, and it is measurable and mostly
avoidable.

## Measured

One real `charter frame` launch on the private-server path, instrumented, on a fast
machine with a four-slot plane:

```
tmux invocations before the attach        42
wall time inside them                    293 ms      (~7 ms each)
launcher wall clock at the attach        383 ms
```

so **~76% of the time before anything can be on screen is sequential tmux round trips**.
The operator who filed #728 measured the same shape at ~4x the scale on their machine —
1.76 / 1.87 / 1.77 s to the first rule over three runs, with an independent decomposition
putting ~856 ms of it in ~64 round trips at ~13.4 ms each.

## They are not sequential by necessity

By verb, on the run above:

```
set-option 18   set-hook 7   set-environment 4     <- 29 of 42 read nothing back
split-window 4  list-windows 2  display-message 1  <- 13 that do, or that order matters for
list-sessions 1 list-panes 1  select-window 1
select-pane 1   source-file 1   -f 1
```

**29 of the 42 are fire-and-forget writes** whose results nothing reads. `tmuxctl.chain`
already exists to send several commands as one invocation — its docstring documents tmux
parsing and executing the whole list server-side — and `commands_frame` already uses it for
the overlay close. The obvious groups:

- `_dress_window`: `remain-on-exit -w` plus the five `_CHROME` window options plus #657's
  two `-p` rules — 8 calls, one chain, no ordering change (the whole group must still
  precede the first `split-window`, which chaining preserves).
- `_arm_panel_respawn`: one `set-hook` per pane.
- The per-pane `set-option`s that follow each `split-window` — they need that split's pane
  id, so they chain per pane rather than globally.

That is on the order of 30 round trips removed with no reordering of anything that reads a
result.

## The other half

charter's own import is paid once by the launcher, once by the detached `frame-gather`
child, and **once per panel pane** — seven or more times per launch on the shipped frame.
It is the only lever that also shortens the empty-box window *after* the attach, since that
window is exactly the panel processes' startup.

## Why this is its own issue

It has no display component, it wants its own before/after numbers on both tmux 3.7c and
[...]

PR #816, opened 2026-09-02
The chat bar's + makes a chat, instead of naming the command that would

diazoxide:
**Stacked on #815** — base is `tabs-strip`. Merge that first; this branch contains it.

> *"`+` button not working for creating new session."*

It was never a button. The chat bar's add-chat affordance was a **sentence**:

```
  chats  *api.1  + charter <harness> opens another
```

True, names the command that does it, and sits at the end of a row of clickable tabs beginning with a `+`. Every terminal an operator has used puts a `+` there and every one of them means *new*. So it got pressed.

```
  chats  *api.1   api.2   +
```

## Why it needed a new command

The obvious fix — have the panel run `charter <harness>` — does not work, and the way it fails is instructive. A click's work runs detached with all three streams on `/dev/null`, and `cmd_launch` reads a non-tty stdout as *this process cannot be the operator's terminal* and `os.execvp`s the bare harness into the void. No frame, no chat, and no process left to say so.

`attach=False` is the seam that says *build the frame, do not become the terminal*. It has existed since `_open_workspace` needed it for §4k and was reachable only in-process. `charter frame-new-chat` is the first spelling that can ask for it, and `cmd_new_chat` is `_open_workspace`'s shape with one guard **inverted**: that one refuses a workspace whose name is already a live session; this one refuses to proceed unless `_plane_session` can prove the session is this plane's. Same §3.3 rule, opposite sign, written out rather than shared because a helper taking a flag for which way to point would be one function claiming to answer two questions.

## Four stops, four sentences

All on the frame's attention row, because a `+` that silently failed would be this report one release later: a frame inside the operator's own tmux, a session this plane cannot prove is its own, a chat recording no launchable harness, and a workspace directory charter cannot enter.

## Two decisions worth the paragraph

**The `+` is drawn at every chat count now, not only at one.** The old rule was right about a *sentence* — a reminder for a plane that had not discovered chats — and wrong about a *button*: an operator with two chats wanting a third is exactly who presses it. It cannot share a row with a `+N`, structurally: the ladder draws the affordance only on the rung where every name fits and a count only on the rung where they do not.

**The workspace bar gets none.** A new chat is nothing but a press; a new workspace is a directory and a *name*, and #518 refuses to create one on a typo. The handler is handed the command as **data** rather than deriving it from what the renderer drew, so the property survives a renderer that grows a note later — pinned by a case that publishes the map by hand and asks the workspace handler directly.

## Verification

- Full suite: **10238 tests, 8 skipped, green.**
- Real tmux on **3.7c and the 3.2 floor** (20 cases in `test_a_real_click_on_a_real_tab_bar_switches`, green on both): a real client on a real pty, the `+` pressed through the client's own terminal, the answer written on the frame's own attention row by a real detached `charter frame-new-chat` that resolved which frame it belonged to, established it was on charter's own server, and proved the workspace's session was this plane's with a real `list-panes`. Plus: the keyboard stays on the harness, the cell next to the `+` reaches nothing, an unpaired release reaches nothing, and a stopped press created nothing.
- Mutation-checked: dropping either guard, `attach=True`, dropping the handler's `add` data, not publishing the affordance's columns, and not passing the note each turn the suite red.

**One stated limit.** The launcher itself is asserted against a stand-in rather than run for real: `commands_frame.SOCKET` is a module constant naming the one shared server every frame on the machine runs on, so a test that let the launch run would build a session on the operator's live charter server. Exactly one call is faked; everything up to it is real on both tmux versions.

## Note on file ownership

This touches `charter/commands_frame.py`, which another agent holds this session for the #780 tmux round-trip batching. The change is **purely additive** — one new function appended at the end of the file, plus two module constants — so it should merge cleanly, but it is worth knowing before this lands beside that work.

https://claude.ai/code/session_014ctfuhnX3LdXbtx3SMUvTu

### DIFF
--- a/charter/cli.py
+++ b/charter/cli.py
@@ -769,7 +769,7 @@ def _add_frame_parsers(sub) -> None:
                                          "frame-probe", "frame-respawn", "frame-density",
                                          "frame-resize", "frame-gather", "frame-switch",
                                          "frame-toggle", "frame-chrome", "frame-chat",
-                                         "frame-quit", "frame-close",
+                                         "frame-new-chat", "frame-quit", "frame-close",
                                          "frame-transcript"}
 
     # Which harness (by `.name`, never `.cli_name` — that's the dict key below) has
@@ -1004,6 +1004,28 @@ def _add_frame_parsers(sub) -> None:
     ch.add_argument("chat_id")
     ch.add_argument("--chat", dest="chat", default="")
     ch.set_defaults(func=commands_frame.cmd_chat)
+
+    # The chat bar's `+`, and the one thing no other spelling of `charter <harness>` can
+    # ask for: a launch that must NOT become the operator's terminal. `cmd_launch` has
+    # taken `attach=False` since `_open_workspace` needed it (§4k), and until now it was
+    # reachable only in-process — so a panel that wanted to add a chat had `charter claude`
+    # and nothing else, which on a detached child with `/dev/null` for stdout takes
+    # `bypass` and `os.execvp`s a bare harness into the void.
+    #
+    # **A sibling of `frame-chat` rather than a flag on it.** They are two different
+    # operations that happen to share a noun: one moves this client to a window that
+    # exists, the other makes one. `frame-chat`'s positional would have to become optional
+    # to carry a `--new`, which makes `charter frame-chat` alone — a plausible typo —
+    # parse into something that CREATES.
+    #
+    # **No name and no positional at all**, which is #518's line held: the chat's id is
+    # allocated (`state.workspace_prefix`) and its workspace is the one this chat is in for
+    # life (§4j), so there is nothing here for an operator to type and nothing to validate.
+    # `--chat` is where the press happened, exactly as it is for `frame-toggle` and
+    # `frame-quit`, and it falls back to `$CHARTER_SESSION_ID`.
+    nc = sub.add_parser("frame-new-chat")
+    nc.add_argument("--chat", dest="chat", default="")
+    nc.set_defaults(func=commands_frame.cmd_new_chat)
 
     # §4i's quit. Started DETACHED by the palette's confirmation row
     # (`commands_frame._start_leaving`) and typeable by hand, which is why it warns on its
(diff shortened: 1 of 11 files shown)


diazoxide:
Note for anyone reading this later: this merged into `tabs-strip`, **not into `main`** — GitHub did not retarget it when #815 was squash-merged, so `git merge-base --is-ancestor 6470663 origin/main` answers NO and none of this content is on main.

`tabs-strip` has since gone stale enough that merging it would revert #819, #820, #821 and #823: `git diff --stat origin/main origin/tabs-strip` is 1521 insertions and **2936 deletions**, including all 459 lines of #820's test file.

The same three commits are rebased onto `origin/main` at 6e89277 and reopened as **#825**, with base `main`. There, `git diff --diff-filter=D --name-only origin/main <branch>` is empty. Please take #825 and leave `tabs-strip` alone.

diazoxide:
**Correcting my comment above: the "merging `tabs-strip` would revert #819/#820/#821/#823" claim was false.** I withdraw it, and I apologise for putting a wrong statement about other people's work on a merged PR.

It rested on a **two-dot** `git diff --stat origin/main origin/tabs-strip`, which compares *tips* — so a branch merely behind main is charged for every file main gained since the merge base, as though merging it would delete them. A merge applies `merge-base..head`, not `tip..tip`. Measured:

```
git diff --stat            origin/main..origin/tabs-strip   ->  40 files, 1522+, 4023-
git diff --stat            origin/main...origin/tabs-strip  ->  16 files, 2539+,  112-
git diff --diff-filter=D --name-only origin/main...origin/tabs-strip  ->  (empty)
```

**Nothing would have been reverted, and #820's test file was never at risk.** The tell is that the two-dot deletion count grew from 2936 to 4023 while the branch did not change — main had simply moved on. Use three dots for this assertion.

What stands from the comment above is the only part that mattered: this PR's content is genuinely **not on main** — `git merge-base --is-ancestor 6470663 origin/main` answers NO and `cmd_new_chat` is absent from `origin/main` — so it does still need taking through **#825**, which is based on `main` and shows no deleted files under the three-dot check.

PR #818, opened 2026-09-02
The two tab strips get a keyboard route: F2, next tab, Enter

diazoxide:
**Stacked on #816** — base is `plus-makes-a-chat`, which is itself stacked on #815. Merge in order: #815, #816, this.

The tab bars became clickable, and clicking was the only way to use them. That is backwards for the rule charter states to every provider author who might add a pointer affordance of their own, in `component.EVENT_KINDS`:

> A component whose only route to a piece of state is a click has no route to it on most planes. **Give every pointer affordance a key as well.**

`[frame] mouse` is off by default, and a bar can never grow a keyboard of its own — `key` is in `EVENT_KINDS` and deliberately not in `events.DELIVERED`, because tmux routes typing to the *active* pane and that pane is the harness the frame exists to protect. `builtin_actions._register_selection` is charter keeping that rule for the repo table. These four rows are charter keeping it for the two strips.

```
    repo: select the next row
    repo: select the previous row
    chat: the next tab                 <- new
    chat: the previous tab             <- new
    workspace: the next tab            <- new
    workspace: the previous tab        <- new
```

Each walks the strip **one step in the order the strip is drawn**, wrapping, and spawns the argv a click on that tab already spawns — so the keyboard and the pointer cannot come to two answers about what "the next tab" is, or about what happens on arrival, refusals included. `_walk` is factored out of `_select` so the repo table and the two strips walk by one function.

## Three decisions worth the paragraph

**They do not keep the palette open, unlike the repo rows.** Those write two files and leave it standing. A switch moves the *client* to another window, and the palette's pane is in the window being left — a row that asked to stay open would be asking to stay open somewhere the operator no longer is.

**"Nowhere else to go" is reported by the RUN, not by `available`.** `test_frame_palette_names.TheRosterIsNeverReadUntilSomethingIsTyped` pins that opening the palette reads no roster at all — *`F2` on a plane with forty workspaces costs what it cost with none* — and `available` is asked for every row on every draw. An availability that asked `switch.workspaces()` would enumerate the plane on every `F2`, including the ones opened to detach. So the walk reports it: a step that lands where it started starts nothing and says the sentence instead. There is a new case asserting the palette still reads no workspace roster, so this cannot be reintroduced quietly.

**That test is also sharper than a count.** `len(names) > 1` reads *is there anywhere to go* as *are there two places*, and those differ on a real state: a frame whose recorded workspace has been deleted is on no name at all, so the one workspace left **is** somewhere to send it — and a count would refuse the only row that could.

## Verification

- Full suite: **10261 tests, 8 skipped, green.**
- Mutation-checked: no wrap, one start for both directions, the two commands swapped, a count for availability, no spawn, a flipped step, and dropping the strip-of-one report — each turns the suite red.
- The two argv spellings are held equal to `builtins._CHAT_SWITCH` / `_WORKSPACE_SWITCH` by a case, since `_STRIPS` respells rather than imports them (a palette process has no other reason to load a renderer module).

https://claude.ai/code/session_014ctfuhnX3LdXbtx3SMUvTu

### DIFF
--- a/charter/frame/builtin_actions.py
+++ b/charter/frame/builtin_actions.py
@@ -45,7 +45,7 @@ import subprocess
 
 from typing import Callable, NamedTuple
 
-from .. import contain, util
+from .. import util
 from . import action, actions, chats, choose, state, switch, tmuxctl
 
 #: What marks the density a frame is currently on. **One constant, not two**: it is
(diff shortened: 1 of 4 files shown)


diazoxide:
Superseded — this branch's base was never `main` and it had gone stale enough that merging it would have reverted #819, #820, #821 and #823 (2936 deletions, including all 459 lines of #820's test file). Rebased onto `origin/main` at 6e89277 and reopened against **main**: #816 → #825, #818 → #826. Both now show no deleted files against main.

Issue #829, opened 2026-09-02
Tab strip: hover, multi-row, activity marks and the third colour state — what was left out of #815/#825/#826, and why


diazoxide:
Follow-on work for the tab strip, from the session that shipped #815 / #825 / #826. Three
things were deliberately left out, and two findings will be re-litigated by whoever picks
this up unless they are written down. Everything below was measured on **tmux 3.7c and the
3.2 floor** (`~/.local/share/charter-testing/tmux-3.2`), with a real client on a real pty.

The operator's own words for the whole piece of work were:

> *"now we are showing plain text, no colors, no hover effect, no active tab/session color,
> `+` button not working for creating new session, when workspaces are more we are showing
> `+N` now in tabs — but user can't click and see other workspaces, we don't have any
> separator borders. All together we need to have one intuitive and harmonic design."*

Five of those six are answered by #815 (block on the active tab, rules-gated separators,
clickable `+N`), #825 (the `+` makes a chat) and #826 (a keyboard route). **Hover is the
sixth and is the first item below.**

---

## 1. Hover — a measurement, with the product call left open

**Not a decision. The operator has not ruled and this issue does not rule for them.**

Every mechanism claim holds. Measured with three panes — a `cat` harness asking for nothing,
a "bar" pane asking `1006,1003`, and a second panel asking `1006,1000` — reports injected
into the client's pty, **identical on 3.2 and 3.7c**:

| gesture | bar (1003) | other panel (1000) |
|---|---|---|
| motion over the bar | `\x1b[<35;6;1M` | — |
| motion over the other panel | — | — |
| motion over the harness | — | — |
| press over the bar | `\x1b[<0;6;1M` | — |
| wheel over the bar | `\x1b[<64;6;1M` | — |

So: **1003 is delivered to a pane that is not active; it cannot poison the harness; it does
not wake other panels; and it does not cost the pane its clicks or its wheel.** What tmux
asks the outer terminal for, with its own mouse on: `?1003h ?1006h` on 3.2, and
`?1000h ?1002h ?1003h ?1006h` on 3.7c — so the pane's own 1003 is what propagates. With
tmux's mouse **off** it asks for nothing at all, so hover needs `[frame] mouse = true`
exactly as clicks do.

### The blocker: there is no pointer-LEAVE signal

When the pointer leaves the bar, the next report goes to whichever pane it entered, and
that pane is not the bar. Measured: motion over the harness reaches nobody, and motion over
[...]

PR #840, opened 2026-09-03
A tab strip grows a row when its tabs overflow, and gives it back when they fit (#829)

diazoxide:
Closes item **2** of #829.

> *"panes are resizable — and user can resize and it should show more tabs opened on new resized rows."*

On this plane's own fifteen workspaces, rung 1 needs **274 columns**. At 160 the strip drew seven names and counted the rest; at 120, five. Widening the terminal bought nothing, because the pane was one row and stayed one row.

```
before, 160 columns
  workspaces   authority-audit   autonomy   charter-update-skill   default   fleet  *harness-wrapper   news-dispatch-guard   opencode-integration  +7

after, 160 columns
  workspaces   authority-audit   autonomy   charter-update-skill   default   fleet  *harness-wrapper   news-dispatch-guard   opencode-integration
               plane-shape   relations-and-delegations   showcase   statusline-improvements   todos   tracking-github-issues   user-reporting
```

## The shape

A panel cannot make its own pane taller, so this is a **launcher size** and not a renderer setting:

* `slots.bar_rows_wanted(fid, slot, *, pane_cols, cap)` — how many rows the names need. It **asks the renderer**: it composes the strip at 1, 2, 3 rows and stops at the first height that holds every name, so the sizer and the renderer cannot come apart the way #500 shipped twice one pane over. Measured over every mark × every width × two caps: the pane is **never** sized taller than the strip fills.
* `layout.slot_sizes(..., bar_rows=...)` → `layout._grown` — spends the harness's **surplus** (`harness_rows` of the ungrown map minus `HARNESS_MIN_ROWS`), a row at a time in split order so two strips share a short budget. `repos` is not what pays: `_DROP_ORDER` gives the bars up before it.
* `layout.BAR_MAX_ROWS = 3` — measured through the same cut: these fifteen need 2 rows at 160, 3 at 120, 4 at 100, 6 at 80.
* `layout.pane_cols(order, slot, *, window_cols)` — `repos_cols` generalised, so a strip split after `right` is measured at its own pane's width (#500, one slot over). `repos_cols` delegates to it.
* `commands_frame._slot_sizes` gains `order` and `window_cols`; all three call sites (both launches, `_relayout`, `_reassert_sizes`) already had them.

**It grows only when the tabs overflow.** A fixed two-row launch would cost a row off the harness on every plane that places a bar whether or not the names overflow — a cost everyone pays for a case only some hit, and rows against the harness are what #740 settled once already.

## Clicks

`slots.TABS` was keyed by COLUMN. On a two-row strip that does not degrade to answering nothing — it answers the row above about a click on the row below, which is `EVENT_KINDS`' *fires wrongly*. The map is `(row, col)` now, **with no default row**: a caller that forgot which row it is asking about is a `TypeError`, not a switch to the wrong tab.

Rows are cut the way the single row always was: `_cuts` gives the page boundaries, pages are grouped into runs of `rows`, and the strip draws the run holding the mark. A run is a function of the names, the width and the row count alone, so pressing a drawn tab redraws the same run with only the `*` moved (pinned at 1/2/3 rows × 5 widths × every drawn name).

**At `rows=1` the renderer is byte-identical to `main`** — differentially checked over 7 name lists × every `here` × both note settings × widths 0–299, rows and click map together.

## The 3.2 floor, measured rather than assumed

* The launch is **not gated**: `split-window -l 3` is version-free, so a strip is born at the height its names need on 3.2 as on 3.7c.
* `_install_resize_hook` issues **no command at all** below `RESIZE_HOOK_FLOOR` — confirmed on `~/.local/share/charter-testing/tmux-3.2`, where `set-hook -w window-resized …` answers `invalid option: window-resized`, rc 1.
* Driving `_reassert_sizes` against a real 3.2 server gives `{top: 1, workspaces: 3, bottom: 1, repos: 6}`, harness 25 — **byte for byte what 3.7c does**. So the floor loses only the automatic trigger; `charter frame-resize`, which `tmuxctl.below_resize_hook_message` already names, re-sizes the strip there.

## Tests

* `tests/test_a_tab_strip_grows_a_row_when_its_tabs_overflow.py` — 40 cases: the want, the growth budget, the harness floor, the ceiling, the pad, the ASCII property at every width × every row count, and the run-stability property.
* `tests/test_a_real_resize_gives_a_real_strip_another_row.py` — 5 cases on a **real tmux server**, including a real `charter panel workspaces` process redrawing into a row it did not have. Without `ctx.height` that case fails with the `+7` still on screen in a two-row pane.

## Two things the issue got slightly wrong

* The constant is `tmuxctl.below_resize_hook_message`, not `below_resize_floor_message`.
* `chats`/`workspaces` are **not** in `layout.SLOT_SIZE` — `_PLACED` filters on `c.id in SLOT_OF` — so `layout` derives no geometry for them at all; their height comes from the committed arrangement (#687). The effect is what the issue says; the mechanism is the arrangement echo, and it is why a bar has to be *placed* for any of this to be about it.

https://claude.ai/code/session_016VxfcoaaCZBq3rTNeb6e2R

### DIFF
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -2091,6 +2091,23 @@ def _slot_sizes(fid: str, slots: list[str], *, window_rows: int, pane_cols: int,
     for `layout.repos_rows`' reason about *pinned_rows*: a policy applied at both ends is a
     bound one end cannot make observable.
 
+    **Every slot is asked, and there is deliberately no `slot in BARS` filter in front of
+    it.** There was one, and the deletion sweep reported dropping it as a survivor — the
+    two spellings of "a slot that draws no strip wants nothing" cannot both be observable,
+    and the one that stays is the one that DOES the answering. `bar_rows_wanted` answers
+    `1` for a name it has no strip for; no slot can ever be sized below one cell
+    (`component.cells` refuses it and `layout.repos_rows` floors the table at
+    `SLOT_SIZE["repos"]`), so a want of `1` is at or under every slot's own height and
+    `layout._grown` grows nothing for it. Measured over 3,648 sizings — four pin
+    configurations, three split orders, thirty-eight window heights and eight widths —
+    filtered and unfiltered answer the identical map every time, and an unfiltered
+    comprehension asks about `top`, `bottom`, `repos` and `right` exactly once each and is
+    told `1` for all four.
+
+    Deleting it is also what makes that answer REACHED: the refusal in `bar_rows_wanted` is
+    a production path now rather than a defensive line, and
+    `tests/test_a_tab_strip_grows_a_row_when_its_tabs_overflow` asks it directly.
+
     *order* is the split ORDER, which is not always *slots* — `_relayout` re-splits a
     permutation of the recorded panes and it is that permutation which decides how wide a
     pane ends up (#500). It is what the strips' own widths are derived from
@@ -2109,7 +2126,7 @@ def _slot_sizes(fid: str, slots: list[str], *, window_rows: int, pane_cols: int,
         bar_rows={slot: frame_slots.bar_rows_wanted(
             fid, slot, cap=layout.BAR_MAX_ROWS,
             pane_cols=layout.pane_cols(order, slot, window_cols=window_cols))
-            for slot in slots if slot in frame_slots.BARS})
+            for slot in slots})
 
 
 def _launch_sizes(fid: str, slots: list[str], *,
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -2049,8 +2049,8 @@ def _spawn_gather(fid: str, ws: str) -> None:
         return
 
 
-def _slot_sizes(fid: str, slots: list[str], *,
-                window_rows: int, pane_cols: int) -> dict[str, int]:
+def _slot_sizes(fid: str, slots: list[str], *, window_rows: int, pane_cols: int,
+                order: list[str], window_cols: int) -> dict[str, int]:
     """`layout.slot_sizes` for a frame that has a plane behind it — the only place in
     charter where a committed arrangement becomes a pane height.
 
@@ -2082,11 +2082,34 @@ def _slot_sizes(fid: str, slots: list[str], *,
     the frame did not have, and `_reassert_sizes` on every `window-resized`. They differ
     only in how the table pane's width is arrived at, which is why that arrives here as
     *pane_cols* already measured rather than as a window to measure.
+
+    **A tab strip is a third such fact and it arrives the same way** (#829). How many rows
+    a strip needs is a function of the names this plane has — `chats.roster`,
+    `switch.workspaces` — so it is a plane read, which is what this function is the
+    boundary for, and `frame_slots.bar_rows_wanted` is the one place it is made. The
+    ceiling on it is `layout.BAR_MAX_ROWS`, carried down from here rather than read there,
+    for `layout.repos_rows`' reason about *pinned_rows*: a policy applied at both ends is a
+    bound one end cannot make observable.
+
+    *order* is the split ORDER, which is not always *slots* — `_relayout` re-splits a
+    permutation of the recorded panes and it is that permutation which decides how wide a
+    pane ends up (#500). It is what the strips' own widths are derived from
+    (`layout.pane_cols`), the same list the caller already hands `layout.repos_cols` for
+    the table's, and it is a separate argument rather than *slots* reused precisely because
+    those two lists differ on one of the three paths.
+
+    Nothing extra is read for a frame that places no bar, which is charter's own and very
+    nearly everyone's: the comprehension below is empty and `layout.slot_sizes` is handed
+    an empty map it does not spend.
     """
     return layout.slot_sizes(
         slots, window_rows=window_rows,
         content_rows=frame_slots.repos_rows_wanted(fid, pane_cols=pane_cols),
-        pinned_rows=layout.pinned_repo_rows())
+        pinned_rows=layout.pinned_repo_rows(),
+        bar_rows={slot: frame_slots.bar_rows_wanted(
+            fid, slot, cap=layout.BAR_MAX_ROWS,
+            pane_cols=layout.pane_cols(order, slot, window_cols=window_cols))
+            for slot in slots if slot in frame_slots.BARS})
 
 
 def _launch_sizes(fid: str, slots: list[str], *,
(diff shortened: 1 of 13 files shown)


diazoxide:
Filed the `test_claims` collision this branch hit as #842 rather than patching the matcher here — `_REDACT_VERB` carries `strip` and `_UNIVERSAL` carries `each`, so "each tab strip" reads as a redaction promise. The last commit rewords the one sentence; the matcher is #842's.

diazoxide:
## The sweep's verdict, answered

`no verdict: 1 survivor so far, 5 not measured; 39 of 48 measured, 9 out of time` — three findings, three different answers. Rebased onto `5bc12f7` first.

### 1. The survivor is **deleted**, not pinned

`slots.bar_rows_wanted` had an early `return rows` the moment the click map held every name, sitting beside the line that records the tallest height the strip fills. Dropping it survived because **the two answer the same number, always**:

* a run holds every name exactly when the cut has no more pages than the strip has rows;
* at that height the run *is* the whole list, so the strip fills every row it was given and the remaining line records it;
* a greater height cannot beat it — there are no more pages to draw, so the strip draws fewer rows than it is offered and that height is not recorded.

Measured rather than argued: **406,000 inputs** (32 name lists, every mark, every width 0–300, five ceilings) — zero differences. An equivalent mutant and dead code are the same finding, so it is gone, and the reasoning is in the docstring where the next reader will meet it.

### 2. All five unmeasured are **non-termination** — caught, not unpinned

Every one is `layout._grown`'s `while budget > 0 and any(n > 0 for n in short.values())`. Each mutation leaves a loop whose body cannot change its own condition. Reproduced, with the state that hangs each:

| mutation | hangs on |
|---|---|
| drop `any(…)` | nothing wants to grow; a want equal to the size; one strip wants two rows; a want below the size |
| drop `budget > 0` | both want, budget is short |
| `budget > 0` → `budget >= 0` | both want, budget is short |
| `n > 0` → `n >= 0` | a want equal to the size; one strip wants two rows |
| `any` → `all` | nothing wants to grow (`all([])` is vacuously true) |

The original terminates on all five. This is the #837 shape, where the right answer was to say so rather than bound the loop — bounding turns a caught mutant into a silent survivor.

**That is not what has been done here.** The deal is counted off the budget itself: one round hands out at least one row while demand and budget remain, so `range(budget)` is more rounds than the deal can ever need. The bound is *data*, not a guard:

* `range(budget - 1)` under-deals → **1 failure**
* dropping `short[slot] > 0` → **40 failures**
* dropping `budget > 0` → **4 failures**
* no mutation of the block can hang

Equivalence to the `while` proven over **1,593,480 inputs**.

### 3. The nine out of time follow from those five

Each hanging mutant burned a full per-mutation timeout; `Sweep shard 1 of 2` was cancelled at its report deadline. Removing them is most of the fix. The surface is smaller too — **48 mutations → 32**:

* the equivalent early return, gone;
* the four single-row rungs restored to the lines they have on `main`, behind a `row` wrapper, so they are no longer charged to this diff;
* the per-row composition flattened from six conditional expressions to two statements about the run (the leading count belongs to the first row and the trailing one to the last, which is a fact about the *run*, not a test to re-ask per row);
* one local no longer spelled `strip`, which `tools/sweep.py`'s `swap-synonym` reads as the string method — the same word collision as #842.

### What is left is not this branch's to fix — #843

`tests/test_frame_tmux_integration` is **27.6 s** and is **82% of the cost** of a frame-area selection; it appears in **651 of the selection map's 16,545 entries**. Against `SECONDS_A_MUTATION = 60` the real cost measured here was ~120 s. #843 has the numbers and proposes #841's fix (hold the server rather than rebuilding it) applied to that module — which would also take 27 s off every `unittest discover` on every Python in the matrix.

`SHARD_BUDGET`/`timeout-minutes` headroom deliberately left alone.

diazoxide:
## `3 survivors` — one deleted, two pinned

First, the part of the previous round that worked: the verdict went from

```
no verdict: 1 survivor so far, 5 not measured; 39 of 48 measured, 9 out of time
```

to a plain `3 survivors` — **both shards completed and every mutation was measured**. Removing the five non-terminating mutants and cutting 48 → 32 was the whole of that.

### Ruled out first: the multi-row path is exercised

Two of the three were `collapse-ifexp` on a multi-row conditional, which would be a much bigger finding if `r == 0` were always true. It is not. Instrumenting `slots._compose` across the six bar modules (263 tests):

```
lines RETURNED: {0: 5539, 1: 38829, 2: 29922, 3: 9547, 4: 2602, 5: 2531,
                 6: 3466, 7: 667, 8: 606, 9: 605, 10: 76, 11: 531, 12: 536,
                 13: 927, 14: 868}
```

29,922 compositions return two rows and 9,547 return three, so `under` is reached tens of thousands of times. Both survivors are missing *properties*, not a missing path.

### 1. `commands_frame.py:2112` `slot in frame_slots.BARS` — **deleted**

It and `bar_rows_wanted`'s `entry is None` refusal are two spellings of one rule, and only one can be observable. The one that stays is the one that *does the answering*: `bar_rows_wanted` answers `1` for a name it has no strip for, and no slot can ever be sized below one cell (`component.cells` refuses it; `layout.repos_rows` floors the table at `SLOT_SIZE["repos"]`), so a want of `1` is at or under every slot's own height and `_grown` spends nothing on it.

Measured over **3,648 sizings** — four pin configurations, three split orders, thirty-eight window heights, eight widths — filtered and unfiltered answer the identical map every time, and an unfiltered comprehension asks about `top`, `bottom`, `repos` and `right` exactly once each and is told `1` for all four. Re-adding the filter now changes no test.

Deleting it also promotes the refusal from a defensive line to a production path.

### 2. `slots.py:3844` `lead if r == 0 else under` — **pinned**

`under` is `" " * tui.width(lead)`, so the heading and its blank are the **same width by construction**. Every column assertion in the suite is unmoved by drawing one in place of the other; only the text tells them apart, and nothing read the text. The property — *the heading names the strip once* — is a case now, and it kills the mutation **in both directions** (3 failures each way).

### 3. `slots.py:3948` `note if not leading and not trailing else ""` — **pinned**

Collapsing it does two things, and the second is the dangerous one:

* the affordance rides a run that starts mid-list, so `+11 … +` share a strip;
* `afters[-1]` is `trailing or tail`, so when a trailing count exists the *text* is unchanged — but `add = _span(bottom, ends[bottom], tail)` now claims the cell the `+N` is drawn in. `_bar_events` asks `switch_to`, then `add_at` — so a click asking to **see** the other chats would **make** one, and the row looks identical either way. That is `EVENT_KINDS`' *fires wrongly*.

Two cases: the named run at 120 columns with the mark on the third page, and a sweep over every width × row count × three marks asserting no cell answers `more_at` and `add_at` together. **301 failures** under the mutation, 1 the other way.

### The wrapper caution

Checked: `row`'s `[(before, body, drawn)] if body else []` is killed in both directions (51 failures one way, 16 failures + 43 errors the other), so the wrapper made nothing vacuous.

Nothing was bounded or simplified to make a survivor go away. **31 mutations now, from 48.**

Issue #713, opened 2026-08-30
`ChromeIsOneColour` flakes under load on `main` — 10 of 45 runs, eight different cases, all at `_screenshot`'s `server exited unexpectedly`


diazoxide:
`tests/test_frame_tmux_integration.py`'s `ChromeIsOneColour` flakes under load, on `main`,
independently of #694. Found while measuring #694 (PR #712) and deliberately not fixed
there: it is a different mechanism, in a class neither of #694's tests is in.

## Measured

Paired legs, both worktrees run simultaneously under 10 spin loops on a 14-core darwin box,
`python3.14`, tmux 3.7c, 45 runs each:

| leg | runs | failed |
|---|---|---|
| `main` @`d40d998` | 45 | **10** |
| PR #712's branch | 45 | **6** |

The branch touches nothing this class reaches, so the two numbers are one number: **~18%
of loaded runs of this class fail**. Unloaded it is green — 8931 tests OK on 3.14 and on
3.12 on the same machine.

It is not one test. Across those 90 runs, eight distinct cases failed:

```
4  test_the_frame_draws_every_rule_the_same_inside_the_operators_own_tmux
3  test_the_frame_does_not_move_when_focus_does
2  test_the_unfixed_frame_really_does_draw_its_rules_in_two_colours
2  test_no_shipped_design_leaves_a_seam_between_two_panes_of_one_colour
2  test_an_unsurfaced_rule_really_does_leave_a_seam_between_painted_panes
1  test_the_unset_really_does_take_the_harnesss_edges_back_off
1  test_the_previous_design_really_does_draw_one_rule_in_two_colours
1  test_every_rule_in_a_surfaced_frame_is_one_colour
```

## The failure

Every one of them is the same line — `_screenshot`'s own gate, not the colour assertion the
test is about:

```
File "tests/test_frame_tmux_integration.py", line 4478, in test_the_frame_draws_every_rule_the_same_inside_the_operators_own_tmux
  own = self._screenshot(arm=True)
File "tests/test_frame_tmux_integration.py", line 4336, in _screenshot
  self.assertEqual(r.returncode, 0, r.stderr)
AssertionError: 1 != 0 : server exited unexpectedly
```

`server exited unexpectedly` is tmux's own words for a client whose server went away
mid-command. So the subject is the class's fixture — the forked pty client and the server
it is attached to — and not what any of these eight tests assert about colour.

## Why it is worth a number

[...]

PR #864, opened 2026-09-03
The tmux integration tests stop waiting for the next look (#843)

### LINKED ISSUE #843: The sweep's per-mutation estimate is half the real cost wherever test_frame_tmux_integration is in the selection
The deletion sweep sizes a shard at `SECONDS_A_MUTATION = 60`. On #840 the measured cost was about **twice** that, and the run reported

```
no verdict: 1 survivor so far, 5 not measured; 39 of 48 measured, 9 out of time
```

with `Sweep shard 1 of 2` cancelled at its report deadline. The branch's own answer was to cut the surface (48 → 32 mutations, and the five that did not terminate are gone), but the reason the estimate is out is not specific to that branch and will not go away.

## The measurement

`tests/test_frame_tmux_integration.py` takes **27.6 s** on a workstation (`python3 -m unittest tests.test_frame_tmux_integration`, tmux 3.7c, warm). It starts a real tmux server per test.

Every other module in a typical frame-area selection, together, is under six seconds:

```
 0.22s  tests.test_a_click_on_a_tab_bar_switches
 0.66s  tests.test_a_panel_pane_exists_once_per_placed_component
 0.70s  tests.test_frame_bars
 0.42s  tests.test_frame_chat_switch
 1.83s  tests.test_frame_density
 0.20s  tests.test_the_keyboard_walks_the_tab_strips
27.56s  tests.test_frame_tmux_integration      <- 82% of the selection's cost
```

So a mutation anywhere the selection reaches that module costs ~33 s locally before anything else runs. CI is slower: #840's shards measured roughly **120 s** a mutation against the 60 s the sizer assumes.

## The blast radius

From a real selection map (16,545 entries), **651** of them name `tests.test_frame_tmux_integration`:

```
 75  charter/commands_frame.py
 40  charter/instance.py
 34  charter/frame/state.py
 32  charter/frame/layout.py
 21  charter/workspace.py
 16  charter/config.py
```

Any branch touching the launcher, the layout arithmetic, the frame state or the plane paths pays it on every mutation. That is most frame work.

## Why this is worth fixing rather than absorbing

`per_shard()` is `(SHARD_BUDGET - SHARD_FIXED) // SECONDS_A_MUTATION` = **28**. A branch with 29 mutations therefore gets two shards of at most 28 — and if its selection is tmux-heavy, each shard needs about 56 minutes for 28 mutations against a 40-minute budget. The failure is silent in the sense that matters: it reports `no verdict`, which is neither a pass nor a survivor, so the author cannot tell an unmeasured mutation from a killed one without downloading the artifacts.

Raising `SECONDS_A_MUTATION` is the wrong fix: it shrinks `per_shard()` and buys more shards, but `MAX_SHARDS = 8` still caps the total, and every OTHER branch then gets more runners than it needs.

## What would fix it

**#841's change, applied to `tests/test_frame_tmux_integration.py`.** That issue's fix — the workspace-tab tests hold their tmux server instead of rebuilding it between themselves — is exactly the shape here: `_TmuxServerFixture.setUp` stands up a server per test and `_teardown_socket` kills it, so a module with dozens of cases pays dozens of server starts. Cases that only ever read geometry (`test_a_bar_this_plane_placed_does_not_take_the_tables_rows`, `test_the_table_panes_width_is_tmuxs_answer_and_not_the_recorded_order`, the resize cases) could share one server and one window per class, the way #841 now does, and the ones that genuinely need a private server keep one.

That is a change to the suite's slowest module, so the win is not only for the sweep: `unittest discover` pays the same 27.6 s on every run, on every Python version in the matrix.

Filed from #840, where it cost nine mutations their measurement.

diazoxide:
Closes #843.

`tests/test_frame_tmux_integration.py` is the suite's slowest module and the most expensive
thing a frame-area mutation can select. It was **28.2 s** on a workstation and **21.5 s** on
`ubuntu-latest`, where it is 84% of a frame selection's 25.7 s. It is now **19.3 s** on the
same workstation, with four more tests than it had.

## What the issue proposed, and why it is not the fix

#843 asks for #841's construction — hold the tmux server per class instead of rebuilding it
per test — on the grounds that the module "starts a real tmux server per test". It does, and
that is worth about **1.3 s of 29 s**.

Measured on this machine (tmux 3.7c, darwin), 10 builds and 10 connects on a throwaway
socket:

```
BUILD   (socket file absent): mean 12.2 ms
CONNECT (server standing)   : mean  5.1 ms
kill-server                 :       4.8 ms
delta per birth             :       7.1 ms
```

One run of the module makes 171 `new-session` calls and 118 `kill-server` calls, so the
whole server-lifecycle premium is ~1.3 s of a 29 s module — **4%**, not 82%. That is not a
flaw in #841: #841 and #713 were **flake** fixes (`server exited unexpectedly`, rc 1), and
their headline number — 12 births to 2 — is a count of races removed, not of seconds. #843
transfers a flake fix as a cost fix, and the arithmetic does not carry over.

Two other claims in #843 do not survive measurement either:

* **"CI is slower."** On run 33751547084 (`ubuntu-latest`, 3.12) the module is **21.5 s**,
  *faster* than the workstation's 28.2 s. Its 84% share of the selection does hold: 21.5 s
  against 4.16 s for the other six modules.
* **"roughly 120 s a mutation."** #840's two sweep runs of the same branch measured **84 s**
  and **185 s** a mutation. See below — the difference is not the selection.

## Where the time actually was

Wrapping `subprocess.run` and `time.sleep` for one run of the module:

```
tmux subprocess.run total :   11.58s  (1987 calls)
other subprocess.run total:    0.18s  (17 calls)
time.sleep total          :   16.68s  (154 calls, mean 102 ms)
```

**Fifty-seven per cent of the module was `time.sleep`**, and it is one number repeated
twenty times. Every wait here polls, which is right — `_DEADLINE`'s own note says a wait
"returns the instant its condition holds". Except it cannot: it returns at the next look,
and the next look was a flat `time.sleep(0.1)` in all twenty loops. That tick is a floor
under every wait in the module whether or not anything is slow. On CI, 57 of the module's
102 tests finish inside a single tick, so the floor is most of what the other 45 pay.

The single most expensive test on CI, at **5.0 s of the 21.5**, is the `pane-died` negative
control, which bought a short wait by shortening its deadline to five seconds.

## The change

**1. A wait looks again sooner, then backs off.** `_Poll` replaces the flat tick at all
twenty loops: 2 ms to the first look back, doubling to a 100 ms ceiling. Doubling rather
than a smaller flat number because `_await_client` runs a `tmux list-clients` per look — a
flat fine tick would spawn ~4,000 processes for a client that never registers. A long wait
costs what it always did (~205 looks over a 20 s deadline against 200 before).

**2. A wait for absence is bounded by a fact, not a stopwatch.**
`_hook_reaches_a_shim`'s docstring already carried the argument: *once tmux has destroyed a
pane, `pane-died` for it can no longer fire at any deadline.* That is a fact tmux can be
asked for, so the control now asks it — `settled=destroyed` — instead of guessing five
seconds. Faster **and stricter**: five seconds was paid in full on every healthy run, and a
machine that fired the hook at 5.1 s would have made the control PASS on its timer running
out. Bounded by the fact it keeps the full `_DEADLINE` and still answers in milliseconds.

## What it tolerates: nothing

No sleep added, no retry, no assertion widened, no deadline raised, no return code stopped
being checked, no condition changed. Settling ends the *wait*, never the reading: the marker
is read once more on the way out, so an argv that landed in the same instant is still
reported — and the control still asserts the pane's absence on its own account, so a pane
tmux KEPT fails it rather than satisfying it.

Four new cases, all four red on `origin/main` (`FAILED (failures=1, errors=3)`):

* the first look back is `_POLL_FIRST` and the gap settles at `_POLL_MAX`, read off
  `time.sleep` rather than off a stopwatch;
* a wait told nothing can change asks once and stops instead of spending 20 s;
* a wait that settles still reports what arrived in the same instant;
* the module contains no `time.sleep(<a number>)` outside the five listed waits-for-absence
  — which is the anti-vacuity half: on `origin/main` that test names all twenty flat polls.

## The measurement

Three runs a side, one workstation, tmux 3.7c:

| | before | after |
|---|---|---|
| module wall | 29.6 / 27.5 / 27.3 s | 17.9 / 19.2 / 19.6 s |
| `time.sleep` | 16.68 s over 154 calls | 5.86 s over 191 calls |
| tests | 102, OK | 106, OK |

Full suite, run the way CI runs it (`unittest discover -s tests`): see the comment below —
this clone sits inside the operator's plane, so `test_plane_write_guard`'s checkout pin
fails here for the reason it exists to catch and fails identically on `origin/main`.

## Why the estimate was not the lever instead

`SECONDS_A_MUTATION = 60` really is low — but its error is not selection-shaped, so a
selection-weighted estimate would not have saved the run this issue was filed from. #840's
two sweep runs, same branch and same selection:

| run | plan | dealt | wall for them | per mutation |
|---|---|---|---|---|
| 33722127015 | 32, 2 shards | 16 | 1349 s | **84 s** |
| 33715835528 | 48, 2 shards | 15 of 24 | 2780 s | **185 s** |

The second is the `39 of 48 measured, 9 out of time` run, and its shard 1 was cancelled at
the runner's own 60-minute cap. What made it 185 s was **three UNRESOLVED mutations** —
`decide` runs the subset twice before it will say "unresolved", so each is up to
2 × `SUBSET_TIMEOUT` (900 s) of a worker, roughly half the shard's worker-seconds between
them. No estimate built from the selection map can predict which mutant fails to terminate.
Raising the constant is ruled out by #843 itself (`MAX_SHARDS` caps the fan-out and every
other branch over-provisions). So the surface is the lever that moves every branch honestly,
and `SHARD_BUDGET` and `sweep.yml`'s `timeout-minutes` are untouched.

No production file was touched.

https://claude.ai/code/session_01Us4MbwkSCJCxwt1CjAd1sf

### DIFF
--- a/docs/news/unreleased-the-tmux-integration-tests-stop-waiting-for-the-next-look.md
+++ b/docs/news/unreleased-the-tmux-integration-tests-stop-waiting-for-the-next-look.md
@@ -1,0 +1,94 @@
+---
+version: unreleased
+headline: The tmux integration tests stop waiting for the next look, so the suite's slowest module costs a third less on every run and every mutation
+---
+
+`tests/test_frame_tmux_integration.py` is the slowest module in this suite and the most
+expensive thing the deletion sweep can select — #843 counted it in 651 of the selection
+map's 16,545 entries, so a branch touching the launcher, the layout arithmetic, the frame
+state or the plane paths pays it on every mutation, and pays it twice (`sweep.decide`
+re-runs a subset that goes red, to confirm the red). Measured on `ubuntu-latest` it was
+**21.5 s** of a frame selection's 25.7 s — 84% of it — and on a workstation **28.2 s**.
+
+It is now **19.3 s** on the same workstation, with four more tests than it had.
+
+## Where the time was, measured rather than assumed
+
+Wrapping `subprocess.run` and `time.sleep` for one run of the module:
+
+```
+tmux subprocess.run total :   11.58s  (1987 calls)
+time.sleep total          :   16.68s  (154 calls, mean 102 ms)
+```
+
+**Fifty-seven per cent of the module was `time.sleep`**, at a mean of 102 ms a call, and
+the reason is one number repeated twenty times. Every wait here polls — which is right, and
+`_DEADLINE`'s own note says why: *"every wait below returns the instant its condition
+holds, so the number is what a LOADED runner may take and never what a healthy one does."*
+Except that it cannot return the instant its condition holds. It returns at the next look,
+and the next look was a flat `time.sleep(0.1)` in every one of them. That tick is a floor
+under every wait in the module whether or not anything is actually slow — and on CI, where
+57 of the module's 102 tests finish inside a single tick, the floor is most of what the
+other 45 pay.
+
+## Two changes, and neither of them shortens a wait
+
+**A wait looks again sooner, and then backs off.** `_Poll` replaces the flat tick at all
+twenty polling loops: the first look back is 2 ms, doubling to a 100 ms ceiling. A
+condition that holds in 5 ms is now seen at 6 ms instead of at 100. A wait that runs long
+costs what it always did — about 205 looks over a 20 s deadline against the 200 the flat
+tick took — which is the reason for doubling rather than for a smaller flat number:
+`_await_client` runs a `tmux list-clients` per look, and a flat fine tick would have spawned
+some four thousand processes for a client that never registers.
+
+Every deadline is the number it was, every condition is the condition it was, and a machine
+where a condition never holds spends exactly the time it spent before and fails with the
+same message. What changed is how often the loop looks, which is not a fact about the code
+under test.
+
+**A wait for absence is bounded by a fact instead of by a stopwatch.** The single most
+expensive test in the module — `test_a_panel_dying_here_is_destroyed_unless_charter_arms_
+the_window`, 5.0 s of the 21.5 s on CI — is the negative control for charter's `pane-died`
+funnel: the pane dies on a server at tmux's own defaults, and nothing must reach a shell.
+It bought its right to a short wait by *shortening the deadline to five seconds*, and
+`_hook_reaches_a_shim`'s docstring already carried the better argument for it: **once tmux
+has destroyed a pane, `pane-died` for it can no longer fire at any deadline.** That is a
+fact tmux can be asked for, so it is now asked: the control hands the wait a `settled`
+predicate — "has tmux taken the pane away?" — and the wait ends the moment it holds.
+
+That is faster *and* stricter. Five seconds was wrong in both directions: it was paid in
+full on every healthy run, and on a machine slow enough to fire the hook at 5.1 s the
+control would have PASSED on its timer running out — the "test that cannot fail" shape this
+module has shipped several ways. Bounded by the fact, it keeps the full `_DEADLINE` for a
+machine that needs it and still answers in milliseconds. Settling ends the *wait*, never
+the reading: the marker is read once more on the way out, so an argv that landed in the
+same instant is still reported, and a pane tmux KEPT still fails the control by its own
+assertion rather than satisfying it.
+
+## What it tolerates: nothing
+
+No `sleep` was added, no retry, no assertion widened, no deadline raised, no return code
+stopped being checked, and no test's condition changed. The five remaining literal sleeps in
+the module are all waits for something *not* to happen, which cannot be made cheaper by
+looking more often because there is nothing to look at — and they are now enumerated by
+value in a test, so a flat poll cannot creep back into one of the twenty loops unnoticed.
+
+Four new cases, all four red on the version they fix:
+
+* the first look back is `_POLL_FIRST` and the gap settles at `_POLL_MAX`, read off
+  `time.sleep` rather than off a stopwatch;
+* a wait told nothing can change asks once and stops, instead of spending 20 s;
+* a wait that settles still reports what arrived in the same instant — the half that keeps
+  the negative control able to fail at all;
+* and the module contains no `time.sleep(<a number>)` outside the five that are listed.
+
+## The measurement
+
+Same 102 tests, all passing, three runs a side on one workstation (tmux 3.7c, darwin):
+
+| | before | after |
+|---|---|---|
+| module wall | 29.6 / 27.5 / 27.3 s | 17.9 / 19.2 / 19.6 s |
+| `time.sleep` | 16.68 s over 154 calls | 5.86 s over 191 calls |
+
+Nothing to adopt — no production behaviour changed, and no production file was touched.
--- a/tests/test_frame_tmux_integration.py
+++ b/tests/test_frame_tmux_integration.py
@@ -80,6 +80,7 @@ one does.
 
 from __future__ import annotations
 
+import ast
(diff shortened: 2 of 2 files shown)


diazoxide:
**Local full suite, run the way CI runs it** (`python3 -m unittest discover -s tests`):

```
Ran 10730 tests in 536.258s
FAILED (failures=2, errors=2, skipped=9)
```

All four are environmental and not this branch's. This clone lives at
`workspaces/<ws>/charter` *inside* the operator's own control plane — which is exactly the
arrangement `test_plane_write_guard` exists to refuse, and `charter`'s own branch guard
requires a workspace clone rather than the plane root, so there is nowhere else to run it
from here:

```
FAIL:  test_plane_write_guard.WhatIsGuarded.test_the_suite_reads_the_checkout_these_modules_were_loaded_from
FAIL:  test_plane_write_guard.WhatIsGuarded.test_the_plane_a_worktree_redirects_to_is_still_one_the_guard_refuses
ERROR: test_statusline_crash_guard.RenderNeverCrashesCase.test_a_broken_persona_chips_call_does_not_crash_render
ERROR: test_statusline_crash_guard.RenderNeverCrashesCase.test_a_broken_repo_row_does_not_crash_render
```

Same four, same counts, on `origin/main` in the same clone with this branch stashed:

```
=== ON MY BRANCH ===     Ran 23 tests in 0.088s   FAILED (failures=2, errors=2)
=== ON origin/main ===   Ran 23 tests in 0.091s   FAILED (failures=2, errors=2)
```

Neither module is touched by this branch. `test.yml` above is the reading that counts.

The module this PR is about, on this branch:

```
Ran 106 tests in 17.659s
OK
```

diazoxide:
**The measurement that matters, now taken on CI rather than projected.**

Both numbers read the same way out of `test.yml`'s own verbose log — the wall between the
module's first test-start line and the first line after its last — and both are the
**3.13** matrix job, so nothing but this branch differs:

| | module | tests | rest of #843's frame selection | module's share |
|---|---|---|---|---|
| `main` @ ac3d21c (run 33751547084) | **22.9 s** | 102 | 5.00 s | 82.1% |
| this branch (run 33759857191) | **13.3 s** | 106 | 4.81 s | 73.5% |

**42% off the module**, with four more tests in it, and #843's headline — "82% of a frame
area selection's cost" — is now 73.5%.

The test that was the module's most expensive on CI is gone from the top of the list
entirely. Slowest six now:

```
 1.417s  test_a_real_drag_across_the_table_width_removes_the_pane_and_brings_it_back
 1.169s  test_a_corrupt_version_file_does_not_kill_a_live_panel
 0.626s  test_a_state_bump_through_the_real_hook_repaints_the_table_and_the_alert_row
 0.519s  test_a_hostile_hotkey_from_charter_toml_runs_nothing_at_launch
 0.518s  test_the_frame_draws_every_rule_the_same_inside_the_operators_own_tmux
 0.427s  test_the_placeholder_does_not_exit_on_its_own
```

`test_a_panel_dying_here_is_destroyed_unless_charter_arms_the_window` was **5.048 s** on
`main` — the single most expensive test in the module and 22% of it. It is now below every
line above.

Whole suite on that job: `Ran 10730 tests in 361.478s`, and the shard fixed cost the sweep
pays for a baseline comes down with it.

Issue #21, opened 2026-08-09
vaults.json stores ABSOLUTE paths, so the vault registry cannot be committed or shared


diazoxide:
## What happens

`.charter/vaults.json` records each vault's file with an absolute path:

```json
"devops": {
  "provider": "reference",
  "persona": "devops",
  "config": { "file": "/Users/aharon/IdeaProjects/volaticloud/.charter/vaults/devops.json" }
}
```

## Why it matters

Reference vaults are **safe to commit** — they hold `op://` URIs, never values. We verified that and now commit ours: 59 references across three vaults, so a fresh clone or another agent inherits the credential wiring instead of rediscovering it.

But the **registry** cannot go with them, because it hard-codes one developer's home directory. So the committed vault files are invisible to charter on any other machine until someone re-registers them by hand. We had to write a `charter-register-vaults.sh` that loops `charter vault add … --provider reference` purely to reconstruct state that is otherwise already in git.

## Suggestion

Store the path relative to the control-plane root when it lives inside it, and resolve at load time. Absolute stays valid for vaults deliberately kept outside the repo.

That single change makes `.charter/vaults.json` committable, which makes the whole reference-vault feature portable — currently its main limitation is not the vaults but the index that finds them.

## Aside

`charter vault list` also prints those absolute paths in its STATUS column (`not created yet (/Users/…)`), which is noisy and leaks local layout into terminal output others may see.

PR #914, opened 2026-09-05
A plane declares a format version a reader can refuse, and charter stops rather than guesses (#913)

### LINKED ISSUE #913: A plane declares a format version a reader can refuse, the way a workspace already does
The first concrete step toward charter being interface-agnostic, and it is smaller than it sounds because **charter already has the idiom one level down**.

## What exists

`workspace.STRUCTURE_VERSION = 5` is stamped into every workspace's `.charter-structure`, and `charter workspace reinit` is the upgrade path. A workspace whose layout predates the current version says so and can be repaired.

**The plane itself has no such stamp.** Neither does the frame state under `.charter/`.

## Why that matters now

Research into how comparable projects expose a stable seam (`docs/research/2026-09-05-interface-agnostic-cores.md`) found that git — charter's closest sibling — says exactly one thing to a program reading its files directly, and it is a rule for **refusing**:

> an implementation that does not understand `core.repositoryformatversion` **"MUST NOT operate on that repository"**

That is the whole contract. Not a schema, not a spec — a way to know you are out of your depth and stop. `packed-refs` and the loose-ref format are specified only in C, and gitoxide shipped subtly wrong refs for years because git silently `rtrim`s on read. A consumer with no way to refuse **guesses**, and that is the failure mode.

charter is about to grow consumers that are not the frame — that is the direction of travel. Today any of them would guess.

## The work

1. **A plane format version**, stamped where `charter init` scaffolds and read wherever charter opens a plane. The same shape as `STRUCTURE_VERSION`: a single integer, bumped when the on-disk layout changes in a way an older reader cannot survive.
2. **A refusal, not a warning.** A reader that does not understand the version must decline to operate rather than proceed on a guess. `doctor` reports it; the CLI refuses it. This is the whole point — a version nobody refuses on is decoration.
3. **The frame state too**, or an explicit statement that `.charter/frame/` is private and carries no promise. Either is defensible; leaving it ambiguous is not.

## What this is not

**Not a schema and not a spec.** Do not document the shape of `gather.json` or `workspace.json` as a contract — git's lesson is that the format is truth, and the *commands* are interface. This issue buys exactly one thing: a consumer's ability to know it must stop.

**Not a daemon**, and not a step toward one. charter's writers are hooks inside the agent's turn, where the standing rule is that a hook may cost a session its briefing, never its turn.

## Care

- `STRUCTURE_VERSION` moved 4 → 5 this week (#889 made `workspace.json` mandatory), so the upgrade path is warm and worth reading before adding a second version number beside it.
- Two version numbers that can disagree is worse than one. Say plainly how the plane version and `STRUCTURE_VERSION` relate — nested, independent, or one subsuming the other — and pin it.
- A plane created before this exists has no stamp. Absent must mean something definite, and "treat absent as current" is the choice that makes the feature useless the first time it is needed.

diazoxide:
Closes #913.

charter is about to grow consumers that are not the frame, and today any of them would
guess. This gives a program that opens a plane a way to know it is out of its depth and
**stop**.

## The number already existed. The refusal did not

`charter init` has stamped `schema = 1` into `charter.toml` since the beginning
(`commands._render_charter_toml`), and `instance.load` has raised on a version from the
future for just as long. **So this does not add a second plane-level version number** — the
issue's own "two version numbers that can disagree is worse than one" is answered by not
minting one. What was missing was the two properties that make the number worth having.

**1. `config.derive` caught the exception and carried on.** It recorded the message as
`CONFIG_ERROR`, set `cfg = {}` and handed every command a plane made of fallback defaults.
A plane declaring `schema = 99` did not stop charter; it made charter quietly act on a plane
it had already been told it could not read. The sharpest case is the guard:
`known_forges_report` opens `charter.toml` through the same `load`, got the refusal and
returned an empty forge set — a PreToolUse guard that looks present and denies nothing.

`config.PLANE_REFUSAL` is now recorded apart from `CONFIG_ERROR` (which stays set, because
`doctor`'s `charter.toml` row and `channel.channel`'s documented fallback key off it), and
`cli.main` declines the command. Four commands still run, each a question about *charter*
rather than a read of the plane's contents: `doctor` (where it is reported), `version` and
`_version-check`, and `update` — the only remedy, since nothing but a newer charter can
understand a newer plane. `--version` and `--help` need no exemption; argparse's own actions
exit before the gate. `init` and `reinit` are deliberately **not** exempt: writing into a
layout you have been told you do not understand is the guess at its most damaging. Neither
is `hook`, which turns that silent guard fail-open into a visible one, at the cost of the
session's briefing and never its turn (rc 1, never `hooks.DENY_EXIT`).

**2. An absent stamp meant "whatever this charter is".** `cfg.get("schema", SCHEMA)` — the
reading that makes the feature useless the first time it is needed. The day `SCHEMA` moves
to 2, every unstamped version-1 plane would claim to be a version-2 plane, and a charter
that understands 2 would read a version-1 layout with version-2 rules.

Absent now means `instance.UNSTAMPED = 1`: the layout every plane had before planes declared
one. Definite, and *true* — there has only ever been one plane format, so every plane that
exists today is a version-1 plane and stays readable forever. Refusing them was never on the
table; the choice was only between "1" and "current", and "current" is the one that lies
later. The two readings are indistinguishable today (`SCHEMA` is 1), which is why the test
that pins it moves `SCHEMA` before it looks.

A `schema` charter cannot compare against at all (`"2"`, `1.5`, `true`) used to fall through
`isinstance(found, int)` and read as understood. Refused on the same terms. `true` is the
sharp one: `isinstance(True, int)` is `True`, so it would otherwise compare as 1.

## How the plane version relates to `STRUCTURE_VERSION`

**Nested, and only the outer one refuses.** Stated at `instance.SCHEMA` and pinned in both
directions.

| | `instance.SCHEMA` | `workspace.STRUCTURE_VERSION` |
|---|---|---|
| asks | may this charter operate on this plane at all? | is this one workspace's interior current? |
| answers | yes / no | current / stale |
| a bad answer is | refused — nothing heals it but a newer charter | repaired, additively, by `charter workspace reinit` |

A workspace-layout change an older charter *survives* is, by definition, not a change that
requires refusal, so it moves `STRUCTURE_VERSION` alone — which is exactly what 4 → 5 was.
Two numbers can only come to disagree if something compares them; nothing does, and
`ThePlaneVersionAndTheStructureVersionAreNeverCompared` is what keeps it that way: moving
`SCHEMA` must not change a workspace's staleness verdict, and moving `STRUCTURE_VERSION`
must not change the plane's.

## `.charter/frame/` is private, and says so in code

`frame.state.NO_FORMAT_PROMISE`. Not "because it is internal" — everything in charter is
internal until somebody reads it — but because **a format version buys the ability to
refuse, and refusal is worth something only when the writer and the reader can be different
charters.** A plane's files are committed, shared and outlive any install. Nothing under
`.charter/frame/` is: it is written by a frame launcher and read by that same frame's
panels — one process tree, one machine, one charter — and residue from any other charter is
`reap`ed rather than read, because liveness there is a pid or a tmux window. A version
stamped there could never fire, and the issue's own rule is that a version nobody refuses on
is decoration. Two consequences, and they are the whole promise: charter may change the
shape of anything under `.charter/frame/` in a patch release with no bump and no migration,
and a program reading it has no contract to hold charter to. A tokenised tripwire (not a
grep — the constant itself discusses `instance.SCHEMA` at length) fails if any module under
`charter/frame/` gains a format version in *code*.

## Two findings out of writing the tests

- **`args.command` is not the subcommand.** The `exec` subcommand under `charter secret`
  gives its own positional the dest `command` (`_sa_exec`), so argparse overwrites the
  subparser name with the child command line. A gate reading it got `TypeError: unhashable
  type: 'list'` out of `not in` — a crash report filed against charter, on the one plane
  charter had decided not to touch. The gate reads `argv[0]`.
- **Bare `charter` fell through to argparse's usage message**, which is byte-identical to
  what a plane declaring no harness gets, so a refused plane looked like an unconfigured
  one. `_bare_launch` refuses it through the same `_plane_refusal` helper the gate uses, so
  the bare form and the typed one cannot come to differ.

## Not a schema and not a spec

The shape of `gather.json`, of `workspace.json`, of anything else charter writes is
documented to nobody and stays free to change. git's lesson is that the format is truth and
the *commands* are interface. This buys exactly one thing: a consumer's ability to know it
must stop.

## Verification

`python3 -m unittest discover -s tests -q` — **`Ran 11610 tests in 629.286s` / `OK
(skipped=11)`**, trailer read rather than the exit code. 39 of them are new.

https://claude.ai/code/session_01Us4MbwkSCJCxwt1CjAd1sf

### DIFF
--- a/charter/cli.py
+++ b/charter/cli.py
@@ -1908,6 +1908,20 @@ def _bare_launch(argv: list[str]) -> tuple[list[str], int | None]:
         return argv, None
     from . import config
 
+    # *The plane's format version is one charter cannot place.* Reported here for the same
+    # reason the refused-default arm below is: a plane charter has refused declares nothing
+    # charter can read, `[harness] default` included — `config.HARNESS` derives from the
+    # empty config every time — so falling through hands the operator argparse's usage
+    # message, byte-identical to what a plane that declared no default gets. Bare `charter`
+    # would then look like a plane with no harness rather than one charter refused. The
+    # gate in `main` cannot cover this: it reads the typed subcommand, and a bare argv has
+    # none. Both call the same :func:`_plane_refusal`, so there is one message and one
+    # exemption rule between them.
+    refusal = _plane_refusal(None)
+    if refusal:
+        util.err(refusal)
+        return argv, 1
+
     declared = config.HARNESS
     # Truthiness, not `is not None`, and the same test `doctor` makes of the same key.
     # `contain.readable` never returns a blank string, so on anything `harness_of` produced
@@ -1963,6 +1977,25 @@ def _bare_launch(argv: list[str]) -> tuple[list[str], int | None]:
 _DESPITE_REFUSAL = frozenset({"doctor", "update", "version", "_version-check"})
 
 
+def _plane_refusal(typed: str | None) -> str | None:
+    """What to print instead of running *typed*, or ``None`` to carry on.
+
+    One function for the two callers — `_bare_launch` (which has no subcommand to pass) and
+    the gate in :func:`main` (which passes the typed token) — so the message and the
+    exemption rule cannot come to differ between the bare form and the typed one.
+
+    ``None`` is never exempt, and that is deliberate rather than incidental: it is what a
+    bare ``charter`` passes, and opening a frame over a plane charter cannot place is the
+    most expensive guess available to it.
+    """
+    from . import config
+    refusal = config.PLANE_REFUSAL
+    if not refusal or typed in _DESPITE_REFUSAL:
+        return None
+    return (f"{refusal} Nothing was run. `charter doctor` reports it; "
+            f"`charter update` is the way out.")
+
+
 def main(argv=None) -> int:
     argv = list(sys.argv[1:] if argv is None else argv)
     # First, so what it returns is then hoisted, split and parsed exactly as the same
@@ -1981,16 +2014,24 @@ def main(argv=None) -> int:
         # gap signal needs catching here rather than a third `except` down there.
         _hint_gap_if_unknown_command(parser, argv)
         raise
-    # The refusal, and it is read HERE for two reasons. It is after `parse_args`, so charter
-    # has understood the command before declining it and `--version`/`--help` have already
-    # exited from inside argparse. And it is before the two rewrites below: `secret exec`
-    # overwrites `args.command` with the command line to run, so the subcommand's own name
-    # is only readable on this side of that line.
-    from . import config as _config
-    refusal = _config.PLANE_REFUSAL
-    if refusal and args.command not in _DESPITE_REFUSAL:
-        util.err(f"{refusal} Nothing was run. `charter doctor` reports it; "
-                 f"`charter update` is the way out.")
+    # The refusal, and both halves of where it sits are load-bearing.
+    #
+    # *After `parse_args`*, so charter has understood the command before declining it and,
+    # more to the point, so `--version` and `--help` are already gone: both are argparse
+    # actions that exit from inside `parse_args`, which is what keeps them working on a
+    # plane charter will not otherwise touch. A gate above this line would have to
+    # re-implement that, and would refuse the two questions an operator asks first.
+    #
+    # *Reading `argv[0]` and not `args.command`.* They agree for almost every command and
+    # then do not: `charter secret exec` gives its own positional the dest `command`
+    # (`_sa_exec`), so argparse overwrites the subparser name with the child command line
+    # and `args.command` is a LIST there — unhashable, and a `not in` against a frozenset
+    # raises rather than refuses. `argv[0]` is the token the operator typed, `_bare_launch`
+    # has already rewritten a bare `charter` into it, and the root parser carries no flag
+    # that survives to this line, so it is the subcommand every time.
+    refusal = _plane_refusal(argv[0] if argv else None)
+    if refusal:
+        util.err(refusal)
         return 1
     if exec_command is not None:
         args.command = exec_command
--- a/charter/cli.py
+++ b/charter/cli.py
@@ -2029,7 +2029,12 @@ def main(argv=None) -> int:
     # raises rather than refuses. `argv[0]` is the token the operator typed, `_bare_launch`
     # has already rewritten a bare `charter` into it, and the root parser carries no flag
     # that survives to this line, so it is the subcommand every time.
-    refusal = _plane_refusal(argv[0] if argv else None)
+    #
+    # Subscripted with no emptiness guard because `parse_args` has already made one: the
+    # subparsers are `required=True`, so an empty argv exits from inside it and never
+    # arrives here. A guard would be an unreachable branch, which is the shape this
+    # repository deletes.
+    refusal = _plane_refusal(argv[0])
     if refusal:
         util.err(refusal)
         return 1
--- a/charter/cli.py
+++ b/charter/cli.py
@@ -1938,6 +1938,31 @@ def _bare_launch(argv: list[str]) -> tuple[list[str], int | None]:
     return [default, *argv], None
 
 
+#: The commands that still run on a plane charter has refused (`config.PLANE_REFUSAL`).
+#:
+#: Three of them, and each earns its place by being about CHARTER rather than about the
+#: plane's contents: `doctor` is where the refusal is reported, `version`/`_version-check`
+#: answer "which charter is this" — the first question anyone asks on reading the refusal —
+#: and `update` is the only remedy, since nothing but a newer charter can understand a
+#: newer plane. Refusing the remedy would leave a plane with no way out but a hand edit.
+#:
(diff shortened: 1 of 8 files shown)


diazoxide:
## The sweep did not measure this branch, so I measured it by hand

**Verdict on the checks list:** `no verdict: 1 not measured; 5 of 22 measured, 17 out of time`.

`Sweep shard 1 of 1` started 20:11:51Z and was cancelled at 21:12:08Z — exactly its `timeout-minutes: 60`. `plan` had sized 22 mutations into one shard (22 < 28 per shard), and the eight `charter/config.py` mutations ate the hour: `config` is imported by every module, so its selection map is close to the whole suite and one mutation there is one near-full-suite run. The shard never reached `charter/instance.py`, where this branch's refusal actually lives.

That is a gate that ran out of clock, not a finding. Re-dispatching cannot fix it — `workflow_dispatch` takes no inputs and `plan` re-derives the same single shard. Filed upstream as **#915**.

## All 22, hand-applied by `file:line`

Each mutation applied into its exact line span, `__pycache__` cleared on **both** transitions, subset run, reverted, `__pycache__` cleared again. RED proves a kill; nothing here needed escalation because nothing came back green.

| # | site | operator | result |
|---|---|---|---|
| 0 | `charter/cli.py:1993` | drop-if (`_plane_refusal`'s whole guard) | RED — 16 failures, 5 errors |
| 1 | `charter/config.py:688` | retune-string `"CONFIG_ERROR"` | RED — 5 failures |
| 2 | `charter/config.py:688` | retune-string `"PLANE_REFUSAL"` | RED — 9 failures |
| 3 | `charter/config.py:689` | narrow-except `PlaneFormatUnknown`→`ZeroDivisionError` | RED — 13 failures, 2 errors |
| 4 | `charter/config.py:691` | retune-string `"CONFIG_ERROR"` | RED — 2 failures |
| 5 | `charter/config.py:691` | retune-string `"PLANE_REFUSAL"` | RED — 13 failures, 2 errors |
| 6 | `charter/config.py:692` | narrow-except `Exception`→`ZeroDivisionError` | RED — 3 errors |
| 7 | `charter/config.py:693` | retune-string `"CONFIG_ERROR"` | RED — 2 failures |
| 8 | `charter/config.py:693` | retune-string `"PLANE_REFUSAL"` | RED — 1 failure |
| 9 | `charter/doctor.py:277` | drop-if (the `CONFIG_ERROR` row) | RED — 1 failure |
| 10 | `charter/doctor.py:282` | collapse-ifexp → the fallback arm | RED — 1 failure |
| 11 | `charter/doctor.py:282` | collapse-ifexp → the refusal arm | RED — 1 failure |
| 12 | `charter/doctor.py:400` | drop-if (the `schema` row's refusal) | RED — 2 failures |
| 13 | `charter/frame/state.py:181` | retune-string `NO_FORMAT_PROMISE` | RED — 2 failures |
| 14 | `charter/instance.py:63` | retune-constant `UNSTAMPED` 1→2 | RED — 3 failures |
| 15 | `charter/instance.py:97` | drop-if (absent→`UNSTAMPED`) | RED — 4 errors |
| 16 | `charter/instance.py:97` | retune-string `"schema"` | RED — 16 failures, 1 error |
| 17 | `charter/instance.py:99` | retune-string `"schema"` | RED — 17 failures, 13 errors |
| 18 | `charter/instance.py:100` | drop-if (the `bool`/non-`int` refusal) | RED — 3 failures, 1 error |
| 19 | `charter/instance.py:125` | drop-if (`found is None` → raise) | RED — 3 failures, 1 error |
| 20 | `charter/instance.py:132` | drop-if (`found > SCHEMA` → raise) | RED — 12 failures, 1 error |
| 21 | `charter/instance.py:132` | **shift-boundary `>` → `>=`** | RED — 23 failures, 9 errors |

**22 of 22 killed. 0 survivors — nothing to pin, nothing to delete, nothing unkillable.**

Two are worth naming because they are the ones the branch was written to survive:

- **#21, `charter/instance.py:132`, `>` → `>=`.** The shift-boundary attack on the version comparison. Killed by `test_exactly_the_version_this_charter_understands_is_accepted`, which sits on `SCHEMA` itself rather than five versions away from it.
- **#14, `charter/instance.py:63`, `UNSTAMPED` 1 → 2.** The absent-stamp constant. Killed by `test_the_constant_says_one_and_the_reader_uses_it`.

Separately hand-checked earlier and also RED: `cli.py`'s `_plane_refusal(argv[0])` re-spelled `argv[-1]` — 2 failures, from the two cases added for exactly that (`charter update --to 0.59.0` and `charter status --workspace doctor`).

## Suite

`python3 -m unittest discover -s tests` on the final tree — **`Ran 11612 tests` / `OK (skipped=9)`**, trailer read rather than the exit code. Green on 3.11, 3.12, 3.13 and 3.14 in CI, on both triggers.

PR #916, opened 2026-09-07
The deletion sweep sizes a shard by what its mutations cost, not by how many there are (#915)

### LINKED ISSUE #915: The deletion sweep sizes shards by mutation count, so a diff touching charter/config.py runs out of the hour with 17 of 22 unmeasured
`tools/sweep.py --plan` sizes shards by how many mutations there are, never by what they cost — so a 22-mutation diff that touches `charter/config.py` gets one shard and runs out of the runner's hour with 17 of 22 never measured.

## Measured, on #914

`charter/config.py` is the file every other module imports. Its selection map is close to the whole suite, so one mutation there is one near-full-suite run.

```
$ python3 tools/sweep.py --plan
  22 mutations → 1 shard(s) of at most 28 each
    charged: charter/cli.py, charter/config.py, charter/doctor.py,
             charter/frame/state.py, charter/instance.py
```

One shard, because 22 < 28. What the runner then did:

```
Sweep shard 1 of 1   started 20:11:51Z   completed 21:12:08Z   cancelled
```

Exactly 60 minutes — `timeout-minutes: 60`. The verdict:

```
gate: 0 unpinned, 0 in masked cluster(s), 0 platform-deferred, 1 unresolved,
      17 out of time, 0 not applied, 0 withheld, 4 pinned
deletion sweep / no verdict: 1 not measured; 5 of 22 measured, 17 out of time
```

**5 of 22.** The eight `charter/config.py` mutations are the expensive ones; the shard spent its hour on them and never reached `charter/instance.py`, where the branch's actual refusal logic lives.

Reproduced locally (14 cores, `--jobs 8`, `--budget 0`): 5 pinned, 8 `unresolved`, 9 `out_of_time` — and every one of the 8 unresolved carries `{'green': False, 'ran': 0, 'detail': 'timed out after 900s'}`. Not a finding. A clock.

## Why re-running does not help

`workflow_dispatch` takes no inputs, and `plan` re-derives the same 1 shard from the same 22 mutations. The sweep cannot be re-run into a verdict on this branch; it will spend another hour reaching the same five.

## The shape of it

`plan` divides mutations by `MUTATIONS_PER_SHARD`. That is a count, and cost is not a function of count — it is a function of the **selection map**, which `--warm-map` already measures and which `plan` already has in hand. A mutation in `charter/config.py` and one in `charter/frame/state.py` differ by two orders of magnitude in test-seconds and are dealt as equals.

The consequence is worst exactly where the gate matters most: a small, careful diff that touches a widely-imported module looks cheap to `plan` and is not, so the branches that most need a verdict are the ones that cannot get one.

Not proposing the fix here — sizing by measured subset size, capping mutations-per-shard by predicted seconds rather than by count, and giving `workflow_dispatch` a shard override are all plausible and they trade against each other. What is not defensible is that the number is a count today while the docstring on `--shard` already talks in terms of the plan being *dealt* — the dealing is fair, the hands are not.

## What #914 did instead

Hand-applied all 22 by `file:line`, clearing `__pycache__` on both transitions, RED confirmed on each, reverted. 22 of 22 killed, 0 survivors. That is a stronger result than the gate would have produced and it does not scale past a diff this size, which is the point.

diazoxide:
`tools/sweep.py --plan` divided the mutation count by a per-shard count. Cost is not a
function of count — it is a function of the selection map, which `--warm-map` measures in
that very job and which `plan` then threw away.

## The measurement, on #914's charged set

Reconstructed with `--ref 40ace71 --base 9d18d55`, which reproduces the issue's plan
exactly — same five files, same 22 mutations.

**Before** (`origin/main`):

```
  22 mutations → 1 shard(s) of at most 28 each
    charged: charter/cli.py
    charged: charter/config.py
    charged: charter/doctor.py
    charged: charter/frame/state.py
    charged: charter/instance.py
```

**After** (this branch, same tree, same warmed map):

```
  22 mutations → 8 shard(s); 174 min of predicted test time over 450 test modules,
                 1707 s on the heaviest shard (a shard is sized for 1680 s)
    charged: charter/cli.py — 1 min predicted
    charged: charter/config.py — 95 min predicted
    charged: charter/doctor.py — 5 min predicted
    charged: charter/frame/state.py — 4 min predicted
    charged: charter/instance.py — 69 min predicted
::warning title=The deletion sweep is over its budget::22 mutations is 174 minutes of
predicted test time, and the heaviest of 8 shards would carry 1707 s against the 1680 s a
shard is sized for. Nothing here is dropped …
```

The map, traced on this repository, is why:

| mutated symbol | test modules selected, of 450 |
| --- | --- |
| `charter/instance.py::load` | 337 |
| `charter/config.py::derive` | 323 |
| `charter/frame/state.py` (module scope) | 105 |
| `charter/doctor.py::check_control_plane_config` | 25 |
| `charter/cli.py::_plane_refusal` | 6 |

Fifty-four times apart in the tests they select, dealt as equals, and 22 < 28.

## How a shard is sized now

```
seconds_for(m) = SECONDS_A_MUTATION_FIXED + SECONDS_A_SUITE_RUN × (modules selected / 450)
```

`shards_for(costs)` then asks for **the fewest shards whose heaviest hand fits**
`SHARD_BUDGET - SHARD_FIXED`, and it asks through `shard_of` — the same round-robin the
shards deal with — rather than through a second copy of the dealing rule. Dividing the
total would call a plan a fit when one shard is carrying both expensive mutations.

Both constants are **solved from the two runs that ran out of time**, whose maps are known
and very different:

```
#626:  53×F + 53×(13/450)×S   = 2880 s        (charter/gitconfig.py, 13 of 450)
#914:   5×F + (6+4×323)/450×S = 3000 s        (cli.py ×1, config.py ×4)
        →  F ≈ 25 s,  S ≈ 996 s
```

960 rather than 996 because it is corroborated independently: the unmutated baseline is
240 s alone on the runner, a shard runs `--jobs 4`, and `unittest` is single-threaded — four
concurrent whole-suite runs queue on a four-vCPU runner rather than overlapping. And
`SUBSET_TIMEOUT` is 900, which is the same number from the other side: a subset that is the
whole suite is the run that reaches it, and on #914 every `charter/config.py` mutation did.

The model reproduces #626 at 2794 s against 2880 measured, and #914 at 2894 s against
~3000 — i.e. it predicts, from the map alone, that one shard reaches about five of those 22.
Both are asserted.

**The model does not predict the full-suite run `decide` makes when a subset does not kill,
nor a survivor's confirmation.** Those depend on the answer and not on the diff. They are
what the twenty minutes between `SHARD_BUDGET` (40) and `SHARD_REPORT_AT` (50) are for, and
that gap is untouched — asserted, so a later edit that spends it is red here.

## When cost wants more than eight shards

**#914 lands on this**: eight shards are sized for 1680 s each and its heaviest hand is
1707. Something has to give. What gives is the **budget**, never the plan.

Every mutation is still dealt. The shards carry more than they were sized for, stop at
`SHARD_REPORT_AT`, and report what they measured, so `no verdict: N out of time` stays
reachable and honest — which it must, because the verdict is the check's name. Capping the
plan to make it fit would make `no survivors` a claim about a sweep that never ran; that is
strictly worse than the bug. Raising `MAX_SHARDS` was not taken either: it is a limit on how
much of the runner pool one branch may hold, it is not a correctness knob, and no finite
value of it closes the case anyway.

What changes is *when you find out*. The warning fires on the plan job, before a runner is
spent on measuring, and names predicted minutes rather than a mutation count — the count
being exactly what could not see #914 coming, 22 being comfortably inside the old ceiling
of 224.

## The `workflow_dispatch` shard override, deliberately not taken

The issue names it as a third option. It is a safety valve for when prediction is wrong, not
a fix: a manual override does not help the branch that silently got no verdict, because
nobody knew to override. It would also install a second sizing authority, free to disagree
with the measured one — #670's shape, on the number that decides whether the answer arrives.
The annotation is the automatic version of the same warning, and it arrives before the hour
is spent rather than after.

## What stays exactly as it was

- **The count path is bit-identical.** With no map every mutation is priced at
  `SECONDS_A_MUTATION`, eight hands of 28 × 60 s are exactly the 1680 s a shard is sized
  for, and the 224/225 warning boundary this suite already pinned survives the change of
  unit untouched. `--plan` on its own still costs one `ast` pass; `sweep.yml` always warms.
- **The dealing.** `shard_of` still depends on nothing but the diff. A deal that consulted
  the map would be a slice each shard derived from its own measurement, and two shards
  disagreeing by one module would sweep one slice twice and another never while the merge
  step reported a complete sweep.
- **`SHARD_BUDGET`, `SHARD_REPORT_AT`, `timeout-minutes: 60`, `MAX_SHARDS`.** No constant
  that was a deadline moved.

One behaviour does change as a consequence of the ordering: `plan` now warms the map before
it sizes, so a trace that refuses (`the selection map is empty — refusing to sweep blind`)
leaves no `shards` output and `collect` says `no verdict: the sweep never sized itself`
rather than `no verdict: N of N shards did not report`. The shards never ran in that case
either way, so the new sentence is the true one.

## Verification

`python3 -m unittest discover -s tests -q` — read the trailer, and it is `OK`.
`tests/test_sweep.py` alone is 461 tests. The before/after above is a real `--plan` run
against a real traced map (450 modules), not a fixture.

Closes #915.

https://claude.ai/code/session_018hUBNraNfixfMMK5Jc6CNy

### DIFF
--- a/docs/news/unreleased-the-deletion-sweep-sizes-a-shard-by-what-it-costs.md
+++ b/docs/news/unreleased-the-deletion-sweep-sizes-a-shard-by-what-it-costs.md
@@ -1,0 +1,136 @@
+---
+version: unreleased
+headline: The deletion sweep sizes a shard by what its mutations cost, not by how many there are
+---
+
+`tools/sweep.py --plan` decides how many machines a branch's deletion sweep needs. It did
+that by dividing the mutation count by a per-shard count, and cost is not a function of
+count.
+
+On #914 that produced a check named `no verdict: 5 of 22 measured, 17 out of time`, after
+one shard burned the runner's full hour:
+
+```
+$ python3 tools/sweep.py --plan
+  22 mutations → 1 shard(s) of at most 28 each
+    charged: charter/cli.py, charter/config.py, charter/doctor.py,
+             charter/frame/state.py, charter/instance.py
+
+Sweep shard 1 of 1   started 20:11:51Z   completed 21:12:08Z   cancelled
+```
+
+One shard, because 22 is less than 28. Eight of those 22 mutations were in
+`charter/config.py` — the module every other module imports — and the shard spent its hour
+on them. It never reached `charter/instance.py`, where that branch's refusal logic lives.
+
+## The number the plan already had and threw away
+
+The sweep does not run the whole suite for every mutation. It runs the test modules the
+**selection map** measured executing the mutated function, which is what turns a four-minute
+suite into a two-second one. That map is traced by the plan job itself — `--warm-map` is
+there so the eight shards downstream can restore it instead of measuring it eight times.
+
+So the plan job held, in the same process, the one thing that says what a mutation costs.
+Measured on this repository:
+
+| what the mutation touches | test modules it selects, of 450 |
+| --- | --- |
+| `charter/config.py::derive` | 323 |
+| `charter/instance.py::load` | 337 |
+| `charter/frame/state.py` | 105 |
+| `charter/doctor.py::check_control_plane_config` | 25 |
+| `charter/cli.py::_plane_refusal` | 6 |
+
+Two mutations in one diff, fifty-four times apart in the tests they select — and nineteen
+times apart once the fixed cost every mutation pays is added back in — dealt as equals. The
+plan job measured that, sized the fan-out off a count, and deleted the measurement.
+
+## Two terms, because the cost has two halves
+
+`seconds_for` prices one mutation as a fixed part plus a measured part:
+
+```
+SECONDS_A_MUTATION_FIXED  +  SECONDS_A_SUITE_RUN × (modules selected / modules in the suite)
+```
+
+The fixed part is the sandbox restore, the interpreter, and `unittest discover`'s own walk
+of `tests/` — the count-shaped half, and the only half a count could ever see. The second
+is the run itself.
+
+Both constants are **solved from the two runs that ran out of time**, whose selection maps
+are known and very different. #626 measured about 53 mutations of `charter/gitconfig.py`
+(13 of 450 modules) in ~2880 s; #914 measured 5 — one of `charter/cli.py` at 6 of 450, four
+of `charter/config.py` at 323 — in ~3000 s. Two runs, two unknowns, and they give 25 s and
+about 996 s. The round number is 960 because it is corroborated twice over: the unmutated
+baseline is 240 s alone on the runner, a shard runs `--jobs 4`, and `unittest` is
+single-threaded — so four concurrent whole-suite runs queue on a four-vCPU runner rather
+than overlapping. `SUBSET_TIMEOUT` is the same number from the other side, and on #914 every
+`charter/config.py` mutation reached it: `{'green': False, 'ran': 0, 'detail': 'timed out
+after 900s'}`.
+
+What the model deliberately does **not** predict is the second and third runs a verdict can
+need — the full suite that has the last word, and a survivor's confirmation. Those depend on
+the answer and not on the diff, so nothing can see them coming. They are what the twenty
+minutes between `SHARD_BUDGET` and `SHARD_REPORT_AT` are for, and that gap is not spent
+here.
+
+## #914's own plan, before and after
+
+```
+before:  22 mutations → 1 shard(s) of at most 28 each
+
+after:   22 mutations → 8 shard(s); 174 min of predicted test time over 450 test
+         modules, 1707 s on the heaviest shard (a shard is sized for 1680 s)
+           charged: charter/cli.py — 1 min predicted
+           charged: charter/config.py — 95 min predicted
+           charged: charter/doctor.py — 5 min predicted
+           charged: charter/frame/state.py — 4 min predicted
+           charged: charter/instance.py — 69 min predicted
+         ::warning::22 mutations is 174 minutes of predicted test time, and the
+         heaviest of 8 shards would carry 1707 s against the 1680 s a shard is
+         sized for…
+```
+
+The `charged:` list now carries what each file is predicted to cost, because "which file is
+making this sweep slow" is the question #914 could not answer from anything the plan
+printed — it took a person reading the map by hand.
+
+## Sized against the heaviest hand, and not against the total
+
+The plan is dealt round-robin, so a plan totalling two shards' worth of seconds can still
+put both of its expensive mutations on one shard. Dividing the total would call that a fit.
+`shards_for` instead asks for the fewest shards whose **heaviest hand** fits, and it asks
+through `shard_of` — the same function the shards deal with — rather than through a second
+copy of the dealing rule sitting beside it.
+
+The dealing itself is unchanged, and deliberately: it depends on nothing but the diff. A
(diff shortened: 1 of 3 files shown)


diazoxide:
## The gate's own check on this branch says `nothing to sweep`, and that is not a verdict on the change

`DEFAULT_PATHS = ("charter",)` and `sweep.yml` passes no `--path`, so the deletion sweep does not sweep `tools/`. This branch touches `tools/sweep.py`, `tests/test_sweep.py` and `docs/news/` and nothing under `charter/`, so the plan job measured **0 mutations**:

```
Size the sweep › How many mutations, and how many jobs they need
  selection map: 18067 source files in 413s
  0 mutations → 1 shard(s); 0 min of predicted test time over 450 test modules,
                0 s on the heaviest shard (a shard is sized for 1680 s)
```

Green, honest, and worth nothing as evidence about the change. So the sweep was run by hand against the diff it is actually about.

## `--path tools`, at `1f4edbb` against `40ace710e240`

```
$ python3 tools/sweep.py --gate --path tools --base origin/main --jobs 6 --budget 0

baseline          : Ran 11634 tests — OK
mutations applied : 20
pinned            : 20
SURVIVED          : 0
UNRESOLVED        : 0
wall clock        : 16.3 min

gate: 0 unpinned, 0 in masked cluster(s), 0 platform-deferred, 0 unresolved,
      0 out of time, 0 not applied, 0 withheld, 20 pinned
gate: no survivors
```

**20 of 20, by `file:line`.** The ones the sizing arithmetic exists for:

| | | |
| --- | --- | --- |
| `sweep.py:3307` | `shift-boundary` | `heaviest_hand(costs, count) <= per_shard_seconds()` — the comparison that decides the fan-out |
| `sweep.py:3339` | `shift-boundary` | `heaviest <= per_shard_seconds()` — the same boundary in `over_budget` |
| `sweep.py:3275` | `swap-synonym` | `max` → `min` in `heaviest_hand`: the *lightest* hand instead of the heaviest |
| `sweep.py:3230` | `collapse-ifexp` ×2 | `len(modules) / suite if modules else 1.0` — both branches, i.e. pricing a file no module covers at nothing instead of at a whole suite |
| `sweep.py:3227` | `drop-if` | `if not suite: return float(SECONDS_A_MUTATION)` — the no-map fallback |
| `sweep.py:3157` / `:3136` | `retune-constant` | `SECONDS_A_SUITE_RUN = 960` and `SECONDS_A_MUTATION_FIXED = 25` — would any number do? No. |
| `sweep.py:3176` ×2, `:3187` ×2 | `unclamp` | both `max(1, …)` clamps, both directions |
| `sweep.py:4320` ×2 | `disable-branch` | `if selection:` — sizing by cost when there is a map, and by count when there is not |
| `sweep.py:4378` | `drop-if` | `if not args.warm_map: return {}` — `--plan` alone must not trace |
| `sweep.py:4345`, `:4347` ×2 | `no-fallback`, `collapse-ifexp` | the per-file predicted cost on the `charged:` lines |

No survivor, no `unresolved`, no `out of time` — so no finding to report against the fix itself.

https://claude.ai/code/session_018hUBNraNfixfMMK5Jc6CNy

Issue #921, opened 2026-09-07
The frame advertises how to make a chat and not how to close one


diazoxide:
*"now for example no way to reactivelly close harness session, we have plus button - that adding new session tab, but no any option to delete/stop, no any modal/drawer for confirmation."*

Closing a chat has been possible the whole time — `F2 → chat: close`, or right-click a tab. The operator could not find either and concluded, from the evidence on screen, that it was not possible. **That conclusion was correct given what the frame says about itself.**

## What the frame advertises

Exactly two things: **`F2 palette`** on the attention strip, and **`+`** on the chats strip.

Not advertised anywhere: `F3`, `F12`, the per-component toggle keys, the right-click tab menu, the border drag that sets strip height, clicking a `+N`, the workspace chip, the persona head, a persona name, a repo row, the wheel, or that a tab is clickable at all.

charter has already made this exact argument about this exact strip, in `ADD_CHAT`'s own comment, about an earlier report from the same operator:

> *"`+` button not working for creating new session."*
> **This was a sentence and the sentence was the defect.** It read `+ charter <harness> opens another` … So it was pressed, and a sentence cannot be pressed.

And `slots.py:3548` records the same lesson from the `+N` reversal — *"a thing on screen that looked pressable, was pressed, and did nothing"* — where the fix was to make the drawn thing do the obvious thing rather than to document it.

## The rule this establishes

**A surface that advertises creating a thing must advertise unmaking it, at the same weight and in the same place.** An unadvertised create costs an operator a feature they did not know about; an unadvertised destroy makes them conclude the system will not let them do it, which is strictly worse and is what happened here.

### 1. `chat: new` as a palette row — this is a live violation of charter's own rule

`frame/events.py:70-72`: *"a component whose only route to a piece of state is a click has no route to it on most planes. **Give every pointer affordance a key as well.**"*

`[frame] mouse` ships **false** (`instance.py:1706`). The palette has no new-chat row. So **on a default plane an operator cannot create a chat from the frame at all**, and the single glyph charter advertises is unreachable. Fix the create side first; it costs nothing on screen.

### 2. A `−` beside the `+`, as an advertisement and not a shortcut

**It must not close on the press.** `builtins.py:526-534` draws the line: making a chat is acceptable on a press because *"nothing is destroyed"*, while `frame-quit` sits behind a palette row *"with a confirmation for exactly that reason"*. §4i: **a pointer opens the question; the keyboard answers it.**

[...]

Issue #903, opened 2026-09-05
A dragged tab strip keeps its height, tabs order by use, and the strip gets compact


diazoxide:
Four changes to the tab strips, reported together from a running plane.

## 1. A drag must stick

*"when switching between workspaces — resized tabs resetting to one line. it should preserve sizes."*

Measured on the reporting plane: **no frame has a recorded `bar_rows` at all.** So the operator is drag-resizing the tmux pane, not pressing `F3` — and `layout._grown` recomputes the height on the next layout pass, which a workspace switch triggers. The drag is discarded.

#880 chose a keybinding over drag-resize *because* the layout owns bar heights, and judged teaching the layout to respect a drag "a much larger change than the one-row default". That was the wrong trade: it leaves the obvious gesture failing **silently**, and a key nobody knows about as the only way to do the thing.

**On a pane resize, adopt the operator's height as this frame's choice** — writing the same stored value `F3` records, so both gestures have one meaning and one place to look.

The hazard, and it is the whole difficulty: **charter resizes these panes itself.** Adopting charter's own resize as "the operator chose this" would pin whatever the layout last computed and freeze the strip forever. So the comparison is against **what charter last asked for**, not against the previous height — a difference means a hand moved it.

## 2. Recency ordering, stable within a session

*"active last used tabs should be in first order, then olds."*

This conflicts with a measured decision and the conflict is the interesting part. `slots._cuts` rejects a mark-centred window outright: *"A window CENTRED on the marked tab moves every column each time the operator switches… on this project's own fifteen workspaces at 160 columns, six of the nine drawn tabs answer a second press at the identical column with a SECOND, different workspace."* `chats.of_workspace` rejects recency for the same reason: *"`api.1` stays leftmost, where an operator learned to look for it, instead of jumping to the end because it happens to be the newest."*

Both are right about **live** reordering. Neither argues against an order that is *fixed while you look at it*.

**Order workspaces by last use, computed once per frame launch and held stable for the session.** A switch updates the record for next time; it does not move a column now. The operator gets their working set up front, and the strip stays a surface you can aim at.

`workspace.last_touched` already derives a timestamp from mtimes — no new state.

Chats keep ordinal order. `api.1` staying leftmost is a stronger promise for a handful of numbered siblings than recency would be.

## 3. The active tab gets an edge, not a background

Today the active tab is `chrome.reverse` (SGR 7) with a `*` prefix.

[...]

Issue #882, opened 2026-09-04
The persona switcher stops lifting the selected persona, and sorts by use


diazoxide:
The persona switcher currently lifts the selected persona to the top of the list. That means the list reorders as you move through it, and the position of every other entry depends on where you already are.

**Remove it.** The default persona is pinned first, always. Everything else sorts by **dispatch count, descending**, tie-broken by memory count.

## Why dispatches rather than memories

Both numbers are available and they rank differently. On this plane today the memory counts read `release 51`, `steward 47`, `statusline 14`, `forge 13`, `reddit 7` — but `release` and `steward` were dispatched today and `reddit` on 2026-08-19.

**Memory count only ever grows**, so it ossifies: a persona used heavily one month outranks one used daily the next, permanently. Dispatch count measures *use*, which is what makes a persona worth reaching for.

"Last dispatched" was considered as a third option and is arguably sharper still, but it makes the list reorder far more often — which is the same defect as lifting the selected item, one cause over.

## Where the ordering lives

`switch.py:117` returns `sorted(n for n in p_mod.list_personas() if p_mod.valid_name(n))` — plain alphabetical. The selected-first behaviour is applied above it.

PR #926, opened 2026-09-07
The workspaces strip draws one order, and it is the plane's (#923)

### LINKED ISSUE #923: The workspace tab order is frozen per chat, so every switch shows a different order
*"when switching workspaces tabs — after each switch — order of workspaces tabs changing, feeling that each switch is opening new window."*

## Measured on the reporting plane

Three chats, three different recorded orders for the same plane-wide list of workspaces:

```
.charter/frame/default.1/tab_order            default           authority-audit  autonomy
.charter/frame/default.2/tab_order            default           harness-wrapper  authority-audit
.charter/frame/harness-wrapper.1/tab_order    harness-wrapper   default          authority-audit
```

Note `default.1` and `default.2` are two chats in the **same** workspace and still disagree.

## The cause

`switch._by_use(fid, names)` freezes the order the first time a frame asks, and never re-decides — that part works exactly as designed, and its docstring is right that *"a frame that re-decided on its next repaint would be the live reordering this exists to refuse."*

The defect is the key. `state.tab_order(fid)` is stored **per chat**, under `.charter/frame/<fid>/tab_order`, and seeded from `chats.touched_by_workspace()` at the moment that chat first renders — so it leads with whichever workspace was most recently touched, which is the chat's own.

**Switching workspaces switches chats.** So each switch draws a strip ordered by a different chat's snapshot, taken at a different moment. The strip is stable within a chat and unstable across exactly the gesture the strip exists for.

## Why this is the harm the original code refused

`slots._cuts` argued against recency ordering with a measurement:

> "A window CENTRED on the marked tab moves every column each time the operator switches… on this project's own fifteen workspaces at 160 columns, six of the nine drawn tabs answer a second press at the identical column with a SECOND, different workspace."

#903 overrode that on the grounds that the objection was about **live** reordering and said nothing against an order fixed while you look at it. That reasoning holds. What it missed is that the thing being held still was scoped to a chat, so the operator meets a different fixed order every time they switch — which delivers the same defect the measurement warned about, by a different route.

## The fix

**The order belongs to the plane, not to a chat.** The workspaces strip draws a plane-wide list; its order should be plane-wide too, decided once and read by every frame.

Points to settle while implementing:

- **Where it lives.** `.charter/frame/<fid>/` is per-chat by construction and `state.NO_FORMAT_PROMISE` says that tree carries no promise outside charter. A plane-scoped order is not frame state and probably does not belong there.
- **When it is re-decided.** "Once, ever" ossifies — a workspace created next month sorts last for good. "Once per plane launch" matches the intent of #903 and keeps the strip still for as long as an operator is looking at it.
- **What a new workspace does.** `_by_use` already appends unknown names in sorted order after the recorded ones, which is the right shape; keep it.
- **The per-chat `tab_order` files become dead.** Remove them and their writer, or state why they stay.

## Care

- The chats strip is **not** affected and must not change: `chats.of_workspace` is ordinal (`api.1` stays leftmost) and #829 recorded that decision deliberately.
- Anything that changes which names appear where re-enters `_cuts`, `bar_rows_wanted` and `_tab_columns`. Check the three together.
- `#767`'s rule stands: a wider bar must never draw fewer names, and a tab must never move under a press.

diazoxide:
Closes #923

*"when switching workspaces tabs — after each switch — order of workspaces tabs changing,
feeling that each switch is opening new window."*

#903's freeze was the right mechanism in the wrong place. `switch._by_use` decides the
order once and never re-decides — which is exactly what `slots._cuts` and
`chats.of_workspace` each demanded with a measurement. The **key** was the defect:
`state.tab_order(fid)` lived under `.charter/frame/<fid>/`, which `frame/state.py`'s
opening line says is per frame and never global. **Switching workspaces switches chats**,
so every switch handed the operator another chat's snapshot of the recency, taken at a
different moment.

## Reproduced, before and after

Three chats, each first painting at a different point in the plane's history — the
operator's own gesture, since arriving in a chat writes to it. Same script against both
trees:

```
before (origin/main @ 89ff785b)
  default.1          default  harness-wrapper  authority-audit  autonomy
  default.2          default  authority-audit  harness-wrapper  autonomy
  harness-wrapper.1  harness-wrapper  default  authority-audit  autonomy
  distinct orders drawn by 3 chats: 3

after (this branch)
  default.1          default  harness-wrapper  authority-audit  autonomy
  default.2          default  harness-wrapper  authority-audit  autonomy
  harness-wrapper.1  default  harness-wrapper  authority-audit  autonomy
  distinct orders drawn by 3 chats: 1
```

`default.1` and `default.2` are two chats of the **same** workspace, which is what makes
this about the key rather than about workspaces.

That reproduction is permanent as
`tests/test_the_workspaces_strip_draws_one_order_for_the_plane.py`. Run against
`origin/main`'s tree it fails 15 of its 20 cases, including
`test_a_tab_is_at_the_same_column_for_both_frames`, which reports the columns that moved:

```
(0, 23): ('harness-wrapper', 'default') … (0, 33): ('harness-wrapper', 'default')
```

— eleven columns answering a press with a second, different workspace. That is #767's rule
broken, and it is the same harm `_cuts` measured against a recency-centred window, arriving
by another route.

## The four things the issue asked to settle

**1. Where it lives.** `config.WORKSPACE_TAB_ORDER_FILE` — `.charter/workspace-tab-order`,
beside `active-workspace`, `sessions/` and `terminals/`. Plane-scoped, per developer,
gitignored, machine-written, `config.replace_for` (#894), 0600. Not under `.charter/frame/`:
that tree is per frame by construction and `state.NO_FORMAT_PROMISE` says so, and `reap`
deletes each frame's copy with the frame — which is the defect. The reader and writer are in
`charter/workspace.py`, next to `list_workspaces`, because the order of a roster belongs
with the roster; `switch._by_use` still decides it and `chats.touched_by_workspace` still
measures the recency (#882's split, unchanged).

**2. When it is re-decided.** Once per plane launch. `state.reap` drops it when a reap
takes every directory the frame root had — charter's one observer of a plane going cold,
already reaching one directory over for `_forget_session`. "Once ever" ossifies: a
workspace made next month sorts last for good. "Once per repaint" is the live reordering
`_cuts` refused. The count is over **every** entry, not this server's, so a frame in the
operator's own tmux — kept by reap's `owner != server` rule — keeps the strip still under
the hand that is looking at it (#767).

**3. A new workspace.** Unchanged: `_by_use` appends unknown names in sorted order after
the recorded ones, and that promise is still kept where it is made.

**4. The per-chat files.** `state.record_tab_order` and `state.tab_order` are **deleted**,
not kept beside the new record — two records of one order are two orders. A case asserts
neither name is reachable and that no frame directory grows the file.

`switch.workspaces` takes **no frame id at all** now. Which frame is asking was never a
question about the order, and asking it was the defect. One consequence is a widening the
issue did not ask for and #882 already argued for: the launch picker and a refusal's
`have: …` sentence used to fall back to plain alphabetical because they run with no frame
and had nowhere to record an answer. They now lead with the working set like everything
else.

## The constraints

* **`_cuts`, `bar_rows_wanted` and `_tab_columns` are checked together** in
  `TheThreeReadingsOfTheStripAgree`: two frames in different workspaces want the same rows,
  publish the identical column map, and no width from 60 to 280 drops more than the one
  name `_cuts`' own docstring admits to.
* **#767 holds.** No tab moves under a press across a switch — the case above — and the
  drop-by-at-most-one sweep is asked of this strip and not only of `_bar`.
* **No live reordering is reintroduced.** The order is still frozen; only its scope
  changed. `TheOrderIsHeldWhileThePlaneIsUp` keeps every one of #903's own cases.
* **The chats strip is untouched.** `chats.of_workspace` stays ordinal (#829), and
  `TheChatsStripIsUnchanged` is still the control.

## One test fixture, and why it was a fixture rather than a defect

`tests/test_a_real_resize_gives_a_real_strip_another_row.py` drives `_reassert_sizes`
against a synthetic frame id. `state.record_asserted_bars` deliberately takes no `create`
— *"a frame with no directory records nothing"* — and the directory existed there only
because `switch.workspaces(fid)` wrote a per-chat `tab_order` on the sizing path and minted
it on the way past. In production `cmd_resize` reads `state.panes(fid)` out of that same
directory before it ever reaches `_reassert_sizes`, so a real frame always has one. The
fixture now states what a launch does instead of leaning on a side effect.

## Shared files

`charter/frame/slots.py` is touched in one place, `_workspaces_strip` (8 lines), well away
from the chats strip's affordances — #921 is working there concurrently.

https://claude.ai/code/session_018hUBNraNfixfMMK5Jc6CNy

### DIFF
--- a/charter/config.py
+++ b/charter/config.py
@@ -774,23 +774,6 @@ def derive(root: Path, start: Path | None = None) -> dict:
     #: keeps its own workspace across restarts — without leaking into other panes.
     d["TERMINALS_DIR"] = state / "terminals"
 
-    #: The order the **workspaces tab strip** draws its names in — one name per line,
-    #: decided once per plane launch and read by every frame (#923). See
-    #: `charter.workspace.tab_order`.
-    #:
-    #: **Here rather than under `.charter/frame/<fid>/`, and that placement IS the fix.**
-    #: #903 put it there, where `frame.state.reap` deletes it with the chat that wrote it
-    #: and where every chat therefore has one of its own — and switching workspaces
-    #: switches chats, so the strip drew a different chat's snapshot after every switch.
-    #: The strip is a plane-wide list, so its order is a plane-wide fact, and it belongs
-    #: beside the other plane-wide, per-developer, machine-written pointers here rather
-    #: than inside a tree whose every path is scoped to one frame by construction and which
-    #: says so in `frame.state.NO_FORMAT_PROMISE`.
-    #:
-    #: Neither committed nor `charter.toml`: it is a machine's reading of which workspaces
-    #: were in use, rewritten whole at the next launch, and nothing a hand maintains.
-    d["WORKSPACE_TAB_ORDER_FILE"] = state / "workspace-tab-order"
-
     #: Persona definitions — **committed** and shared with the team (unlike vaults). A
     #: persona is a directory ``personas/<name>/`` holding ``persona.md`` (the definition),
     #: ``memory/`` and ``refs/`` (persistent, committed knowledge). The legacy flat
--- a/charter/config.py
+++ b/charter/config.py
@@ -774,6 +774,23 @@ def derive(root: Path, start: Path | None = None) -> dict:
     #: keeps its own workspace across restarts — without leaking into other panes.
     d["TERMINALS_DIR"] = state / "terminals"
 
+    #: The order the **workspaces tab strip** draws its names in — one name per line,
+    #: decided once per plane launch and read by every frame (#923). See
+    #: `charter.workspace.tab_order`.
+    #:
+    #: **Here rather than under `.charter/frame/<fid>/`, and that placement IS the fix.**
+    #: #903 put it there, where `frame.state.reap` deletes it with the chat that wrote it
+    #: and where every chat therefore has one of its own — and switching workspaces
+    #: switches chats, so the strip drew a different chat's snapshot after every switch.
+    #: The strip is a plane-wide list, so its order is a plane-wide fact, and it belongs
+    #: beside the other plane-wide, per-developer, machine-written pointers here rather
+    #: than inside a tree whose every path is scoped to one frame by construction and which
+    #: says so in `frame.state.NO_FORMAT_PROMISE`.
+    #:
+    #: Neither committed nor `charter.toml`: it is a machine's reading of which workspaces
+    #: were in use, rewritten whole at the next launch, and nothing a hand maintains.
+    d["WORKSPACE_TAB_ORDER_FILE"] = state / "workspace-tab-order"
+
     #: Persona definitions — **committed** and shared with the team (unlike vaults). A
     #: persona is a directory ``personas/<name>/`` holding ``persona.md`` (the definition),
     #: ``memory/`` and ``refs/`` (persistent, committed knowledge). The legacy flat
--- a/charter/frame/builtin_actions.py
+++ b/charter/frame/builtin_actions.py
@@ -334,7 +334,10 @@ class _Strip(NamedTuple):
 _STRIPS = (
     _Strip("chat", lambda fid: [c.id for c in chats.roster(fid)], lambda fid: fid,
            ("frame-chat",), chats.ONLY_CHAT),
-    _Strip("workspace", switch.workspaces, switch.current_workspace,
+    # The workspace lister takes no frame since #923 — the strip draws a plane-wide
+    # list, so its order is the plane's and not this chat's — and the lambda is what
+    # keeps `names(fid)` one signature for both strips rather than two.
+    _Strip("workspace", lambda _fid: switch.workspaces(), switch.current_workspace,
            ("frame-switch", "--workspace"), ONLY_WORKSPACE),
 )
 
--- a/charter/frame/choose.py
+++ b/charter/frame/choose.py
@@ -202,17 +202,22 @@ def names_of(noun: str, fid: str = "") -> list[str]:
     `workspace.valid_name` / `persona.valid_name` / `instance.change_name_ok` on the way
     out, which is #442's rule applied where a directory name becomes a row.
 
-    *fid* is required in practice for :data:`CHANGE` and for :data:`WORKSPACE`, and ignored
-    by :data:`PERSONA`. Changes are per WORKSPACE and which workspace is a question about
-    the frame, not about this process (#512); workspaces are ORDERED per frame since #903,
-    so that the palette and the strip beside it draw one roster in one order. It defaults
-    to `""` rather than being made positional-required so that the two callers that predate
-    it read unchanged; `switch.changes("")` resolves the same "no frame, resolve locally"
-    path every other frame surface takes for an empty id, and `switch.workspaces("")`
-    answers alphabetically rather than inventing an order it has nowhere to record.
+    *fid* is required in practice for :data:`CHANGE` and for :data:`CHAT`, and ignored by
+    :data:`PERSONA` and by :data:`WORKSPACE`. Changes are per WORKSPACE and which workspace
+    is a question about the frame, not about this process (#512). It defaults to `""`
+    rather than being made positional-required so that the two callers that predate it read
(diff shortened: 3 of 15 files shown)


diazoxide:
## Deletion sweep

The verdict on `2628b83e` was **`1 survivor, 1 not measured`**. Both are addressed.

**`charter/frame/state.py:2452` — `[swap-synonym] sorted → list` — PINNED.**
The `sorted` predates this branch; moving the walk into a comprehension put this
branch's name on the line. It had never been pinned because every case in the
repository that spells an expected `removed` list removes exactly one directory. It is
not decoration: the walk is over a directory listing, whose order is the filesystem's,
and `reap` returns what it removed. `tests/test_frame_state.py` now makes eight chats in
reverse name order and asserts the forward one — a fixture whose creation order and name
order agreed would pass with the sort gone. Hand-applied `list`, watched it go red
(`ws-2.1, ws-6.1, ws-1.1, …`), reverted, `__pycache__` cleared on both transitions.

**`charter/config.py:792` — `[retune-string] "WORKSPACE_TAB_ORDER_FILE"` — measured
locally; the shard ran out of budget, not out of tests.** The shard reported
`timed out after 900s` at its own fifty-minute report point, which is "not measured" and
not a pin. Measured by hand instead: renaming the key to `XWORKSPACE_TAB_ORDER_FILEX`
takes the attribute off `charter.config`, and the two covering modules go red in **0.22
seconds** with 36 errors (`AttributeError: module 'charter.config' has no attribute
'WORKSPACE_TAB_ORDER_FILE'`). The literal is pinned; it was the 326-module fan-out that
did not fit in what was left of the shard.

The other eight mutations came back pinned, including the three the issue's own care list
predicted would matter — `state.py:2455` (`drop-if` on the liveness keep), `state.py:2452`
(`drop-comprehension-if` on `is_dir`), and `workspace.py:878` (`swap-synonym` on
`line.strip`).

diazoxide:
## Rebased onto `09aa3b0e`

`main` moved three commits while this was open — #925 (the `-` beside the `+`), #928 (the
quit confirmation), #924 (the sweep's refusal reporting). Rebased, head is `845e6a13`.

**One hand-resolved conflict**, in `charter/frame/slots.py` at `_workspaces_strip`'s
return, and it is one line. #925 gave every bar strip a fifth field — the close
affordance — and this branch took the `fid` off the names. Both changes land on the same
statement:

```python
    return (switch_mod.workspaces(), switch_mod.current_workspace(fid), "", "",
            _workspace_counts())
```

Both sides kept whole: #925's five-field shape (`bar_rows_wanted` unpacks five) and this
branch's frameless lister. The two docstring additions merged without a conflict — #925's
paragraph on why the workspaces strip has no `-` sits above this branch's on why its names
take no *fid*, and they are about different halves of the same tuple.

Nothing else collided. `builtin_actions.py` is five lines in the `_STRIPS` table, which
#925 did not touch; `builtins.py` and `leave.py` are untouched by this branch.

**`_cuts`, `bar_rows_wanted` and `_tab_columns` re-checked together** after the rebase, not
one at a time — `TheThreeReadingsOfTheStripAgree` runs all three of two frames in different
workspaces, and `test_frame_bars` plus
`test_a_tab_strip_grows_a_row_when_its_tabs_overflow` were run in the same pass. 395 tests,
green. The full suite on the rebased tree: `Ran 11801 tests — OK (skipped=9)`.

**The reproduction re-measured against the new `main`**, not the old one:

```
before (origin/main @ 09aa3b0e)          after (this branch @ 845e6a13)
  default.1          default  harness-wrapper  authority-audit  autonomy
  default.2          default  authority-audit  harness-wrapper  autonomy      → all three identical
  harness-wrapper.1  harness-wrapper  default  authority-audit  autonomy
  3 orders for 3 chats                       1 order for 3 chats
```

**The news entry does not overlap the three unreleased ones beside it.** #921's is about
closing a chat, #928's about the quit confirmation's pane, #924's about the sweep; this one
is the workspaces strip's order, and none of them claims anything about another's subject.

PR #968, opened 2026-09-11
`charter claude` runs the command you launch Claude Code with, from `.charter/local.toml`

Biacode:
## What

`charter claude` execs `claude` by name. If you reach Claude Code through something else — `ccs work`, a wrapper that picks which account a chat is billed to; a binary you keep off `$PATH` — you had two doors, both wrong: `charter frame -- ccs work` runs the words and forgets the harness (no `$CHARTER_HARNESS` in the pane, no workspace layer, nothing `charter reopen` can resume), and a shell alias is invisible to an `execvp`.

This adds a **per-developer** file that names the launcher, one table per harness:

```toml
# .charter/local.toml — yours, never committed
[harness.claude]
command = ["ccs", "work"]
```

`charter claude` then runs `ccs work` in `claude`'s place, your arguments follow it (`charter claude -p hi` → `ccs work -p hi`, `--resume <id>` on a reopen), and it is still the Claude Code harness charter is launching.

## Why `.charter/local.toml` and not `charter.toml`

The README's containment rule: *a name charter reads out of a committed file cannot choose what it runs*. That is why `instance.harness_of` only ever *compares* `[harness] default` against the registry and refuses `"clyde"`. A `command` is exactly that choice, so it cannot live in a committed file. `.charter/` is gitignored on every plane `init` wrote, its files are one account's (`STATE_FILE_MODE`), and it already holds the per-developer half of the vault registry over the committed one — the precedent this follows. It is also the right scope on the merits: which account a chat is billed to is one operator's business, and a teammate on the same plane may run plain `claude`. `$CHARTER_HOME` moves the file with the rest of `.charter/`.

## How

- `harness/base.py` — `Harness.launch_argv(extra, command=None)` returns `[*(command or [binary]), *extra]`; still a list, never a joined string.
- `instance.py` — `LOCAL_FILE`, `load_local(state_dir)` (`load` without the schema check), `harness_local_of(cfg)` → `{"command": {cli_name: [words]}, "refused": [sentences]}`. Tables are keyed by the word after `charter` (`claude`, not `claude-code`), `harness_of`'s rule. A table for a harness charter cannot launch, a `command` that is not a non-empty list of non-empty strings (`"ccs work"` is refused, not split — there is deliberately no shell), or a value where a table should be, is **refused and named** through `contain.readable`.
- `config.py` — derives `LOCAL_CONFIG`, `LOCAL_CONFIG_ERROR` (a malformed file degrades, the way `charter.toml`'s does) and `HARNESS_LOCAL`.
- `commands_frame._launch` — passes the override into `launch_argv`; the pre-tmux `shutil.which` check now asks about `argv[0]` rather than `h.binary`, still scoped to `if h` so `charter frame -- <cmd>` is not narrowed (its test's docstring updated to say so).
- `doctor.py` — a `local.toml` row beside `charter.toml`: WARN naming each refused table or the parse error (charter carries on with the binary, which is a working plane), OK with `claude → ccs work` when in force, OK `no per-developer overrides` otherwise. Added to `_FIXED_CHECK_NAMES`.
- `commands_harness.py` — `charter harness list` prints `→ runs: ccs work` under a harness whose command is in force.

## Tests

`tests/test_a_harness_runs_through_the_command_you_name.py` (new) plus two cases in `tests/test_harness_launch.py`, written first and watched fail:

- the parser: every registered harness, unknown names, `claude-code` vs `claude`, every wrong shape of `command`, a value where a table should be, `refused` is a list on every path, newlines in names and values cannot forge a report line;
- `config`: the file reaches `HARNESS_LOCAL`, absent/malformed/non-UTF-8 degrade, the settings are in `DERIVED`, `$CHARTER_HOME` moves the file;
- the launcher: `--no-frame` execs `["ccs", "work", "-p", "hi"]`, the framed path hands `new-session` the command's words verbatim, the pre-tmux check refuses a missing `ccs` before tmux and names it, and a `claude` off `$PATH` does not stop a `ccs` that is on it;
- `doctor`: row name pinned beside `charter.toml`, each status, hint clipping past three refusals, a newline in a command word cannot forge a second row;
- `harness list`: shows the command, and only for the harness it launches.

Full suite locally: `python3 -m unittest discover -s tests` → 12,024 tests, OK (20 skipped), Python 3.13.

Deletion sweep locally (`tools/sweep.py --gate --base main`): 44 mutations, **39 pinned, 2 survivors, 3 unresolved**. The two survivors were both on the `local.toml` row's hint-clipping line (`> 3` vs `>= 3`, and an unconditional ellipsis); the second commit pins them with a one-refusal case asserting no ellipsis and an exactly-three case on the boundary, each watched turn red under its mutation first. The three unresolved are mutations that kill the runner rather than a test (`if h: return bypass(argv)` execs a real `claude` over the test process on a machine that has one; dropping the parser's `isinstance(section, dict)` guard raises inside `config.derive` at import), so zero tests ran and the sweep declines to count that as evidence. Each is pinned by a direct test in an ordinary run; CI's runner has no `claude`, so the first should resolve there.

## Scope, and a possible follow-up

The override is per **plane** and per developer: every workspace in a plane launches the same way, and a work/personal split is two planes with two `local.toml`s. A per-**workspace** override would need a second per-developer location (`.charter/workspaces/<ws>/local.toml`, say) and a lookup in the launcher once it knows the workspace; it is left out of this PR deliberately, and is a small follow-up if anyone wants it. `charter init`, `doctor --fix` and `charter update` still shell out to plain `claude` for plugin installation — the override is about launching a chat, not about which `claude` administers plugins — and the docs say so implicitly by scoping the setting to `charter <harness>`.

## Docs

- `docs/control-plane.md` — new section *`[harness.<name>].command` — your own launcher, in `.charter/local.toml`*, plus a pointer in the annotated `[harness]` sample.
- `docs/frame.md` — a paragraph under the launch commands.
- `README.md` — one sentence after "`charter claude` needs `claude` on your `PATH`".
- `docs/news/unreleased-charter-claude-runs-the-command-you-launch-claude-code-with.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,10 @@ session, after `init` and `--fix` alike, so restart any session that was already
 **The recording stops before `charter claude`**, because the frame is a full-screen tmux
 client and `capture-demo.sh` records a line transcript; the picture at the top of this page
 is what that step opens. `charter claude` needs `claude` on your `PATH`: without it no frame
-is drawn, and charter says the binary is not installed and exits 127. The frame also needs
+is drawn, and charter says the binary is not installed and exits 127. If you reach Claude
+Code through something else — `ccs work`, a binary off `PATH` — `.charter/local.toml` names
+it, per developer and never committed
+([docs/control-plane.md](docs/control-plane.md#harnessnamecommand--your-own-launcher-in-charterlocaltoml)). The frame also needs
 tmux, and of tmux only its absence stops a launch — below 3.2, the version charter checks
 its requirements against, the frame still starts. `charter opencode` and `charter codex`
 need their own binaries the same way, and `charter claude --probe` says whether a frame can
--- a/charter/commands_frame.py
+++ b/charter/commands_frame.py
@@ -2553,8 +2553,9 @@ def _guest_harness_env(env: dict[str, str]) -> dict[str, str]:
     servers is in that one extra name. On charter's own private server the base the `-e`
     overlays is a server SOME charter launcher started, so its `$PATH` is a charter
     launcher's. On the operator's it is a server THEY started, possibly weeks ago, in
-    another shell — and `cmd_launch` has already resolved the harness binary
-    (`shutil.which(h.binary)`) against charter's OWN `$PATH`. Handing the pane a
+    another shell — and `cmd_launch` has already resolved the harness's first word
+    (`shutil.which(argv[0])` — its binary, or the operator's own `[harness.<name>]
+    command`) against charter's OWN `$PATH`. Handing the pane a
     different one would make that check a promise charter cannot keep: the frame comes
     up, the exec fails, and the operator gets `_UNKNOWN_DEATH_CODE` for a binary charter
     said it had found.
@@ -4542,7 +4543,8 @@ def early_death_message(argv: list[str], code: int, last_words: list[str]) -> st
 
     **The design call #384 left open, and the argument for taking it this way.** `charter
     <harness>` can refuse a missing binary before tmux (see `cmd_launch`'s `if h and not
-    shutil.which(h.binary)`), because charter chose that name itself. `charter frame --
+    shutil.which(argv[0])`), because that word is either a name charter chose itself or one
+    the operator wrote into their own `.charter/local.toml` as an executable. `charter frame --
     <cmd>` cannot be closed the same way without changing what it ACCEPTS, and what it
     accepts is tmux's rule, not charter's — verified against tmux 3.7c:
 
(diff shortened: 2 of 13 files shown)


diazoxide:
Thank you, Artur. This PR shaped how charter now launches a harness through your own command.

The mechanism is superseded by **harness profiles** (ADR 0022), which landed this week. A profile in the gitignored `charter.local.toml` can name any command, including a wrapper like `ccs work`:

```toml
[harness.claude-ccs]
kind = "claude"
command = ["ccs", "work"]
```

On top of this PR's per-harness command, profiles add several profiles per kind, per-profile environment, a one-time approval before a new or changed command runs, a check that charter's plugin is actually wired, and a selector when a chat opens.

What this PR specified that profiles had not yet pinned is now on `main` under your name, in #1003:
- **716d8f8** — your launch tests, re-pointed at profiles:
  - the command's words come first and the operator's arguments (`--resume <id>`) follow;
  - the wrapper is the program checked before tmux, so a missing `ccs` is refused and a missing `claude` does not stop `ccs`;
  - a non-UTF-8 local file is refused with a sentence.
- **75493d2** — your `ccs work` use case in `docs/harnesses.md`, plus the `$CHARTER_HOME` note.

Closing this one in favour of those. Thanks again for the careful work.

PR #1103, opened 2026-09-15
A cross-server chat tab names its way back, Esc closes a selector on either server, a workspace tab joins its own session, and a transcript viewer dies with its chat

diazoxide:
Four defects the operator hit after the per-plane tmux server upgrade (ruling 46, ADR 0023), which splits a plane's chats across the legacy `charter` socket and this plane's own `charter-plane-<hex>` server on upgrade day. Each was diagnosed red-loop-first.

## 1. A chat tab on the other server names its own way back

A client cannot move between two tmux servers, so pressing a tab for a chat still on the old server can only refuse. It no longer reads as an ordinary tab:

- the chat strip marks it — `~` in the mark cell where an idle tab is blank (`slots._OFF_SERVER_MARK`);
- `F2 → chat` says "on another server — charter frame-quit, then charter" on its row (`choose.OTHER_SERVER_NOTE`);
- the refusal names the way out in the same breath: quit that frame with `charter frame-quit` in its project, then `charter` brings its chats back on this plane's own server (`chats.OTHER_SERVER`).

One reading (`chats.off_server`) feeds all three so they cannot disagree, and the server comparison is now `tmuxctl.same_server` with the `LEGACY_SOCKET` fallback, so a blank record and a spelled `charter` read as the one legacy server they share.

## 2. Esc closes a selector on either server

`launcher.framed_chat` and `_close_the_cancelled_chat` proved the selector pane against the single server the record named. An upgrade-day record pointing at the wrong server left Esc asking it, finding nothing, and closing nothing. They now probe the servers a chat can be on — recorded one first, then this plane's own, then an operator's tmux, then legacy (`_prove_this_pane`). **The pid proof is untouched and is what makes probing safe:** `live_pane_by_pid` matches a live pane with this process's own pid — one pane on the machine — so no probe can reach another plane's window.

## 3. A workspace tab joins its own live session

Pressing a tab for a workspace whose session is already running on this plane's own server refused it as "probably another plane's". On a per-plane server the only sessions are this plane's — measured on the operator's machine, that `marketing` session carried this plane's own `@charter_plane`. The refusal in `_open_workspace` and the background/handoff path is now gated behind `_is_own_plane_server`, so the tab focuses/joins the session (opening a chat if it has none). The refusal is kept exactly as is for the legacy socket and an operator's own tmux, where another plane still can hold a session of that name.

## 4. A transcript viewer dies with its chat

"Previous transcript" opens a pager in its own window beside the chat, carrying no `@charter_chat`. Closing, quitting, or the harness dying left that window standing — a session holding nothing charter recognised, which defect 3 then refused. Viewers are now marked (`@charter_transcript`), killed with the chat on close and quit (`_stop_chats` → `_kill_transcript_windows`), and swept when their chat is no longer live before a launch reads liveness (`_sweep_orphan_transcripts`); one viewer is kept per chat. Measured on real tmux that a `pane-died` hook cannot reliably kill a sibling window, so the harness-died case is covered by the sweep rather than a hook.

## Tests

- `chats.check` cross-server refusal wording + `chats.off_server` + `slots._OFF_SERVER_MARK` on the strip + `choose` `F2 → chat` note;
- `framed_chat`/`_close_the_cancelled_chat` server probe (unit, RED under the old single-server ask) and Esc on a real legacy-socket selector (real tmux); the existing plane-socket Esc case stays pinned;
- a workspace tab and a background open joining an own-server session (updating the pre-ruling-46 refusal tests);
- real-tmux viewer teardown on close and quit, an orphan sweep after a dead chat, and one-viewer-per-chat.

News: `docs/news/unreleased-the-per-plane-server-upgrade-stops-stranding-chats.md`.

Modules run locally, each `( unset TMUX TMUX_PANE; python3 -m unittest tests.<module> )`: the four fix modules plus the background-open, teardown, two-planes, launcher, palette, handoff, tab-click, slots and switch modules — all green.

https://claude.ai/code/session_01CgvXBsD94ZwsnaGt3Wi5S5


diazoxide:
Follow-up commit on this branch: CI was red on the first commit for one reason in two shapes. The orphan-transcript sweep made its own `list-windows` from `_launch`, which was both the seventeenth invocation the launch-cost pin (`test_a_frame_sends_one_invocation_per_batch`) refuses, and a tmux call the launch fixtures do not stand in for — `TheLauncherTakesTheDecision` and five sibling fixtures mock `_live_chats`/`_live_sessions` by name and leave `tmuxctl.run` real, so the new read reached a socket and `tests/_planeguard` raised (thirteen ERRORs).

`_live_chats` now reads the chat, the viewer's chat and the window id in **one** listing (`_LIVE_CHATS_FORMAT`) and carries the viewers on its answer (`_LiveChats`, a `set` subclass, so every union, membership test, `is not None` and every `set()` stand-in is untouched — a plain set carries no viewers and the sweep no-ops, which is right when nothing real was listed). The sweep consumes those rows and lists nothing; `_kill_transcript_windows` takes rows a caller already holds. The launch stays at sixteen invocations.

Fields are separated by `|` rather than `\t`, and the reason is pinned: `TheChatListIsParsedLineByLine` holds that a one-field line with a tab at its edge still names its chat, which a tab split would read as three fields with an empty chat. Exactly three fields is the structured row; anything else is the stripped whole line, so `reap`'s keep list stays over-keeping and a viewer is only ever killed off a row whose chat and window both hold to their shapes. One assertion updated: `test_the_reap_set_carries_the_chats_as_well_as_the_sessions` now checks that a listing *asks* the chat option and none asks the window name, rather than equality with the old one-field format.

Locally green after the fold: the two red modules, `test_frame_launcher`, and the launch, handoff, teardown, two-planes, selector and real-tmux nets.

diazoxide:
Round 3 (this round's commit): the 16 deletion-sweep survivors on `92d1733` and the re-verification's six items, in one push.

**Viewer identity, closed as a class.** The viewer's plane now lives in `@charter_viewer_plane`, a window-only option no session ever carries, so tmux's window-then-session fallback can never hand a stamp-less window its session's marker. `cmd_transcript` writes the stamp FIRST and the chat mark only on success; `_live_chats` carries a row as a viewer only with both present; `_kill_transcript_windows` kills only when the stamp equals `_this_plane()`. The `(own and not plane)` exception is deleted. Red without their guard: a stamp-less window in a session whose session-level `@charter_plane` is the killer's survives close, quit and transcript on a shared server and on the plane's own; the positive control (A's own stamped viewer dies on A's close and quit) is red when the veto is `False`.

**Survivors → outcome** (each replayed by hand: mutate, drop `__pycache__`, run, restore, sha-check):
- `viewer_of and`, `window and` — deleted; `_FRAME_ID_RE.fullmatch` / `_WINDOW_ID_RE.fullmatch` already refuse the empty string.
- `_transcript_windows` `is None` arm, `_sweep_orphan_transcripts` `is None` return — deleted; `getattr(…, "viewers", ())` yields no rows either way, so nothing is aimed at.
- `p.strip` vs `lstrip`; the two viewer-row matches — pinned (`TheChatListIsParsedLineByLine`).
- empty-set return; `chat not in chats_set`; the `rows` else-branch — pinned (`TheKillIsAimedAtAWindowTmuxJustNamed`).
- both halves of the kept shared-server refusal in `_how_a_background_open_would_go`; the `in _live_sessions` half in `_open_workspace` — pinned by standing `_is_own_plane_server` in as False, the documented condition.
- `_off_server`'s `except Exception`; `roster`'s `else frozenset()`; `_candidate_servers`' dedupe — pinned.

**Also:** `_stop_chats`' window-id test asserts the exact call sequence (the kill, then one `list-windows`, nothing else). `EscClosesASelectorRecordedOnTheLegacySocket` is deleted — its `LEGACY_SOCKET` patch never reached the pane's process, so it pinned nothing; the skew case stays and is red on main. Wording: `OTHER_SERVER`, `OTHER_SERVER_NOTE`, `frame.md` and the news now say `charter frame-quit` records and stops every chat of the project on every server, the frame you are in included, and that `charter reopen` (or `charter`) brings them all back; the viewer paragraphs say exactly what the tests prove.

Issue #1116, opened 2026-09-16
Real-tmux tests fail for runner load, not behaviour: wall-clock bounds against the sweep's shards


diazoxide:
Every real-tmux test waits for tmux to do something by polling until a wall-clock bound. On a quiet machine those waits finish in a second or two. On CI they run beside the deletion sweep's shards — up to 14 of them against a 20-job ceiling — and the same wait can blow a bound that has three orders of magnitude of headroom locally.

That makes a class of test that fails for load, not for behaviour. A red run then means "the runner was busy", which is the most expensive kind of red: it teaches everyone to re-run rather than to read.

## What happened (the case that prompted this)

On #1113 (`8b2d08a`), `test (3.12)` failed and the other three interpreters passed on the same runner image:

```
FAIL: test_a_signal_death_ends_the_tab
      (test_an_ended_harness_keeps_its_tab_on_a_real_server.AnExitOnARealServer) (ending='exit 0')
AssertionError: False is not true : no exit code was recorded for beta.1:
  chat=beta.1 pane=%0 dead='1'
  ended=False exit=None drawn=True
```

Run `35078103480`, job `104735211464`.

Diagnosed as contention, not behaviour, on five pieces of evidence: the hook writes `${v:-N}`, so even a statusless death records a number and `exit=None` means the hook had not run at all; the failure was on the `exit 0` arm, where the status certainly exists; `ended=False` showed neither `pane-died` hook had left a trace; the other three Python versions passed on the same image, and a tmux or kernel gap cannot be Python-version-dependent; and the class completes locally in 4.5s, and 1.9s on a second run, against the 10s that blew.

The immediate fix was one bound raised to 45s, stated once with that measurement. That reduces the flake; it does not remove the class, and a timing bound cannot be pinned by a mutation, so it is a number chosen from evidence rather than a guarantee.

## Two honest options

1. **Serialise the real-tmux jobs against the sweep** so they never contend — a workflow-level change (job dependencies, or a concurrency group the sweep shares).
2. **Assert on the hook's completion rather than on wall-clock** — wait for an event tmux can report, so the test's patience does not depend on how loaded the box is.

Either removes "the runner was busy" as a way for the suite to go red. The first is cheaper; the second is stronger.

## Expected

- A real-tmux test that is red only when charter's behaviour is wrong.
- If a wall-clock bound survives anywhere, it states the measurement it came from, as `_HOOK_SECONDS` now does.

[...]

PR #1120, opened 2026-09-16
CI runs the real-tmux tests on a tmux that can report a death (#1116)

diazoxide:
Closes nothing on its own — it removes the cause #1116 is about, and says on that issue what it does not fix.

## The finding

`ubuntu-latest` ships tmux **3.4**, and tmux 3.4 can lose a SIGCHLD.

When it does, the server still gets the pty EOF, so `server_destroy_pane` runs and closes `wp->fd` — **which is the entirety of `#{pane_dead}`**. It then returns at `server-fn.c:329`, on

```c
remain_on_exit != 0 && (~wp->flags & PANE_STATUSREADY)
```

**before** `notify_pane("pane-died")`. `PANE_STATUSREADY` is set only by `server_child_exited`, and that is the function the lost signal never reaches.

So the pane is left reading `#{pane_dead}` `1` with `#{pane_dead_status}` and `#{pane_dead_signal}` **both empty**, and the `pane-died` hook — which charter's whole ended-tab presentation runs off — has not fired and never will. It is permanent, not late.

## The measurements

With charter stripped out entirely: one tmux server, `remain-on-exit on`, a `sleep 300` pane, one `pane-died` hook, `kill -9` on the pane's pid.

| tmux | misses |
|---|---|
| 3.4 | **8 in 60** |
| 3.4 (same image, same libevent) | **5 in 40** |
| 3.5 (same image, same libevent, same script) | **0 in 40** |

And charter's own gate case, `AnExitOnARealServer.test_a_signal_death_ends_the_tab`: **5 misses in 30** on 3.4, **0 in 30** on 3.5.

That is why `test_a_signal_death_ends_the_tab` has failed on `main` and on two pull requests, across `test (3.12)`, `(3.13)` and `(3.14)`, and why three deletion-sweep shards refused their entire slice with *"the baseline is red"*.

**The Python-version argument in #1116 does not survive this.** Which interpreter happened to be red is scheduling noise on a rate, not evidence about a capability — a rate of roughly one in eight, spread across a four-job matrix, looks exactly like "one version is flaky".

## The pin

`.github/actions/tmux` — a local composite action — builds **tmux 3.5a** from the tmux project's own release tarball, pinned by sha256 and checked with `sha256sum -c`, and removes the distro's 3.4 so there is one tmux on the machine.

Built rather than installed because there is no package: `noble`'s tmux *is* 3.4, a newer distro's `.deb` is built against a different libevent, and a third-party prebuilt binary is a supply-chain step this repository declines elsewhere for the reason `test.yml`'s `dev-install` note gives.

Every job that runs the suite uses it, and only those:

| workflow | job | why |
|---|---|---|
| `test.yml` | `test` | the matrix |
| `release.yml` | `test` | the last gate before an irreversible PyPI upload |
| `sweep.yml` | `plan` | traces the **whole suite** to build the selection map |
| `sweep.yml` | `sweep` | its first act is the unmutated baseline; a red baseline refuses the whole slice |

`sweep.yml`'s `collect` and `test.yml`'s `dev-install` are deliberately absent — neither opens a pane.

## How a downgrade fails loudly

Twice, because one check is not enough:

1. **In the job, before the suite starts.** A step of its own reads `tmux -V`, parses it, and exits non-zero below the floor with a `::error::` naming the mechanism and the miss rates. Separate from the build step on purpose: a downgrade that comes from a `$PATH` the action does not own, or an edited build, must still hit it.
2. **From inside the suite.** `tests/test_ci_runs_a_tmux_that_reaps_its_children` asserts the same floor on a real runner (`GITHUB_ACTIONS`, not `$CI` — a developer exporting `CI=1` has not agreed to have their suite refused on their distribution's tmux), **and** reads the workflows to pin the set of suite-running jobs whole. The first check can only speak for the job it runs in; a *new* job that runs the suite without the pin would never reach it.

The floor is written once, in `CI_TMUX_FLOOR`, and the action's `FLOOR_MAJOR`/`FLOOR_MINOR` are held to it by that module — #670's shape applied to a version instead of a duration.

**It is a CI floor and deliberately not charter's own.** `tmuxctl.FLOOR` is unchanged: raising it would turn a CI defect into a refusal aimed at operators running 3.4 perfectly happily.

## The diagnostic gap

This is the half that matters beyond this PR. `_why` printed `dead=` and **neither of the two fields beside it** — so the real failure

```
chat=beta.1 pane=%0 dead='1'
ended=False exit=None drawn=True
```

could only be read as *"the hook fired and charter mishandled it"*, which is the reading #1116 recorded before raising a wall-clock bound against a contention theory this refutes. The one thing that would have named the cause was the thing not printed.

It now prints `pane_dead_status`, `pane_dead_signal`, a legend for what the empty combination means, and the version of the server that owns the pane. Forced on a healthy tmux, the SIGKILL arm reads:

```
chat=beta.2 pane=%1 dead='1' pane_dead_status='' pane_dead_signal='kill'
  (legend: dead=1 with BOTH of those EMPTY means tmux destroyed the pane without ever
   firing pane-died — a SIGCHLD the server lost, not charter. Below tmux 3.5 that is
   about 1 run in 8. See #1116.)
  server tmux='3.7c'
```

`pane_dead_signal='kill'` is exactly the field that is empty on 3.4. The next failure of this shape names itself.

## What is deliberately not changed

- **The 45 s bound.** The measurements say the event arrives promptly or never arrives, which is not evidence about what a contended runner needs. Shrinking it on this finding would repeat the mistake that set it. The note beside it is *corrected rather than deleted* — the reasoning is the record, and a wrong "why" left standing is how the next reader gets sent the same way.
- **The test itself.** Not skipped, not re-timed, not softened.
- **`tmuxctl.FLOOR`,** and therefore anything an operator sees. So **no news entry**: nothing operator-visible moves. What tmux 3.4 costs an *operator* — the same lost hook, in their own frame — is a real question and a separate one; it is raised on #1116 rather than folded in here.

## What this does not fix

The pin removes the cause of the failures actually observed. It does **not** make the real-tmux tests independent of wall clock: #1116's second option — waiting on an event tmux reports rather than on a bound — is still open, and still the stronger answer. Said there, not duplicated as a new issue.

## Checks

- `tests.test_ci_runs_a_tmux_that_reaps_its_children` (new)
- `tests.test_an_ended_harness_keeps_its_tab_on_a_real_server`
- `tests.test_workflows` — one case widened: a local `uses:` carries no `@` by GitHub's rule, and this repository had no local action until now. Carved out the way `EveryActionIsPinnedToSomethingImmutable` already carves it; a third-party ref with no version is still a failure.
- `tests.test_sweep`


diazoxide:
**Green on `8f73823`, and one side finding this run turned up that is not mine to fix here.**

| check | result |
|---|---|
| `test (3.11)` | pass, 10m01s |
| `test (3.12)` | pass, 8m43s — *the interpreter that used to fail* |
| `test (3.13)` | pass, 10m16s |
| `test (3.14)` | pass, 10m43s |
| `Size the sweep` | pass, 9m55s |
| `Sweep shard 1 of 1` | pass, 36s |
| `Add up what the shards found` | pass |
| verdict | **`nothing to sweep`** — 0 shards refused |
| CodeQL / Analyze (actions, python) | pass |

Every job reported `tmux on this runner: tmux 3.5a`, and `test_the_tmux_on_this_runner_is_at_or_above_the_floor` **ran rather than skipped** on all four interpreters.

**What the sweep does and does not prove here.** The diff touches no file under the swept paths, so the plan was `0 mutations → 1 shard` and the shard short-circuited at 36s — it never reached the unmutated baseline. So the sweep is *not* the evidence that the baseline is green under the pin. What is: the four matrix jobs, each a full 15,097-test run including every real-tmux module, and `Size the sweep`'s own whole-suite trace of 531 modules.

---

### Side finding: `SHARD_FIXED_COSTS["the selection map, traced"]` has drifted

`Size the sweep` measured its trace at **559 s** over **531 test modules**:

```
selection map: tracing 531 test modules (4 at a time)…
selection map: 23204 source files in 559s
```

The constant records **285 s**, taken from nine cache-miss traces in the 242–285 s band. That was honest when it was written; the suite has roughly doubled since. With the real number the itemisation is `559 + 240 + 38 + 15 + 3 = 855 s` against a `SHARD_FIXED` of **720 s** — so the fixed-cost budget no longer covers what a shard actually pays before its first mutation.

`test_the_budget_covers_the_itemisation_it_is_written_from` passes today only because the stale 285 keeps the sum under the budget. That is precisely the #670 shape that class exists to catch, surviving because the constant is stale rather than because the arithmetic is wrong.

**Deliberately not fixed in this PR.** Re-sizing the shard budget is a change to how the gate deals work, it needs its own measurements across several branches, and folding it into a CI tmux pin would put two unrelated judgements behind one review. The 38 s this PR adds is itemised honestly and is not what pushes the total over — the drift is there without it.

diazoxide:
**Review gap closed — the guard now fails closed. Green on `32d4849`.**

The reviewer was right and the hole was mine: `_runs_the_suite` knew three spellings, so `python -m unittest tests.test_x`, `pytest tests/` and `make test` all answered "runs no tests". A job written any of those ways sat outside the set the per-job assertion quantifies over *and* outside the pinned list — both halves green, running on tmux 3.4. The module's own claim that such a job "would never reach" the in-job check was true only for spellings it already knew.

`classify` now has three answers instead of two:

- **`RUNS`** — `_RUNNERS` widened to unittest in either form, pytest, py.test, nose/nose2, tox, `make test`/`make check`, run-tests.
- **`NOTHING`** — genuinely runs no tests.
- **`UNKNOWN`** — test-shaped and unrecognised. **Fails**, by name, in `NoJobsTestInvocationGoesUnclassified`.

The default for a shape nobody anticipated is now "this must be pinned", not silence. The `sweep.py` clause is inverted for the same reason — every mode runs tests *except* `--verdict`, so a sweep mode added later is pinned by default rather than by somebody remembering the line.

**Proved against real workflow files, not just the classifier.** Dropping a temporary workflow into `.github/workflows/`:

| injected job | before | now |
|---|---|---|
| `run: pytest tests/`, no tmux action | silently green | fails the per-job assertion **and** the pinned list |
| `run: python tests/harness_driver.py --all` | silently green | fails `test_every_job_this_reader_meets_it_can_answer_for` |

Cases cover each spelling that used to be missed, six that must be loud, five that must stay quiet (fail-closed must not mean fail-always — a reader that called everything UNKNOWN would be switched off within a week), and `_RUNNERS ⊆ _TEST_SHAPED`, so dropping a runner from the narrow pattern makes it UNKNOWN rather than NOTHING.

**On the optional note: taken, with a correction.** `local_target` alone is *not* the tighter key — it resolves any relative path, `actions/checkout@v7` included, so keyed on it by itself every third-party ref would be excused and the case would check nothing. What it adds over `is_local` is real, so the condition is `is_local(ref) and local_target(ref) is not None`: `./../elsewhere` is local by spelling and resolves outside the tree, and now stays subject to the assertion instead of being excused by its first two characters.

### CI

`test (3.11)` 9m53s · `test (3.12)` 9m49s · `test (3.13)` 10m37s · `test (3.14)` 10m40s · `Size the sweep` 9m03s · `Sweep shard 1 of 1` 51s · `Add up what the shards found` pass · verdict **`nothing to sweep`**, 0 shards refused · CodeQL + both Analyze pass.

Every job logged `tmux on this runner: tmux 3.5a`. All five new cases **ran rather than skipped** on the runner, `test_the_tmux_on_this_runner_is_at_or_above_the_floor` included. 15,101 tests, `OK (skipped=19)`.

Sweep is still `0 mutations → 1 shard` — the diff touches no swept path — so as before the shard is not the evidence the baseline is green under the pin; the four matrix runs and the 531-module trace are.

Issue #1114, opened 2026-09-16
Exit gate task 3: a tab can carry a name a person chose, and the id keeps doing the linking


diazoxide:
Task 3 of the one-exit-gate plan. Build it from the merged documents:
- spec: `docs/superpowers/specs/2026-09-15-one-exit-gate.md`
- plan: `docs/superpowers/plans/2026-09-15-one-exit-gate.md`, section "Task 3"

Depends on task 1 (#1106, merged) and task 2 (#1112).

## What it delivers

**A tab can carry a name a person chose** (decision 11). The id stays immutable and keeps doing the linking; the title is for people.

- **Set it from:** a rename row in the tab menu and in `F2` (a one-line input), an optional title at `+`, or the first line of a handoff brief.
- **Shown wherever a chat is named to a person:** the strip, the tab menu, quit rows, the picker. A workspace tab's selector shows chat titles only if it names chats.
- **The Claude session name becomes `<title> · <id>`**, or `<id>` when untitled. Measured: the name shows in the `claude --resume` picker, and Claude's in-session `/resume` hides the session it is in.
- **A rename takes effect on the tab at once**, and reaches Claude at its next start or resume, through `--name` again.
- **Charter never types `/rename` into a harness** (ADR 0018). It passes the name on the next launch, or not at all.

Today there is no title and no rename: a tab is its bare id.

## Rulings this task implements

1. **The id is never replaced by the title.** Every link, record, kill, reap and claim keeps using the id; the title is display only. A chat with no title behaves exactly as today.
2. **A title is text a person wrote, so it is contained** before it reaches a terminal, a record, a tmux option or a harness argument — `contain` on the way out, a bound on the way in.
3. **Renaming touches no harness.** No `send-keys`, no `/rename`. The new name reaches Claude Code only at the next start or resume.
4. **Codex and opencode:** neither takes a name at launch, so their tabs carry the title in charter's own surfaces only. Say so in `docs/harnesses.md` rather than implying parity.

## Done when

The plan's task 3 checklist and "Done when" are met, as one reviewed PR:
- tests first, each red without its guard;
- the isolation guards in place, and no real harness run or prompt spent;
- docs (`docs/frame.md`, `docs/harnesses.md`, `CONTEXT.md`) and a news entry moving with the code;
- every sweep survivor pinned or deleted, and any mutation the sweep cannot measure in time measured by hand.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

Issue #1112, opened 2026-09-16
Exit gate task 2: a harness that ends keeps its tab, and closing the tab is the one way to end a chat


diazoxide:
Task 2 of the one-exit-gate plan. Build it from the merged documents:
- spec: `docs/superpowers/specs/2026-09-15-one-exit-gate.md`
- plan: `docs/superpowers/plans/2026-09-15-one-exit-gate.md`, section "Task 2"

Depends on task 1 (#1106, merged). Closes nothing on its own; it is the amendment that makes the gate possible.

## What it delivers

**A harness that ends keeps its tab** (decisions 4, 5 and 12). Today every harness exit kills the window and reaps the chat, so a chat is gone the moment its harness exits.

- **Clean exit** (`/exit`, Ctrl+D): back to the profile selector, with **resume `<session name>`** preselected, then start fresh, then close the tab.
- **Crash** (a non-zero exit): the harness's last lines stay on screen, with the same three choices.
- **No conversation yet:** only start fresh and close.
- **A background or handed-off chat** keeps its tab too, marked ended in the strip.
- **Closing a tab is the one way to end a chat, and it ends it for good.** It confirms while a harness is live; an already-ended tab closes without asking.
- **Quit records ended-but-open tabs**, with their session link, so `charter reopen` brings them back ended and resumable.
- **`charter frame -- <cmd>` keeps today's behaviour:** its window closes when the command exits, and the exit code is returned to the caller. Operator's ruling.

## Rulings this task implements

1. **One function clears the ended state**, and every start path calls it: picking a profile, resuming from the drawer, starting fresh, and reopen. "Start fresh" goes to the profile selector.
2. **No kill or respawn acts on a record alone.** One tmux listing proves the drawer marker, `@charter_chat` and this plane's marker, on the recorded server, targeted by pane id. Respawn without `-k`. The ended state is claimed atomically, so the `pane-died` hook and a late check cannot both pass. Each guard gets a test that is red without it. This is the #933 and #1103 rule.
3. **End of input is not Esc.** A dropped terminal, a killed server or a closed pty leaves the tab ended and open: no closed mark, no forgotten transcript, no dropped manifest entry. Only a real Esc keypress closes an ended tab.
4. **Ctrl+C is its own key.** It keeps cancelling on every surface that cancels today, and on an ended tab's selector it does nothing.
[...]

Issue #1106, opened 2026-09-15
Exit gate task 1: a chat id is never handed out again, and every tab is linked to one harness session


diazoxide:
Task 1 of the one-exit-gate plan. Build it from these documents, merged via #1104:
- spec: `docs/superpowers/specs/2026-09-15-one-exit-gate.md`
- plan: `docs/superpowers/plans/2026-09-15-one-exit-gate.md`, section "Task 1"

Closes #1101.

## What it delivers
- A chat id names one chat for the life of the project and is never handed out again. The counter is `.charter/frame/chat-ids.json`, kept under a lock and raised above every trace: chat directories, `<id>.transcript`, the quit manifest, `.charter/sessions/<id>.*`. A restored chat keeps its id.
- Every tab is linked to exactly one harness session.
  - **Claude Code:** charter chooses the id and starts it with `--session-id <uuid> --name <name>`.
  - **Codex and opencode:** charter records the session id each harness reports.
  - **Resume** is offered only when the conversation exists: `--resume` for Claude Code, `codex resume <id>` for Codex, `opencode -s <id>` for opencode.

Nothing about how a chat ends changes in this task.

## Rulings this task implements
1. **Open question 1** (how an opencode chat is found on the hook path): **(A)**. Resolve the chat by `$TMUX_PANE` against the recorded panes, and refuse when more than one chat matches.
2. **Which harness sent a report** comes from the report itself, never from inherited env (`CHARTER_HARNESS`, `CLAUDE_PID`).
   - **Claude Code:** a report counts as the chat's own only when `CLAUDE_CODE_SESSION_ID == payload.session_id` AND either:
     - **adopt:** the payload's id is the chat's recorded id, and its `CLAUDE_PID` is adopted; or
     - **follow:** its `CLAUDE_PID` equals the adopted pid, as with `/clear`.

     The adopted pid comes from the adopting report, never from `#{pane_pid}`. The parent pid is never consulted.
   - **Codex and opencode:** the first reported id of each start is adopted; later ids in that start are ignored. Limit: `/new` is not followed.
3. **Reaping:** a server that did not answer proves nothing. Only a server that `tmuxctl.nothing_listening` confirms gone counts as holding no panes when a kept id is claimed. Any other failure refuses and names the directory.
4. **Counter:** the mark is written before the claim, and a failed write hands out nothing. The digit bound is `_MAX_ORDINAL_DIGITS`. The lock opens through `config.open_for`.

[...]

Issue #812, opened 2026-09-02
A workspace tab opened from a frame records the socket's absolute path, so every switch out of it is refused as a guest tmux


diazoxide:
A workspace tab opened from inside a frame creates a chat that **cannot be switched out of**.
The tab works, the new workspace opens, and then every tab — including the one you came from —
answers:

```
cannot switch: this chat is a window in your own tmux, where a workspace is not a session
— open the other workspace with `charter <harness> --workspace …`
```

Reported live by the operator: restart, session restored on `harness-wrapper`, click `fleet`,
land in a new empty chat, then no way back. `Ctrl-C` in the new chat closed it and the original
frame reappeared — so nothing was lost, but the tab is a one-way door.

## The mechanism

`is_operator_socket` is a **leading-slash test** (`frame/tmuxctl.py:298`):

```python
return bool(server) and server.startswith("/")
```

`cmd_launch` decides it is a guest from `$TMUX` (`tmuxctl.operator_server`). On this operator's
machine `$TMUX` is:

```
/private/tmp/tmux-502/charter,18923,83
```

— **charter's own socket, spelled as an absolute path.** So:

```
is_operator_socket('charter')                        -> False   (what an ordinary chat records)
is_operator_socket('/private/tmp/tmux-502/charter')  -> True    (what $TMUX yields)
```

A chat opened by the tab records the second spelling, `_switch_workspace` reads it back
(`commands_frame.py:6807-6820`), concludes the frame is a guest on somebody else's tmux, and
refuses by name — correctly, for a premise that is false.

**Two spellings of one socket are not the same string, and the guest test cannot tell them
apart.** The short name and the absolute path name the same server.

## Why the refusal itself is right

The refusal is not the bug and should not be deleted. Inside a genuine operator tmux a
workspace really is not a session — every chat is a `new-window` in the session the operator
was already in — so `switch-client` has nothing correct to do. That reasoning holds. What is
wrong is *concluding this frame is in one*.

## What a fix has to establish

* **Compare servers, not spellings.** `/private/tmp/tmux-$UID/<name>` and `<name>` are the same
  socket; the test needs to resolve both to one identity before asking whether charter started
  it. `tmuxctl.server_argv`'s own `-S` / `-L` split is the place that already knows the
  difference.
* **Then re-ask what a tab should do in a *real* operator tmux.** #811 made every tab open the
  workspace it names. In a genuine guest frame that open still succeeds and the switch still
[...]

PR #1124, opened 2026-09-17
One gate closes charter: F10, the identity row's button and two F2 rows, each detaching only the terminal that asked (#1097)

### LINKED ISSUE #1097: F2 → detach names a chat id as the tmux session, so on a per-workspace session it detaches nothing and says it did
`F2 → detach — leave the harness running` (`frame.detach`, `charter/frame/builtin_actions.py` `_detach`) runs:

```
tmux -L <server> detach-client -s <fid>
```

`<fid>` is a **chat id** such as `default.1`. Since chats became windows in a per-workspace session, the tmux **session** is named after the workspace (`default`) and the chat is a window inside it. tmux parses a target-session holding a dot as `session.pane`, so the target resolves to nothing and the detach fails.

## Measured

On tmux 3.7c, on a throwaway server with session `default` holding a window named `default.1` (and the same shape for `harness-profiles` / `harness-profiles.1`); `has-session -t` uses the same session-target resolver as `detach-client -s`:

```
has-session -t default.1            rc=1  can't find pane: 1
has-session -t default              rc=0
has-session -t harness-profiles.1   rc=1  can't find pane: 1
has-session -t harness-profiles     rc=0
```

The action spawns the command detached (`_spawn`), so the failure is silent: the palette reports `detaching — the harness keeps running`, and nothing detaches.

## Expected

Detach resolves the chat's own session through what charter already records, never through a name. For example, `commands_frame._pane_place(server, harness_pane)` answers the session **id** (`$N`), which is the one unambiguous spelling this file already measured for `new-window`. It should detach the clients of that session, on the chat's recorded server. A test on real tmux should attach a client and see it detach.

Found while scoping a single exit gate for the frame, where "close charter only" is this same detach.

https://claude.ai/code/session_01CgvXBsD94ZwsnaGt3Wi5S5

diazoxide:
Task 4 of the one-exit-gate plan, and the last of the four. Closes #1097. Depends on tasks 1 (#1106), 2 (#1112) and 3 (#1114), all merged.

Leaving charter was four gestures and no answer to *which terminal?* — closing the window detached it, `F2 → detach` was broken, `F2 → charter: quit` stopped the plane, `chat: close` stopped one tab. There is one way out now: `F10` wherever the frame draws, an `F10 close` button at the right end of the identity row, and the same two rows on `F2`.

## Step 0

Taken before any code, on throwaway `-L t4-*` servers started `-f /dev/null` with `TMUX`/`TMUX_PANE` unset — never the operator's servers, no harness, no prompt spent. Two clients attached over two ptys, **on tmux 3.7c and at the 3.2 floor**. Every reading passed on both.

| # | Reading | 3.7c | 3.2 | What it said |
|---|---|---|---|---|
| G1 | `bind -n F2` / `bind -n F10` running `echo "#{client_name}"`, each key's bytes written to client A's pty | PASS | PASS | The file held **A's** name and never B's, for both keys |
| G2 | `detach-client -t <A>` | PASS | PASS | A's `attach` exited; B stayed attached |
| G3 | **the gate** — `MouseDown1Pane` with the panel branch `set-option -F -p -t = @charter_presser "#{client_name}" ; send-keys -M`; 6 alternating SGR presses on a marked panel pane, 3 per client, order reversed on the last pass | PASS | PASS | `show-options -p -v` read back the **clicking** client every time, in either order of last activity, and never the literal `#{client_name}`. **The task continues.** |
| G4 | `has-session -t <session>.<n>` and `detach-client -s <session>.<n>` against session `default` with window `default.1` | PASS | PASS | See below — worse than #1097 says |
| G5 | the `F10` line and the changed hotkey and click binds, `source-file`'d | PASS | PASS | rc 0; `list-keys -T root` read `F2`, `F10` and `MouseDown1Pane` back byte for byte, with the escape hatch still last |
| G6 | `list-clients -F '#{client_name}\t#{session_id}'`, then `display-message -p -t <$N> '#{@charter_plane}'` | PASS | PASS | Both clients listed with their session; the marked session answered its plane |

**G4 is the reading that changed the fix.** #1097 says the detach resolves to nothing. It does — on a chat window with one pane. tmux parses `default.1` as `session.pane`, and pane index 1 exists the moment charter splits any strip off the harness, so on the ordinary framed chat the same command resolves to the **workspace's session** and detaches **every client attached to it**:

```
has-session -t solo.1    (one pane)    rc 1, "can't find pane: 1"
has-session -t solo.1    (two panes)   rc 0
has-session -t default                 rc 0
detach-client -s default.1 (two panes, two clients attached)
                                       rc 0, and NEITHER client left attached
```

So the row that promised *the harness keeps running* took the other operator's terminal with it, silently. The fix is not to make the target resolve; it is to never name a session.

## Before / after

| | Before (`c846740`) | After |
|---|---|---|
| Ways out of charter | four gestures, none called *close charter* | one menu, three ways in (`F10`, the button, `F2`) |
| `F2 → detach` | `detach-client -s <chat id>` — nothing on a bare chat, **every client** on a chat with a strip | `detach-client -t <presser>`, proven first |
| What proves a detach | nothing; the record was the target | one `list-clients` on the chat's recorded server: the presser attached to the session `_pane_place` resolved, and that session marked with this plane |
| Which terminal | unanswerable — the hotkey's bind carried no client since #729 | `#{client_name}` on both key binds; `@charter_presser`, written with `set-option -F`, on the click bind |
| No presser | n/a | the row is listed refused, naming the gesture that works. Never a fallback to every client |
| Cursor on the gate | n/a | pinned to *Close charter* by id, refused or not — `palette.aim` would open on *stop all chats* |
| Detach row's title | `detach — leave the harness running` | `Close charter (keep chats running)` |
| Quit row's title | `charter: quit — stop every harness on this plane` | `Close charter and stop all chats…` |
| Identity row's right end | the charter version | the version, then `F10 close` |
| Keys charter binds | `F2`, `F3`, `F12` | `F2`, `F3`, **`F10`**, `F12` — a component's `key` and `[frame] hotkey` are both refused it |
| Inside the operator's tmux | — | no key, no button, no gate menu; the two rows are in the palette and *Close charter* is refused there with their prefix key named |

## Each plan step, its commit and the test that was red without it

| Step | Commit | Red without it |
|---|---|---|
| `conf_text`'s three binds; `cli`'s `--gate`; `instance`'s reserved key | `8162ccb` | `TheBindsCarryThePresser.test_the_hotkey_bind_passes_client_name_again`, `test_f10_opens_the_gate_with_the_presser`, `test_a_click_on_a_panel_records_its_presser`; `TheKeyIsReserved.test_a_frame_hotkey_equal_to_it_falls_back_to_f2` |
| `builtin_actions._detach` presser-only and proven; `build(client=)` | `36fd9f7` | `CloseCharterDetachesOnlyThePresser` — `test_a_chat_id_is_never_the_target` / `test_no_route_ever_detaches_a_whole_session` were red at `c846740`, and each guard has its own case (another session, another plane, unmarked, unplaceable pane, unshaped client) |
| `gate.py`; `cmd_palette`, `_open_palette`, `_draw_palette` | `bc19efc` | `TheMenu.test_the_cursor_never_opens_on_stop_even_when_detach_is_refused` — **checked by hand**: the `_refilter` override deleted, `__pycache__` cleared, the module run (1 failure), the file restored and its sha compared |
| `_Doors`, `_top`, `_strip_events` | `700260a` | `TheButton` (7 cases, all red before), `AClickReadsTheRecordedPresser` (5) |
| `instance`'s reserved key; `leave.OPEN_QUIT` | `cbb4588` | `TheF2RowsAreTheGate.test_the_stop_row_is_the_gates_words`, `test_close_is_still_the_last_row` |
| the real-tmux half | `715baf3` | `test_the_marker_is_compared_as_a_path_and_not_as_a_string` — see below |
| the floors this moves | `8da86c5` | — |
| `Gate` as an ordinary subclass, not a factory over a cached class | `9d1b298` | — (the cache it removes was a line no input could turn red) |
| the button is compared with the key it names | `bc826ca` | `TheButton.test_the_button_names_the_key_charter_binds` |
| three more unpinnable guards deleted, and the unit half's tmux speaks tmux's format | `f2f31be` | — |

## What the real-tmux test found

`tests/test_the_gate_detaches_a_real_client.py` attaches **two** real clients over two ptys to one real frame and asks tmux afterwards who is still attached. It found a defect no unit test structurally can: `_detach` compared the session's `@charter_plane` with `_this_plane()` as a **string**, and the two sides are read in different processes — the marker holds whatever spelling the writer resolved, and the palette's pane resolves its own from a `-e` payload. On macOS `/var` is a symlink to `/private/var`, so they arrived as `/var/…/.charter` and `/private/var/…/.charter`, and **every detach on this machine was refused** — silently, with the row reporting that the terminal was not attached to the frame it had just been pressed in. Both sides go through `os.path.realpath` now, which is `frame/ended.py`'s own reading of #812 one surface over. `not plane` stays in front of the comparison and is now load-bearing: `os.path.realpath("")` answers the process's cwd.

## Guards, and what pins each

- **Nothing acts on a record alone.** The chat's recorded server and pane say where to look and are never the target; one `list-clients` proves the presser, `@charter_plane` proves the session. Tested against another plane's session, an unmarked one, a client on another session, a pane tmux will not place, and a server that did not answer — each its own case.
- **`-s` appears nowhere**, and neither does a chat id on any argv this module builds (`test_no_route_ever_detaches_a_whole_session`, `test_a_chat_id_is_never_the_target`).
- **The preselected row is never the destructive one** — pinned by id in `gate.Gate`, red without it.
- **Close stays last** in `leave.open_rows`, and `_palette_catalogue`'s last row is still it.
- **Decision 7** — no bind, no button, no menu inside the operator's own tmux, each with its own case and a positive control beside it.
- A value that reaches an argv is held at the boundary: `gate.CLIENT_RE` refuses `#{client_name}` unexpanded, which is what the option would hold if the click bind ever lost its `-F`.

## The sweep

Run `35206062424` on `2002346`: every shard reported, **no refusals**, **8 survivors**. All eight sit on this branch's own commits (`git blame` at the swept sha: `8162ccb` ×1, `36fd9f7` ×2, `bc19efc` ×3, `700260a` ×2); **none is pre-existing**. Read from the run's own `deletion-sweep-*` artifacts.

| # | Survivor | Disposition |
|---|---|---|
| 1 | `builtin_actions.NO_PRESSER_TO_DETACH` — `retune-string` | **pinned** — `TheExitGatesOwnWordsAreTheseWords.test_a_detach_charter_cannot_attribute_names_the_gesture_that_works` |
| 2 | `builtin_actions.NOT_ATTACHED_HERE` — `retune-string` | **pinned** — `…test_a_terminal_this_chat_does_not_hold_is_told_it_is_not_this_chats` |
| 3 | `gate.opens`' `state.own_workspace(fid) or ""` — `no-fallback` | **deleted** — this path draws rows and writes no manifest, so nothing reads `Plan.focus`. `commands_frame._picker` removed the same fallback at the same seam for the same finding; `cmd_quit`, which does record a focus, spells its own |
| 4 | `_top`'s `if not right: return tui.truncate(identity, w)` — `drop-if` | **deleted** — with *right* empty the general path is `tui.Cell(identity, w)` beside a zero-width cell, which IS a truncate. Measured against `tui.truncate` at three widths (room to spare, a cut to `…`, three columns): byte-identical every time. It composes a string; it kills, detaches and claims nothing |
| 5 | `gate.act`'s `if row.note` — `drop-if` | **pinned** — `TheMenu.test_a_row_that_started_nothing_says_why_it_could_not`, paired with `test_a_row_with_nothing_to_say_says_nothing` because a `drop-if` takes the guard AND its body |
| 6 | `_top`'s `… + 1 <= w` — `shift-boundary` | **pinned** — `TheButton.test_the_version_needs_one_spare_column_and_not_two`, measured as the GAP at the narrowest width that draws the version (the version string changes length every release, so a case naming a column would go red at the next bump) |
| 7 | `gate.LABEL` — `retune-string` | **pinned** — `…test_the_gates_header_names_the_noun_before_the_rows` |
| 8 | `catalogue`'s `"" if client else …` — `collapse-ifexp` | **pinned** — `TheMenu.test_a_row_that_can_run_carries_no_reason`, paired with the refused case beside it |

Each of the six pins was hand-checked: the mutant applied, `__pycache__` cleared, the module run, the file restored and its sha compared. All six went red, each caught by the intended case.

Rebased onto `9c35607` (#1122, the tmux 3.5 floor for the ended tab) after the survivor work, with no conflict.

## Verified

Locally, only the modules this touches plus `tests.test_claims` — 1235 tests, green, including the 7 real-tmux cases. No full suite and no sweep locally.

Docs move with the code: `docs/frame.md` (`## Leaving` opens with the gate, the identity row, `## Inside a tmux you already have`, the reserved keys and the `hotkey` paragraph, the palette sample), `CONTEXT.md` (**Exit gate**, **Presser**), and `docs/news/unreleased-one-gate-closes-charter.md`. `[frame] hotkey` has no section in `docs/control-plane.md` — it is documented in `docs/frame.md` only — so the refusal is written there, in both places that already discuss which keys charter has taken.


diazoxide:
Closes #1097. Part of #1115 (task 4 of the one-exit-gate plan).

